PR #5, opened 2020-01-29
Fix/jenkins slave

DerekRoberts:


DerekRoberts:
BCDK has been deprecated and a replacement is on the way.  Rollout will be paused accordingly.

PR #137, opened 2020-03-11
Shared Scripts and Subtree

DerekRoberts:


DerekRoberts:
> Can't we just delete all those DOTjenkins files at this point?
They were used to create our Jenkins instance, but can be removed If we move to GitHub Actions!

PR #304, opened 2020-05-12
adding a timeout to rollout

Sybrand:


oc_deploy.sh
```diff
@@ -36,7 +36,7 @@ OC_APPLY="oc -n ${PROJ_TARGET} apply -f -"
 
 # Deploy and follow the progress
 #
-OC_DEPLOY="oc -n ${PROJ_TARGET} rollout latest dc/${NAME_OBJ}"
+OC_DEPLOY="oc -n ${PROJ_TARGET} rollout latest --request-timeout=1m dc/${NAME_OBJ}"
```
Kyubinhan, line 39: I think we need to adjust `--pod-running-timeout` in the next command to address the error you showed me. We just need to make `oc logs` to wait longer so that it won't timeout. How about give `--pod-running-timeout=3m`?

Sybrand: On further investigation
1) --request-timeout defaults to 0, which means, don't timeout. So this change is garbage just from that perspective.
Re. it failing on log - I actually struggle to see where it's failing...

PR #341, opened 2020-06-04
Task/high availability/325

Sybrand:
Added overrides for adjusting cpu/memory and replicas


Sybrand:
subtree wasn't added quite right on wps-web or wps-api

PR #436, opened 2020-08-18
Patroni pod health checks

andrea-williams:


api/app/main.py
```diff
@@ -86,8 +88,33 @@
 @app.get('/health')
 async def get_health():
     """ A simple endpoint for Openshift Healthchecks """
-    LOGGER.info('/health')
-    return {"message": "Healthy as ever"}
+    try:
+        parts = [
+            'https://console.pathfinder.gov.bc.ca:8443/apis/apps/v1beta1/namespaces/',
```
Sybrand, line 93: I sneaked in an environment variable, could you please us that?: PATHFINDER_BASE_URI

api/app/tests/test_health.py
```diff
@@ -0,0 +1,37 @@
...
+from starlette.testclient import TestClient
+import app.main
+from app.tests.common import MockRequestResponse
+
+
+def test_health_ok():
+    """ Test health endpoint, given that everything is fine """
+    client = TestClient(app.main.app)
+    response = client.get('/health/')
+    assert response.status_code == 200
```
Sybrand, line 15: The 200 check is redundant - you can remove it (I prefer also only having one assertion per test)

api/app/.env.example
```diff
@@ -16,3 +16,7 @@ POSTGRES_PASSWORD="wps"
 POSTGRES_DATABASE="wps"
 POSTGRES_PORT="5432"
 PORT="8080"
+PATHFINDER_BASE_URI=https://console.pathfinder.gov.bc.ca:8443
+PROJECT_NAMESPACE=openshift_project_namespace
```
Kyubinhan, line 20: No need to add an env variable for this because we already keep track of this (check out `openshift/scripts/oc_deploy.sh` line 29)

api/app/main.py
```diff
@@ -86,8 +88,34 @@
...
-    return {"message": "Healthy as ever"}
+    try:
+        parts = [
+            config.get('PATHFINDER_BASE_URI'),
+            '/apis/apps/v1beta1/namespaces/',
+            config.get('PROJECT_NAMESPACE'),
+            'statefulsets',
+            config.get('PATRONI_HEALTH_SUFFIX')
+        ]
+        url = reduce(os.path.join, parts)
```
Kyubinhan, line 99: Man... I really miss template literals in Javascript

Kyubinhan: Oh actually there is a similar way: https://realpython.com/python-string-formatting/#3-string-interpolation-f-strings-python-36

api/app/main.py
```diff
@@ -86,8 +88,34 @@
 @app.get('/health')
 async def get_health():
```
Kyubinhan, line 89: Nice to have: maybe create `health.py`?

api/app/tests/common.py
```diff
@@ -43,15 +43,36 @@ async def __aexit__(self, *error_info):
         """ Clean up anything you need to clean up """
 
 
+class MockRequestResponse:
```
Kyubinhan, line 46: MockRequestResponse and MockResponse seem to be identical, is it intended?

openshift/templates/deploy.dc.yaml
```diff
@@ -128,6 +116,23 @@ objects:
                   value: "5432"
                 - name: POSTGRES_DATABASE
                   value: ${NAME}-${SUFFIX}
+                - name: PROJECT_NAMESPACE
+                  valueFrom:
+                    configMapKeyRef:
+                      name: ${NAME}-global
+                      key: env.project-namespace
+                - name: PATRONI_HEALTH_SUFFIX
```
Kyubinhan, line 124: I am not sure `PATRONI_HEALTH_SUFFIX` is the right name for the value. The value is patroni's pod name, right?

api/app/tests/common.py
andrea-williams: this is moreso a question for @Sybrand , but from what I can tell MockResponse is async and MockRequestResponse isn't

openshift/templates/deploy.dc.yaml
andrea-williams: it's not quite the pod name, since pods are named `patroni-wps-api-dev-0` or `-1` or whatever, and `PATRONI_HEALTH_SUFFIX` just takes the `patroni-wps-api-dev` part. So really it's the patroni cluster name

Kyubinhan: You are right it's a cluster name, my bad. So how about PATRONI_CLUSTER

andrea-williams: we already have a var called `PATRONI_CLUSTER_NAME` from patroni.yaml template - I'm checking to see if I can just make use of that for my purposes too

api/app/main.py
Sybrand: What I'm really looking for, which os.path.join almost gives, is not caring about slashes. Often, in configuration, you'll forget to add a trailing slash - because you're not in the context of the code. So having a function that neatly takes all the parts and makes sure it joins them correctly is real sweet. So you can join ['a', 'b', 'c'] and it becomes 'a/b/c', similarly ['a/','b/', 'c/'] becomes 'a/b/c/' - doesn't throw in weird slashes. Problem with os.path.join, is that it's really made for folders, so if you give it ['b', 'c', '/d'] it gives you '/d' :(. I don't want to do a + b + c.

openshift/templates/deploy.dc.yaml
Kyubinhan: `PATRONI_CLUSTER_NAME` gets created while running `oc_provision_db.sh` so you would wanna do the same in `oc_deploy.sh`

api/app/tests/common.py
Sybrand: Oi - bad class names - we probably need MockAsyncResponse and MockResponse?

api/app/main.py
Kyubinhan: I was referring to 
```
def greet(name, question):
...     return f"Hello, {name}! How's it {question}?"
```
but I see your point

api/app/tests/common.py
Kyubinhan: Sounds good. I thought they are identical but I didn't see `async` for class methods

api/app/main.py
Sybrand, line 99: `url = "/".join(map(lambda part: part.rstrip('/'), parts))`

Sybrand: Taken from: https://stackoverflow.com/questions/1793261/how-to-join-components-of-a-path-when-you-are-constructing-a-url-in-python

api/app/health.py
```diff
@@ -0,0 +1,44 @@
+""" Module to perform health checks.
+"""
+import os
+import json
+import logging
+import logging.config
+import requests
+from app import config
+
+LOGGING_CONFIG = os.path.join(os.path.dirname(__file__), 'logging.json')
```
Sybrand, line 10: Please remove the logging config section - it's configured in main.py - you only need the logger.

PR #441, opened 2020-08-19
Bug - model prediction not showing up

Sybrand:
- added column to db to track if model run is complete.
- returning latest complete model run.
- updated unit tests accordingly.
- updated env-canada code to mark model run as complete when done.
- updated poetry.lock with latest versions


api/alembic/versions/6f6effb6b004_add_complete_flag_to_prediction_model.py
```diff
@@ -0,0 +1,34 @@
...
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # Add "complete" column.
+    op.add_column('prediction_model_run_timestamps', sa.Column(
+        'complete', sa.Boolean(), nullable=True))
+    # Mark all rows as completed. (They might not be, but let's just assume....)
+    op.execute('update prediction_model_run_timestamps set complete = True')
```
Kyubinhan, line 24: Just read the article about this.
It's kind of annoying that you have to do this :(

api/app/db/crud.py
```diff
@@ -41,17 +45,28 @@ def get_prediction_run(session: Session, prediction_model_id: int,
...
+        prediction_model_id=prediction_model_id,
+        prediction_run_timestamp=prediction_run_timestamp,
+        complete=complete)
     session.add(prediction_run)
     session.commit()
     return prediction_run
 
 
+def update_prediction_run(session: Session, prediction_run: PredictionModelRunTimestamp):
+    """ Updarte a PredictionModelRunTimestamp record """
```
Kyubinhan, line 65: Typo Updarte

api/app/db/models.py
```diff
@@ -69,6 +69,14 @@ class PredictionModelRunTimestamp(Base):
...
     # The date and time of the model run.
     prediction_run_timestamp = Column(TIMESTAMP(timezone=True), nullable=False)
+    # Indicate if this particular model run is completely downloaded.
+    complete = Column(Boolean, nullable=False)
+
+    def __str__(self):
+        return ('id:{self.id}, '
+                'prediction_model:{self.prediction_model.abbreviation}:{self.prediction_model.projection}, '
+                'prediction_run_timestamp:{self.prediction_run_timestamp}, '
+                'complete={self.complete}').format(self=self)
```
Kyubinhan, line 79: Nice!

api/app/models/env_canada.py
```diff
@@ -169,48 +174,105 @@ def flag_file_as_processed(session, url):
...
+    # Iterate through url's.
+    for url, _ in urls:
+        # Check if url as been processed.
+        processed_file_record = get_processed_file_record(session, url)
+        # If the url has not been processed, mark as false and stop.
+        if not processed_file_record:
+            logger.info(
+                'for model %s:%s run %s, url %s not yet processed', model, projection, hour, url)
+            complete = False
+            break
```
Kyubinhan, line 198: `SELECT url FROM processed_model_run_urls WHERE url IN ( 'url1', 'url2', ... );` would be more suitable but I would say this is nice to have

api/app/models/env_canada.py
```diff
@@ -169,48 +174,105 @@ def flag_file_as_processed(session, url):
...
+                            # If we've downloaded the file ok, we can now process it.
+                            try:
+                                processor.process_grib_file(
+                                    downloaded, model_info)
+                                # Flag the file as processed
+                                flag_file_as_processed(session, url)
+                                files_processed += 1
+                            finally:
+                                # delete the file when done.
+                                os.remove(downloaded)
```
Kyubinhan, line 262: Do we need another try block inside?

api/app/models/env_canada.py
Sybrand: It would be a ```SELECT count(url)``` and then check if the number is the same. Agreed it would be faster - I thought about it but didn't like a) writing more code and b) sending such a monster query down - going to be a really big query...

api/app/models/env_canada.py
Sybrand: That try is dealing specifically with cleanup of the file, in that context. So yeah - I'd argue we want it, even though it's ugly. Should we chat? Do you have a more elegant proposal?

Sybrand: Sonarcloud doesn't like the cognitive complexity in this section of code though - so I'm going to re-factor it a bit.

api/app/models/env_canada.py
Sybrand: Re-factored - I like it more now.

api/app/models/env_canada.py
```diff
@@ -154,74 +167,171 @@ def download(url: str, path: str) -> str:
...
+        self.files_processed = 0
+        self.exception_count = 0
+        # We always work in UTC:
+        self.now = get_utcnow()
+        self.session = app.db.database.get_session()
+        self.grib_processor = GribFileProcessor()
+
+    def flag_file_as_processed(self, url):
+        """ Flag the file as processed in the database """
+        processed_file = get_processed_file_record(self.session, url)
```
Kyubinhan, line 211: Are you querying this again to update `update_date` field?

api/app/models/env_canada.py
```diff
@@ -154,74 +167,171 @@ def download(url: str, path: str) -> str:
...
+        if processed_file:
+            logger.info('re-procesed %s', url)
```
Kyubinhan, line 213: I don't know this is the right msg for this situation cause we ignore the ones that already processed right?
also typo re-procesed

api/app/models/env_canada.py
```diff
@@ -154,74 +167,171 @@ def download(url: str, path: str) -> str:
...
+            datetime.timezone.utc)
+        # pylint: disable=no-member
+        self.session.add(processed_file)
+        self.session.commit()
+
+    def check_if_model_run_complete(self, urls):
+        """ Check if a particular model run is complete """
+        # pylint: disable=no-member
+        actual_count = self.session.query(ProcessedModelRunUrl).filter(
+            ProcessedModelRunUrl.url.in_(urls)).count()
```
Kyubinhan, line 229: Cooooool

api/app/models/env_canada.py
```diff
@@ -154,74 +167,171 @@ def download(url: str, path: str) -> str:
...
+        """
+        for url in urls:
             try:
                 # check the database for a record of this file:
-                processed_file_record = get_processed_file_record(session, url)
+                processed_file_record = get_processed_file_record(
+                    self.session, url)
                 if processed_file_record:
                     # This file has already been processed - so we skip it.
                     logger.info('file aready processed %s', url)
```
Kyubinhan, line 244: Typo aready

api/app/models/env_canada.py
Sybrand: On it's own this function makes sense - I agree in the context of it's usage, it does a redundant check and log - but I don't mind - the logic is sound by itself.

api/app/models/env_canada.py
Sybrand: See above comment re. function making sense in isolate.

PR #451, opened 2020-08-24
Implement endpoint /noon_forecasts/summaries/ for historic noon forecasts

Kyubinhan:
* Create an endpoint `/noon_forecasts/summaries/`
* Use `defaultdict` which provides the default value for a nonexistent key
* Modify docker-compose files a little bit


api/Makefile
```diff
@@ -85,6 +85,11 @@ shell:
 	# Run shell in virtual environment shell.
 	poetry shell
 
+copy-prod-db:
+	# Download the production database copy
+	# Make sure to login to Openshift first
+	PROJECT=auzhsi-prod POD=patroni-wps-api-prod-2 DATABASE=wps-api-prod scripts/copy_db_from_pod_to_local.sh
```
Sybrand, line 91: pg_dump does an explicit lock on the table it's currently working on, so it's best we ensure that we're running pg_dump on a replica, not a leader. If you hard code the pod, you may inadvertently lock the leader database. I'd prefer we not have this shortcut, unless it's smart enough to first identify a replica.

api/app/forecasts/noon_forecasts.py
```diff
@@ -23,7 +24,7 @@ def parse_table_records_to_noon_forecast_response(data: [NoonForecasts]):
     (which is a NoonForecasts object) and structure it as a NoonForecast
     object, then return the list of NoonForecast objects as a NoonForecastResponse
     """
-    noon_forecasts = {}
+    noon_forecasts = defaultdict(list)
```
Sybrand, line 27: Oh! I love this so much!

api/app/pytest.ini
```diff
@@ -1,7 +1,7 @@
 [pytest]
 filterwarnings =
     ignore::DeprecationWarning
-log_format = %(asctime)s %(levelname)s %(message)s
+log_format = %(levelname)s %(message)s
 log_date_format = %Y-%m-%d %H:%M:%S
-; log_cli=true
-; log_level=DEBUG
+log_cli=true
```
Sybrand, line 6: I'd prefer we don't commit these - it makes for an extremely verbose log file. Was this a mistake or do you really need more detailed logging?

Kyubinhan: Hmm... I set these because I wanted to print out things to debug while testing. Is there another way to enable logging when running tests?

also, quick question, do we store log files for testing?

api/Makefile
Kyubinhan: Okay, I can remove this. So from now on, I need to check the pod is a replica first then run the command right?

Sybrand: Yep - that's best. It will still block updating of the database on that server, but it shouldn't block the leader I don't think.

api/app/pytest.ini
Sybrand: What I've been doing is setting it to log_cli=true and log_level=DEBUG when I'm trying to fix tests or troubleshoot - but when I just want to check if my tests are passing, having a crazy verbose list of results is just too much.
The logs are around for a while in the github actions.

api/app/db/models.py
```diff
@@ -187,3 +187,11 @@ class NoonForecasts(Base):
     danger_rating = Column(Integer, nullable=False)
     created_at = Column(TIMESTAMP(timezone=True), nullable=False,
                         default=datetime.datetime.now(tz=timezone.utc))
+
+    def __str__(self):
+        return (
+            'weather_date:{self.weather_date}, '
+            'created_at:{self.created_at}, '
+            'temp={self.temperature}, '
+            'ffmc={self.ffmc}'
```
andrea-williams, line 196: did you mean to print rh instead of ffmc?

api/app/pytest.ini
Kyubinhan: Okay, changes have been reverted!

Sybrand: Doesn't log_cli=true also give a lot of additional stuff?

api/app/db/models.py
Kyubinhan: No, I wanted to check what one of the fire indexes would look like

PR #454, opened 2020-08-26
Low disk space notification

Sybrand:
- Send notification on rocket.chat if disk space falls below an absolute critical threshold value.


openshift/templates/patroni.yaml
```diff
@@ -152,6 +152,20 @@ objects:
...
+                - name: ROCKET_AUTH_TOKEN
+                  valueFrom:
+                    secretKeyRef:
+                      name: ${NAME}-global
+                      key: rocket.chat-auth-token-secret
+                - name: ROCKET_USER_ID
+                  valueFrom:
+                    secretKeyRef:
+                      name: ${NAME}-global
+                      key: rocket.chat-user-id-secret
```
andrea-williams, line 164: should user IDs be kept secret? because I just moved the user id for bcfw p1 authentication from a secret to a config

openshift/templates/patroni.yaml
```diff
@@ -340,9 +362,16 @@ parameters:
   - description: Version of Patroni image to be used.
     displayName: Patroni image version
     name: IMAGE_VERSION
-    value: "v10-build5"
+    value: "v10-pr-454"
```
andrea-williams, line 365: I'm now super confused about the whole patroni image versioning process, but just wanted to flag for you in case this is unintentional that there's a PR-specific value here

openshift/templates/patroni.yaml
Sybrand: I noticed that - but didn't want to have more to and fro since I think this borders on the philosophical.  I consider a username secret since it's part of a login/identity, so I would prefer it be a secret. Two parts to hacking right - you need a username and a password.

openshift/templates/patroni.yaml
Sybrand: You're not alone! I have myself confused - my though patterns evolved as such:
- Just having 10, becomes a pain, since my PR may be intricately linked to the version of the image. Also there's no context as to what 10 is.
- Using a build number is a pain, because it keeps going up, so I keep needing to change which version I use - again there's no context.
- Giving it a unique name helps me manage the image for my PR, and I know that when my PR goes to prod it will be associated to the correct image.
- Using a PR gives me context as to where this build came from, but may be confusing as subsequent PR's are going to be referring the the last PR that change the image-stream being used.

I'm open to suggestions.
Maybe v10-rev1, v10-rev2 etc. rather than PR? Will lose context but so what?

andrea-williams: I'm perfectly happy with using PR numbers as part of the naming convention. Just wanted to double-check that this is now the process and that it wasn't left over from you experimenting with a specific image just for the PR

openshift/templates/patroni.yaml
andrea-williams: yeah now that I think about it more, it seems safer to me to include user IDs in secrets. From what I understand there's no harm in having a few extra variables in Secrets rather than ConfigMaps, and it should add a bit of extra security

openshift/templates/patroni.yaml
Sybrand: "process" is maybe too generous at this moment in time.

PR #460, opened 2020-08-29
Unit test - fixture re-factor

Sybrand:
# Unit test - fixture re-factor
## Overview
As step 1 for Marvin (hourly actuals) - please review this PR relating to unit testing and fixtures.

## Rationale
I was bothered by the effort involved with having to mock out requests for every test and how cumbersome and lacking the existing fixture code was.

## Highlights
- changed entirely how fixtures are loaded automatically when mocking out requests.
- added mock support for requests.Session type calls.
- moved get_utc_now to shared location, and ensuring it get's automatically mocked out.


api/app/tests/fixtures/loader.py
```diff
@@ -0,0 +1,136 @@
...
+    def get_base_path(self, url: str):
+        """ Get the bast path of the provided url.
+        """
+        # break url into parts.
+        split = urlsplit(url)
+        # construct the base path.
+        base_path = os.path.join(os.path.dirname(__file__), split.netloc)
+        # check if it exists.
+        if not os.path.exists(base_path):
+            if config.get('AUTO_MAKE_FIXTURES'):
```
Kyubinhan, line 62: `AUTO_MAKE_FIXTURES` needs to be in `.env.example` right?

sonar-project.properties
```diff
@@ -11,7 +11,7 @@ sonar.sources=.
 
 sonar.exclusions=**/Makefile, **/*.Dockerfile, api/app/data/**
 sonar.test.exclusions=*.feature
-sonar.coverage.exclusions=api/scripts/**, api/alembic/**
+sonar.coverage.exclusions=api/scripts/**, api/alembic/**, api/app/tests/**
```
Kyubinhan, line 14: What is this for?

api/pyproject.toml
```diff
@@ -37,6 +37,7 @@ autopep8 = "^1.5.3"
 matplotlib = "^3.2.2"
 pytest-xdist = "^1.33.0"
 rope = "^0.17.0"
+docker-compose = "^1.26.2"
```
Kyubinhan, line 40: what is the purpose of this python module?

api/app/tests/common.py
```diff
@@ -46,18 +39,14 @@ async def __aexit__(self, *error_info):
 class MockResponse:
     """ Stubbed response object. """
 
-    def __init__(self, text_response: str = None, json_response: dict = None, status_code=200):
+    def __init__(self, text: str = None, json_response: dict = None, status_code=200, content=None):
```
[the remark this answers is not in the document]
Sybrand: Yeah - that's a good idea, going to rename json_response to json as well

sonar-project.properties
Sybrand: We don't need coverage for test code. Then we'd need to write tests for our tests.

api/pyproject.toml
Sybrand: This way you don't have to install docker compose locally - it's installed by pip, so we have a consistent docker-compose across all our machines.
I ran into this when I got around to installing docker-compose on my ubuntu box, there isn't a nice packaged way to install it, and the only thing using docker-compose is this project - so it just makes sense.

sonar-project.properties
Kyubinhan: Oh I see, I thought it will always be 100% lol

api/app/tests/fixtures/loader.py
```diff
@@ -0,0 +1,136 @@
+""" Code for automatic loading of fixture files based on requests. """
+import os
+from collections import defaultdict
```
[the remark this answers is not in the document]
Sybrand: I tried another approach, have a look - making an infinitely nested default dict using and existing dictionary isn't as clean as I would like (I don't like having to type check the result, and the recursive function I'm using looks pretty ugly to me). I don't think there's a good win here.

sonar-project.properties
Sybrand: Well - it's kinda messy. It wants the test code to run, so if there's any kind of conditional that doesn't happen in your test code, that comes up as not covered. So in this case, because there's a conditional on AUTO_MAKE_FIXTURES, there's test code, that just doesn't run.
That's all good an well, yes, it's code that doesn't run as part of a unit test - but I really don't care. I mean, it's nice to know if we have unit tests that don't get run at all - but that's a different problem.

Sybrand: So I'm fine with just a blanket: don't code unit tests as part of your coverage.

PR #461, opened 2020-08-29
Modification for displaying historic noon forecasts with graph and table

Kyubinhan:
* Send min and max for temp & rh of forecasts instead of percentiles
* Send noon forecasts for both past and future
* Change model class name to `NoonForecasts`


api/app/main.py
```diff
@@ -132,9 +132,10 @@ def get_noon_forecasts(request: schemas.StationCodeList, _: bool = Depends(authe
     set of weather stations. """
     try:
         LOGGER.info('/noon_forecasts/')
-        start_date = datetime.datetime.now(tz=datetime.timezone.utc)
-        end_date = start_date + datetime.timedelta(days=5)
-        return fetch_noon_forecasts(request.stations, start_date, end_date)
+        now = datetime.datetime.now(tz=datetime.timezone.utc)
```
Sybrand, line 135: Once my test PR is merged, please switch to using app.time_utils.get_uct_now - it makes mocking for unit tests real easy if we mock out .now functionality in the same place.

api/app/forecasts/noon_forecasts.py
```diff
@@ -19,9 +18,9 @@ class StationNotFoundException(Exception):
     """ Custom exception for when a station cannot be found """
 
 
-def parse_table_records_to_noon_forecast_response(data: [NoonForecasts]):
+def parse_table_records_to_noon_forecast_response(data: [app.db.models.NoonForecast]):
```
andrea-williams, line 21: does this work even though you're not importing app.db.models?

api/app/forecasts/noon_forecasts_summaries.py
```diff
@@ -38,12 +37,10 @@ def create_noon_forecast_summary(station: WeatherStation,
     for date in nested_dict:
         percentile_values = NoonForecastSummaryValues(
```
andrea-williams, line 38: could you please rename this variable, since it's no longer percentiles but min/max?

api/app/main.py
```diff
@@ -145,7 +147,7 @@ async def get_noon_forecasts_summaries(request: schemas.StationCodeList, _: bool
     """ Returns summaries of noon forecasts for given weather stations """
     try:
         LOGGER.info('/noon_forecasts/summaries/')
-        now = datetime.datetime.now(tz=datetime.timezone.utc)
+        now = time_utils.get_utc_now()
```
Sybrand, line 150: I don't think that the fixtures pick it up correctly when you do it this way.
:(
I think have to import:
```import app.time_utils```
and then use:
```app.time_utils.get_utc_now()```

File: api/app/models/env_canada.py [hunk 5]
```diff
@@ -227,7 +226,8 @@ def check_if_model_run_complete(self, urls):
         # pylint: disable=no-member
         actual_count = get_processed_file_count(self.session, urls)
         expected_count = len(urls)
-        logger.info('we have processed %s/%s files', actual_count, expected_count)
+        logger.info('we have processed %s/%s files',
```
Sybrand, line 229: What are you using for autoformatting? I think it's cutting things shorter than they need to be.
```
"python.formatting.autopep8Args": [
        "--max-line-length=110"
    ]
```

api/app/forecasts/noon_forecasts.py
Kyubinhan: It shouldn't, right?
I am guessing it works because db is imported through `__init__.py`?

Sybrand: Weird.

Kyubinhan: I will add that anyway just in case!

api/app/forecasts/noon_forecasts_summaries.py
Kyubinhan: Sure!

api/app/tests/forecasts/test_noon_forecasts_summaries.py
```diff
@@ -11,12 +11,12 @@
...
 import app.main
 from app.db.models import NoonForecast
-from app import time_utils
+import app.time_utils as time_utils
 
 
 LOGGER = logging.getLogger(__name__)
 
-noon = time_utils.get_utc_now().replace(
+noon = app.time_utils.get_utc_now().replace(
```
andrea-williams, line 19:
```suggestion
noon = time_utils.get_utc_now().replace(
```

PR #465, opened 2020-08-31
Hourly actuals bot ("Marvin")

Sybrand:
- Added hourly actuals bot (a.k.a. "Marvin")
- Re-factored Bender to share code base with Marvin.
- Modified Bender to insert timezone info.
- Added TZTimeStamp type to enforce presence of timezone info on insertion.
- Increased CPU limit on api pods from 300m to 500m (hoping to slightly speed up startup).
- Increased the CPU request (1 to 2) and limit (2 to 4) for the build. (hoping for slightly faster build)
- Changed env canada to run 13 minutes after the hour (as a courtesy to env. canada, so they don't get flooded on the hour)


api/app/fireweather_bot/hourly_actuals.py
```diff
@@ -0,0 +1,119 @@
...
+        # with a fix, this is the easiest work-around.
+        # See https://github.com/pandas-dev/pandas/issues/15988
+        # pylint: disable=unused-variable
+        for index, row in data_df.iterrows():
+            try:
+                # Go from pandas to a python dict.
+                data = row.to_dict()
+                # Fix the timestamp, make it be tz aware.
+                # Note: There must be a more "pandas" way of doing this, but I don't know how.
+                data['weather_date'] = _fix_pandas_datetime(data['weather_date'])
```
andrea-williams, line 79: I believe it's the map() function you're looking for - https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.Series.map.html

api/Makefile
```diff
@@ -135,4 +141,4 @@ docker-run-env-canada:
 docker-run-bender:
 	# Run a python script to download forecast data from BC Wild Fire on docker
 	# Needs to run "make docker-run" first
-	docker-compose exec api python -m app.fireweather_bot
\ No newline at end of file
+	docker-compose exec api python -m app.fireweather_bot_noon_forecasts
```
Kyubinhan, line 144:
```suggestion
	docker-compose exec api python -m app.fireweather_bot.noon_forecasts
```

api/pyproject.toml
```diff
@@ -38,6 +38,7 @@ matplotlib = "^3.2.2"
 pytest-xdist = "^1.33.0"
 rope = "^0.17.0"
 docker-compose = "^1.26.2"
+pgadmin4 = "^4.25"
```
andrea-williams, line 41: What do we use pgadmin for?

api/app/auth.py
```diff
@@ -28,4 +28,4 @@ async def authenticate(token: str = Depends(oauth2_scheme)):
             status_code=status.HTTP_401_UNAUTHORIZED,
             detail=detail,
             headers={'WWW-Authenticate': 'Bearer'},
-        )
+        ) from exception
```
Kyubinhan, line 31: Can you explain the difference between having this and not having?

api/app/db/models.py
```diff
@@ -126,13 +147,47 @@ class ModelRunGridSubsetPrediction(Base):
...
+    precip_valid = Column(Boolean, default=False, nullable=False)
+    precipitation = Column(Float, nullable=False)
+    ffmc = Column(Float, nullable=False, default=math.nan)
+    isi = Column(Float, nullable=False, default=math.nan)
+    fwi = Column(Float, nullable=False, default=math.nan)
+    created_at = Column(TZTimeStamp, nullable=False,
+                        default=datetime.datetime.now(tz=timezone.utc))
+
+
 class NoonForecasts(Base):
```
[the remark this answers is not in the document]
Sybrand: Yep - just merged and pushed.

api/pyproject.toml
Sybrand: pgadmin is the postgresql ui. not sure if it's the same team who maintains it, I think it used to be. It doesn't HAVE to be in as a dependancy, and we could drop it. It's easy enough to just pip install it local.
Now that I think about it -  I'm going to drop it. That and docker-compose - it's just be making the automated testing slower. I can pip install it local and it's fine.

api/app/auth.py
Sybrand: It makes for a better stack trace - as it includes the original exception.

api/pyproject.toml
Sybrand: Oh wait - never mind - we're not doing a --dev so it's fine - never mind.

This is only happening on our local machines, unless I'm missing something, so it's fine.

api/app/auth.py
Sybrand: It actually came up in linting when I upgraded our packages, I didn't know about it until before today.

api/app/fireweather_bot/hourly_actuals.py
Sybrand: @andrea-williams my problem is getting a pandas datetime with a timezone - I don't know how to get pandas to associate a timezone with something, and have it go all the way through to that final dictionary.

PR #470, opened 2020-09-02
Retrieve hourly readings from database.

Sybrand:
- Get hourly readings from database.
- Re-factor code: move choice of using wfwx outside of wfwx.
- Re-factor code: ensure all code is agnostic of using wfwx.
- I really dislike using all caps LOGGER


api/app/tests/test_get_hourlies.py
```diff
@@ -36,19 +67,19 @@ def response(monkeypatch, mock_env_with_use_wfwx, mock_jwt_decode, codes):
 
 
 # pylint: disable=redefined-outer-name
-@then('the response status code is <status>')
+@ then('the response status code is <status>')
```
Kyubinhan, line 70: I wonder why this happens in your code editor

api/app/hourlies.py
```diff
@@ -0,0 +1,63 @@
...
+        if station_readings is None or reading.station_code != station_readings.station.code:
+            station = next(
+                station for station in stations if station.code == reading.station_code)
+            station_readings = WeatherStationHourlyReadings(station=station, values=[])
+            result.append(station_readings)
+        weather_reading = WeatherReading(
+            datetime=reading.weather_date,
+            temperature=get(reading.temperature, reading.temp_valid),
+            relative_humidity=get(reading.relative_humidity, reading.rh_valid),
+            wind_speed=get(reading.wspeed_valid, reading.wind_speed),
```
andrea-williams, line 45: isn't this backwards? shouldn't it be get(reading.wind_speed, reading.wspeed_valid)?

api/app/main.py
```diff
@@ -155,7 +156,7 @@ async def get_hourlies(request: schemas.StationCodeList, _: bool = Depends(authe
     """ Returns hourlies for the last 5 days, for the specified weather stations """
     try:
         LOGGER.info('/hourlies/')
```
andrea-williams, line 158: `logger` ?

andrea-williams (changes requested):
Looks like there's an oopsie in api/app/hourlies.py

api/app/fireweather_bot/common.py
```diff
@@ -81,7 +82,9 @@ def get_station_names_to_codes() -> dict:
     """ Helper function to create dictionary of (station_name: station_code) key-value pairs
     Is used when replacing station names with station IDs in dataframe
     """
-    station_data = _get_stations_local()
+    loop = asyncio.new_event_loop()
+    asyncio.set_event_loop(loop)
+    station_data = loop.run_until_complete(get_stations())
```
Kyubinhan, line 87: Can you explain why it has to be done this way?

api/app/hourlies.py
Kyubinhan: Too bad Python is not strongly typed...

Kyubinhan (commented):
Overall looks good, I left some comments tho.
Question: do you prefer setting a specific query for fixture data in tests? I feel it would be more valuable if we just add mock data into the database and check if the query actually works

api/app/fireweather_bot/common.py
Sybrand: The API runs asynchronously, and when we make http requests from inside the API, we make them asynchronously as well. When we run the bot, it's not running asynchronously - and you can't just call async code from non async code. So that bit is just putting in a block on get_stations. If there's a simpler way to do it, I'd love to find out.
We used to call synchronous code from the bot, but I've re-factored it to not care if it's going against wfx (async) or local db (sync) - so I've had to introduce this bit.

api/app/main.py
Sybrand: I've been slowly making it lower case wherever I find it - but ok - I'll do main too :)

api/app/hourlies.py
Sybrand: omg - well spotted - i'm surprised it didn't complain, since condition is type.

api/app/tests/test_get_hourlies.py
Sybrand: So weird - I didn't want to fight it, and now it's stopped doing it.

andrea-williams (approved):

Kyubinhan:
> @Kyubinhan I started off wanting to do query specific fixture data - but it's not that straight forward. You have to match the filter EXACTLY or you get nothing. If you skip the filter, you get whatever is in the mock. I spent 10 minutes looking for how to construct a filter for an "select where in" style query, and couldn't find it - so I just gave up. I don't have much love for alchemy-mock.

What I meant was to not specify any query for the fixture, as an example, if you look at `test_noon_forecasts_summaries.py`, I only added data to the database, and checked if our query returned the correct data.

Kyubinhan (approved):

PR #475, opened 2020-09-04
Backend - pull most recent historic model for a weather station

andrea-williams:
Added API endpoint to retrieve the last issued historic model run for a requested weather station, from five days ago up to the current time. 
NOTE: SonarCloud is reporting test coverage just below 80%, but is also reporting that `get_most_recent_historic_station_model_predictions()` in app.db.crud has no unit tests, but it does (`test_most_recent_historic_predictions.py`). I suspect that because we're now mocking the crud call rather than inserting data into the DB, sonarcloud is confused.


api/app/models/fetch/predictions.py
```diff
@@ -13,7 +13,7 @@
                          WeatherModelPredictionValues, WeatherModelRun)
 from app.db.models import ModelRunGridSubsetPrediction
 import app.db.crud
-import app.stations
+from app.wildfire_one import get_stations_by_codes
```
Sybrand, line 16: Please revert this change! Let app.stations decide if it's going to go against wf1 or local - I think this may have been a merge conflict resolved incorrectly?

```suggestion
import app.stations
```

api/app/db/crud.py
```diff
@@ -8,13 +8,66 @@
...
     ProcessedModelRunUrl, PredictionModel, PredictionModelRunTimestamp, PredictionModelGridSubset,
-    ModelRunGridSubsetPrediction, NoonForecast, HourlyActual)
+    ModelRunGridSubsetPrediction, NoonForecast, HourlyActual, WeatherStationModelPrediction)
 import app.time_utils as time_utils
 
 logger = logging.getLogger(__name__)
 
 LATLON_15X_15 = 'latlon.15x.15'
 
+# --------------  COMMON UTILITY FUNCTIONS ---------------------------
```
Sybrand, line 18: This module is starting to get a bit busy - I won't complain if someone starts splitting it up....

api/app/models/fetch/predictions.py
```diff
@@ -182,7 +181,7 @@ async def _fetch_model_predictions_by_station_codes(model: ModelEnum, station_co
     """ Fetch predictions from database.
     """
     # Using the list of station codes, fetch the stations:
-    stations = await app.stations.get_stations_by_codes(station_codes)
+    stations = await get_stations_by_codes(station_codes)
```
Sybrand, line 184: please revert this bit.
```suggestion
    stations = await app.stations.get_stations_by_codes(station_codes)
```

Sybrand (commented):
General note re. unit testing if you start doing it - I'm going to move away from using sqlalchemy mock - and just start mocking out crud. Mocking out the db is starting to become more pain than it's worth.

api/app/models/fetch/predictions.py
```diff
@@ -192,6 +190,63 @@ async def _fetch_model_predictions_by_station_codes(model: ModelEnum, station_co
...
+    """ Fetch the most recent historic model predictions from database based on each station's coordinates. """
+    # construct temporary helper dictionary of relevant weather stations
+    stations_dict = defaultdict(list)
+    stations = await app.stations.get_stations_by_codes(station_codes)
+    for station in stations:
+        stations_dict[station.code] = station
+
+    # construct helper dictionary of WeatherModelPredictions
+    # weather_model_predictions_dict is indexed by station_code, then by prediction_timestamp
+    weather_model_predictions_dict = defaultdict(lambda: defaultdict(list))
```
Kyubinhan, line 208: This one too, the structure here would be defaultdict(lambda: defaultdict(WeatherModelPrediction))

api/app/models/fetch/predictions.py
```diff
@@ -192,6 +190,63 @@ async def _fetch_model_predictions_by_station_codes(model: ModelEnum, station_co
...
+    # send the query
+    session = app.db.database.get_session()
+    historic_predictions = app.db.crud.get_most_recent_historic_station_model_predictions(
+        session, station_codes, model, five_days_ago, now)
+    for prediction, prediction_model_run_timestamp, prediction_model in historic_predictions:
+        existing_prediction = weather_model_predictions_dict[prediction.station_code][
+            prediction.prediction_timestamp]
+        # insert the prediction into weather_model_predictions_dict if no entry for prediction.prediction_timestamp exists, or if
+        # prediction_model_run_timestamp.prediction_run_timestamp is more recent than the model_run time for the existing_prediction
+        if(existing_prediction == [] or existing_prediction.model_run.datetime < prediction_model_run_timestamp.prediction_run_timestamp):
```
Kyubinhan, line 222: existing_prediction is an object, so it should check if it's None right?

api/app/models/fetch/predictions.py
```diff
@@ -192,6 +190,63 @@ async def _fetch_model_predictions_by_station_codes(model: ModelEnum, station_co
...
-    """ Fetch 10 day global model weather predictions for a given station."""
+    """ Fetch model weather predictions for a given list of stations and a given model. """
     # Fetch predictions from the database.
     return await _fetch_model_predictions_by_station_codes(model, station_codes)
+
+
+async def _fetch_most_recent_historic_predictions_by_station_codes(model: ModelEnum, station_codes: List[int]) -> List[WeatherModelPrediction]:
+    """ Fetch the most recent historic model predictions from database based on each station's coordinates. """
+    # construct temporary helper dictionary of relevant weather stations
+    stations_dict = defaultdict(list)
```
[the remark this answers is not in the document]
andrea-williams: changed to stations_dict = {}

api/app/models/fetch/predictions.py
andrea-williams: changed to defaultdict(dict)

api/app/models/fetch/predictions.py
andrea-williams: am now checking if existing_prediction is None

api/app/db/crud.py
```diff
@@ -185,6 +243,17 @@ def get_prediction_model(session: Session, abbreviation: str, projection: str) -
         filter(PredictionModel.projection == projection).first()
 
 
+def get_prediction_model_run_timestamp_records(
+        session: Session, complete: bool = True, interpolated: bool = True):
+    """ Get prediction model run timestamps (filter on complete and interpolated if provided.) """
+    query = session.query(PredictionModelRunTimestamp)
+    if interpolated is not None:
+        query = query.filter(PredictionModelRunTimestamp.interpolated == interpolated)
+    if complete is not None:
```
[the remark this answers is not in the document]
andrea-williams: https://www.python.org/dev/peps/pep-0008/#programming-recommendations - using 'is not None' is recommended as per PEP8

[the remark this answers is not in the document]
Sybrand: @Kyubinhan also - reason I wrote it this way, is you get 3 different kinds of behaviour. If I specific "interpolated = True" - I can filter on when interpolation is True, if I specify "interpolated = False" I can filter on when interpolation is False, if I specify "interpolated = None", then I don't filter on those at all, and get all records regardless.

api/app/models/fetch/predictions.py
```diff
@@ -192,6 +190,68 @@ async def _fetch_model_predictions_by_station_codes(model: ModelEnum, station_co
...
-    """ Fetch 10 day global model weather predictions for a given station."""
+    """ Fetch model weather predictions for a given list of stations and a given model. """
     # Fetch predictions from the database.
     return await _fetch_model_predictions_by_station_codes(model, station_codes)
+
+
+async def _fetch_most_recent_historic_predictions_by_station_codes(model: ModelEnum, station_codes: List[int]) -> List[WeatherModelPrediction]:
+    """ Fetch the most recent historic model predictions from database based on each station's coordinates. """
+    # construct temporary helper dictionary of relevant weather stations
+    stations_dict = {}
```
Sybrand, line 201: This 1 line of code would replace those 4:
```suggestion
    stations = {station.code : station for station in await get_stations_by_codes(station_codes)}
```

api/app/models/fetch/predictions.py
```diff
@@ -192,6 +190,68 @@ async def _fetch_model_predictions_by_station_codes(model: ModelEnum, station_co
...
+    stations = await app.stations.get_stations_by_codes(station_codes)
+    for station in stations:
+        stations_dict[station.code] = station
+
+    # construct helper dictionary of WeatherModelPredictions
+    # weather_model_predictions_dict is indexed by station_code, then by prediction_timestamp
+    weather_model_predictions_dict = defaultdict(dict)
+
+    # We're only interested in the last 5 days
+    now = datetime.datetime.now(tz=datetime.timezone.utc)
```
Sybrand, line 211: Please use the one in time_utils - makes for easier deterministic unit test writing.
```suggestion
    now = app.time_utils.get_utc_now()
```

api/app/models/fetch/predictions.py
```diff
@@ -192,6 +190,68 @@ async def _fetch_model_predictions_by_station_codes(model: ModelEnum, station_co
...
+    five_days_ago = now - datetime.timedelta(days=5)
+    # send the query
+    session = app.db.database.get_session()
+    historic_predictions = app.db.crud.get_historic_station_model_predictions(
+        session, station_codes, model, five_days_ago, now)
+    print(historic_predictions)
```
Sybrand: Please drop the print (or make it a logger)

api/app/db/crud.py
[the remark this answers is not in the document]
Kyubinhan: Well, the context is a bit different. Both `interpolated` and `complete` are always boolean. Actually, these if statements are meaningless since they always have default values right?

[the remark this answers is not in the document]
Sybrand: It's not always boolean, I could call `get_prediction_model_run_timestamp_records(session, complete=None, interpolated=None)` to get all records.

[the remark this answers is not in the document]
andrea-williams: hmmm good question. Let's ask @Sybrand - he wrote this part

api/app/models/fetch/predictions.py
Kyubinhan, line 217: I believe this should go away?

andrea-williams: yup

api/app/models/fetch/predictions.py
```diff
@@ -192,6 +190,68 @@ async def _fetch_model_predictions_by_station_codes(model: ModelEnum, station_co
...
+    for prediction, prediction_model_run_timestamp, prediction_model in historic_predictions:
+        station_predictions = weather_model_predictions_dict.get(prediction.station_code)
+        existing_prediction = None
+        if station_predictions is not None:
```
Sybrand, line 220: station_predictions is always going to be "not None" because weather_model_predictions_dict will always return a dict due to `weather_model_predictions_dict = defaultdict(dict)`

Sybrand (changes requested):
Some comments on _fetch_most_recent_historic_predictions_by_station_codes - could we address some of those and I'll re-look?

api/app/db/crud.py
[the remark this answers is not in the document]
Kyubinhan: Okay so then we should change the params so that it won't set these two values True when you pass None, right?

api/app/models/fetch/predictions.py
```diff
@@ -192,6 +190,67 @@ async def _fetch_model_predictions_by_station_codes(model: ModelEnum, station_co
...
+            # construct the WeatherModelPrediction
+            weather_model_prediction = WeatherModelPrediction(
+                station=stations_dict[prediction.station_code],
+                model_run=model_run,
+                values=[prediction_value]
+            )
+            weather_model_predictions_dict[prediction.station_code][
+                prediction.prediction_timestamp] = weather_model_prediction
+    # flatten the nested dict into a regular list of WeatherModelPredictions
+    weather_predictions_response = [item for timestamp in list(
```
Sybrand, line 248: Can this be re-stated as a few separate statements? It breaks my head looking at it. thing for thing in list for thing in list for thing in list - BOOM.

api/app/db/crud.py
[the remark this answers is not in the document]
Sybrand: @Kyubinhan I'm not sure I understand the question. The values aren't set to True when you pass None. They default to True if you pass nothing. If you pass None, they're None?

api/app/models/fetch/predictions.py
andrea-williams: station_predictions will be None if there's no value for the prediction.station_code key in weather_model_predictions_dict

api/app/db/crud.py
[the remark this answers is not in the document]
Kyubinhan: Case is closed :)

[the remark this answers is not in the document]
andrea-williams: @Kyubinhan @Sybrand so we're happy with this as is?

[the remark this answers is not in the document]
Sybrand: I believe so.

api/app/models/fetch/predictions.py
Sybrand: Duh - sorry - brain fail.

Sybrand (approved):

Kyubinhan (approved):

PR #476, opened 2020-09-04
Serve static content

Sybrand:
- move api from / to ./api
- serve up static content from /
- serve up /config.js as template


api/app/frontend.py
```diff
@@ -0,0 +1,51 @@
...
+
+
+templates = Jinja2Templates(directory='static')
+
+
+async def get_config(request: Request):
+    """ Apply jinja template response to config.js
+    """
+    return templates.TemplateResponse(
+        "config.js",
```
Kyubinhan, line 36: I wonder if you need to create `config.js` file under `/static`

Sybrand: I suspect I might have to do that just to get the code coverage good.

Kyubinhan (commented):
Are you intended to merge this pr? if so, we would have to make changes on FE for url changes

Sybrand:
> Are you intended to merge this pr? if so, we would have to make changes on FE for url changes

Yep - there's a PR for the front end, and we'd have to change the env variables when deploying.

Kyubinhan (approved):

andrea-williams (approved):

PR #486, opened 2020-09-14
Task/less patroni warnings

Sybrand:
Temporary workaround for reducing false downtime messages.

### DIFF
--- a/api/app/health.py
+++ b/api/app/health.py
@@ -21,9 +21,11 @@ def patroni_cluster_health_check():
     }
     resp = requests.get(url, headers=header)
     resp_json = resp.json()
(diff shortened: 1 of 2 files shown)


Kyubinhan (approved):
I wonder if we should have a different message when there is no healthy pod.

andrea-williams (approved):

Sybrand:
> I wonder if we should have a different message when there is no healthy pod.

@Kyubinhan nothing actually consumes that text at the moment - but yes, we could have something else. Though it would convey the same message wouldn't it? 0/3 pods is working vs. no pods are working? The key things is that healthy is False.

Any suggestions? I'm up for it. (When we have a Prometheus, we can much more useful states, since we can handle non binary situations - notifying you when the server has issues, but it's still working ok).

PR #511, opened 2020-09-30
handle missing station

Sybrand:


api/app/fireweather_bot/noon_forecasts.py
```diff
@@ -46,6 +46,8 @@ def _parse_csv(temp_path: str):
     with open(temp_path, 'r') as csv_file:
         data_df = pd.read_csv(csv_file)
     station_codes = get_station_names_to_codes()
+    # drop any rows where 'display_name' is not found in the station_codes lookup:
+    data_df.drop(index=data_df[~data_df['display_name'].isin(station_codes.keys())].index, inplace=True)
```
andrea-williams, line 50: isn't this dropping all rows where the display_name is **in** station_codes ?

Kyubinhan: I am guessing `~` is a not operator in this context?

Kyubinhan (approved):

api/app/fireweather_bot/noon_forecasts.py
andrea-williams: doh! yes, `~` is not. completely missed that

andrea-williams (approved):

api/app/fireweather_bot/noon_forecasts.py
Sybrand: yeah - took me a while to figure out how to do a "is not in" on a dataframe.

PR #515, opened 2020-10-07
Fix marking complete issue in env_canada.py

Kyubinhan:
Fix mark_prediction_model_run_processed method in env_canada.py
also remove unintentional(I think) mutation & rename some variables

### DIFF
--- a/api/app/models/env_canada.py
+++ b/api/app/models/env_canada.py
@@ -88,23 +88,23 @@ def parse_env_canada_filename(filename):
     return info
 
 
-def adjust_model_day(now, hour) -> datetime:
+def adjust_model_day(now, model_run_hour) -> datetime:
     """ Adjust the model day, based on the current time.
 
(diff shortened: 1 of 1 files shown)


api/app/models/env_canada.py
```diff
@@ -181,8 +181,9 @@ def mark_prediction_model_run_processed(session: Session,
         year=now.year,
         month=now.month,
         day=now.day,
-        hour=hour, tzinfo=datetime.timezone.utc)
-    prediction_run_timestamp = adjust_model_day(prediction_run_timestamp, hour)
+        hour=now.hour, tzinfo=datetime.timezone.utc)
+    prediction_run_timestamp = adjust_model_day(prediction_run_timestamp, model_run_hour)
```
Kyubinhan, line 185: So the bug was `adjust_model_day` never did the job here because `prediction_run_timestamp`'s hour was set with `model_run_hour` leading to comparing the same thing

andrea-williams (approved):

PR #536, opened 2020-10-23
Matomo

Sybrand:
- Pulled in matomo build and deploy scripts from https://github.com/BCDevOps/matomo-openshift/
- Added matomo variables to environment and global config.
- Ensured that matomo variables are being parsed for static config.

### DIFF
--- a/openshift/matomo/openshift/templates/mariadb/mariadb-build.json
+++ b/openshift/matomo/openshift/templates/mariadb/mariadb-build.json
@@ -1,0 +1,82 @@
+{
+  "kind": "Template",
+  "apiVersion": "v1",
+  "metadata": {
+    "name": "${NAME}-imagestream-template",
+    "labels": {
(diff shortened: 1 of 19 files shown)


openshift/matomo/openshift/templates/mariadb/mariadb-build.json
```diff
@@ -0,0 +1,82 @@
+{
```
andrea-williams, line 1: does this file mean we're going to have a MariaDB server running as well as a Postgres server?

Sybrand: Yep. Only works with MySQL and MariaDB : https://matomo.org/faq/how-to-install/faq_55/

andrea-williams (approved):
Approved because I don't see any issues, but I can barely read bash scripts or Openshift templates

PR #552, opened 2020-11-03
Bender uptime

andrea-williams:


api/app/fireweather_bot/hourly_actuals.py
```diff
@@ -118,6 +120,9 @@ def main():
     except Exception as exception:
         # Exit non 0 - failure.
         logger.error('Failed to retrieve hourly actuals.', exc_info=exception)
+        rc_message = ':scream: Encountered error retrieving hourly actuals\n{}: {}'.format(
```
Sybrand, line 123: You'll need a unit test that asserts that send_rocketchat_notification is called when an exception is thrown and this section of code runs.

Sybrand (changes requested):
Need unit tests to to validate that rocketchat code is called...

Sybrand (approved):

Sybrand (changes requested):
@andrea-williams  - oh snap - just realised something! We need some kind of indication of what server the message is coming from!
Otherwise we may end up chasing our tails when things are failing in dev.
Can you please add that in as part of the message?

Sybrand (approved):

api/app/models/env_canada.py
```diff
@@ -535,19 +541,28 @@ def main():
...
-    # pylint: disable=broad-except
-    except Exception as exception:
+        process_models()
+    except CompletedWithSomeExceptions:
+        logger.warning('completed processing with some exceptions')
+        sys.exit(os.EX_SOFTWARE)
+    except Exception as exception:  # pylint: disable=broad-except
         # We catch and log any exceptions we may have missed.
         logger.error('unexpected exception processing', exc_info=exception)
+        rc_message = ':poop: Encountered error retrieving model data from Env Canada\n{}: {}'.format(
```
Kyubinhan, line 558: I think it would be better if you concatenate the hostname and exception inside of `send_rocketchat_notification`. And you pass the first part of the message and the exception obj as arguments

Kyubinhan (approved):
Left a nice-to-have comment, but overall looks great!

PR #556, opened 2020-11-05
Common Repo

Sybrand:
# Common Repo

## When reviewing
Most of /web is just a straigh up copy from the web repo. Some files have been entirely removed (e.g. openshift templates, github actions etc.). Some file have been merged or 'taken from' (sonarcloud files, github actions, README's etc.).

Notable files in web that have had changes made to them:
- web/public/index.html (got rid of config.js - was hard to make it work)
- web/src/utils/constants.ts
README's in general (they're not really perfect - let me know if you're not ok with them - I'd rather address them going forward)

## Summary of changes
- copied https://github.com/bcgov/wps-web to /web
- modified api/Dockerfile to build static web content, and moved to root
- renamed api/Dockerfile.dev to api/Dockerfile (updated corresponding docker-compose and makefile commands)
- renamed web/Dockerfile.dev to web/Dockerfile (updated corresponding docker-compose)
- created docker-compose.yml in root to start up web and api
- created docker-compose-api-noreload.yml specifically for running the api on mac
- modified imagestream build process by pre-generating base python image with gdal pre-installed
- modified integration workflow to run tests and coverage on web in addition to api
- modified sonarcloud settings to account for web and api coverage
- introduced caching of node modules to speed up build
- introduced caching of python modules to speed up build
- added react env. variables to api project
- replacing %variables% with {{jinja tags}} during npm build
- updated python dependancies
- cleaned up /web removing file and folders that are now redundant
- renamed application from wps-api to wps (in order to bubble up naming to https://wps-prod.pathfinder.gov.bc.ca/)
- removed all references to wps-api and wps-web where applicable.
- updated README's somewhat (more work to be done here!)
- changed backup configuration (less backups, since we don't have enough space, and removed fider) and updated documentation.

## NOTES
- A new patroni cluster will be created when going to production, and the database will have to be restored to it.
- The wps-web route will have to be manually pointed to the new instance.
- The backup process will have to be updated to point to the new database instance.


openshift/unicorn-base/docker/Dockerfile
```diff
@@ -1,15 +1,14 @@
...
-
-# Install dependancies need by python developer packages
-# NOTE: Once we no longer need pyodbc, please remove the apt-get update and install commands below.
-RUN apt-get -y update
-RUN apt-get -y install unixodbc-dev
 # Install old (2.4.*; current debian) version of gdal
+RUN apt-get -y update
 RUN apt-get -y install libgdal-dev
 
+# spotwx has an old certificate, so we have to make debian more forgiving.
```
andrea-williams, line 9: is this needed anymore, now that we're not using spotwx?

openshift/unicorn-base/openshift/build.yaml
```diff
@@ -0,0 +1,58 @@
...
+  phase: build
+  app.kubernetes.io/name: uvicorn-gunicorn-fastapi
+  app.kubernetes.io/managed-by: template
+  app.kubernetes.io/version: "python3.8"
+parameters:
+  - name: NAME
+    value: uvicorn-gunicorn-fastapi
+  - name: SUFFIX
+  - name: VERSION
+    description: Ouput version
```
andrea-williams, line 17:
```suggestion
    description: Output version
```

andrea-williams (approved):

andrea-williams:
There's a few code smells reported but I approved anyway. 
After I approved I noticed there's no "Details" link next to the second "Lint, Test with Coverage & SonarQube (3.8)" action, and the action doesn't appear to be progressing, just stuck waiting...?

Sybrand:
> There's a few code smells reported but I approved anyway.
> After I approved I noticed there's no "Details" link next to the second "Lint, Test with Coverage & SonarQube (3.8)" action, and the action doesn't appear to be progressing, just stuck waiting...?

Those code smells are present in wps-web, I viewed them as out of scope for this ticket.

The action without progress is just an artifact - I change the name of the action, so the old one is still hanging around on this PR.

openshift/unicorn-base/docker/Dockerfile
Sybrand: Thanks I missed updating the text. Wildfire was actually being hosted on an endpoint that also needed this fix - but I've gone and commented it out for now. When we get around to integrating with them again, we can put it back in we we need to.

openshift/unicorn-base/openshift/build.yaml
```diff
@@ -0,0 +1,58 @@
...
+labels:
+  app: ${NAME}${SUFFIX}
+  phase: build
+  app.kubernetes.io/name: uvicorn-gunicorn-fastapi
+  app.kubernetes.io/managed-by: template
+  app.kubernetes.io/version: "python3.8"
+parameters:
+  - name: NAME
+    value: uvicorn-gunicorn-fastapi
+  - name: SUFFIX
```
Kyubinhan, line 15: This should be required I guess?

.dockerignore
```diff
@@ -0,0 +1,3 @@
+**/.gitignore
```
Kyubinhan, line 1: So we need these at root and /web?

openshift/unicorn-base/openshift/build.yaml
Kyubinhan: Seems like doesn't have to be!

Sybrand: Nope - I didn't even use it. The suffix makes sense when we're differentiating between PR and prod builds in the way we build images. We're not actually even using the suffix here. Right now we've got: uvicorn-gunicorn-fastapi   docker-registry.default.svc:5000/auzhsi-tools/uvicorn-gunicorn-fastapi   python3.8-latest,python3.8-TLSv1.0 (so calling it uvicorn-gunicorn-fastapi, and doing the differentiating with tags)

.dockerignore
Sybrand: I'm honestly not sure. They're being handled differently on linux and on mac - which is causing me some confusion. I'm really not sure when they get used and how - I'll play around a bit. Hopefully we can get rid of one.

Kyubinhan (approved):

PR #558, opened 2020-11-09
Task/collect rdps/545

andrea-williams:


Sybrand (approved):
I think we should be able to use the same test code for GDPS, RDPS and HRDPS - but I'm fine with having them separate.

api/app/models/env_canada.py
```diff
@@ -108,15 +108,16 @@ def parse_env_canada_filename(filename):
...
     if model == 'glb':
         model, variable_name, projection, model_run_timestamp, prediction_timestamp = \
-            parse_global_model_filename(filename)
+            parse_gdps_rdps_filename(filename)
         model_abbreviation = ModelEnum.GDPS
     elif model == 'hrdps':
         model, variable_name, projection, model_run_timestamp, prediction_timestamp = \
             parse_high_res_model_filename(filename)
         model_abbreviation = ModelEnum.HRDPS
     elif model == 'reg':
```
Kyubinhan, line 117: Looks like this can be merged with the first if statement

Kyubinhan: Ah nvm

Kyubinhan (approved):

PR #566, opened 2020-11-17
Display RDPS data on graph

andrea-williams:
Note: commit history for this branch is on wps-web repo, branch `story/rdps_graph/537`


web/src/features/fireWeather/slices/modelsSlice.ts
```diff
@@ -2,8 +2,8 @@ import { createSlice, PayloadAction } from '@reduxjs/toolkit'
 
 import { ModelValue, getModelsWithBiasAdj, ModelsForStation } from 'api/modelAPI'
 import { AppThunk } from 'app/store'
-import { isNoonInPST } from 'utils/date'
 import { logError } from 'utils/error'
+import { parseModelValuesSlice } from './parseModelValuesSlice'
```
Kyubinhan, line 6: Can you use an absolute path here?

web/src/features/fireWeather/slices/parseModelValuesSlice.ts
```diff
@@ -0,0 +1,33 @@
...
+  const currDate = new Date()
+  allModelValues.forEach(v => {
+    const isFutureModel = new Date(v.datetime) >= currDate
+    if (isFutureModel) {
+      modelValues.push(v)
+    } else if (separate_noon_values && isNoonInPST(v.datetime)) {
+      noonModelValues.push(v)
+    } else {
+      pastModelValues.push(v)
+    }
```
Kyubinhan, line 24: If you look at the old `modelSlices` the logic is slightly different. We need to always include noon values in the `pastModelValues`

web/src/features/fireWeather/slices/parseModelValuesSlice.ts
```diff
@@ -0,0 +1,33 @@
+import { ModelRun, ModelValue } from 'api/modelAPI'
+import { isNoonInPST } from 'utils/date'
+
+export const parseModelValuesSlice = (
```
Kyubinhan, line 4: Can you change this name as well as the filename? this is technically a helper function, not a slice

Kyubinhan (changes requested):

web/src/utils/storybook.ts
```diff
@@ -60,6 +62,22 @@ const getPastValues = () => {
         rh_tgl_2_median: rh,
         rh_tgl_2_90th: rh - 3
       })
+      _pastRegionalModelValues.push({
+        datetime,
+        temperature: temp + (Math.random() - 0.7) * 7,
```
Sybrand, line 67: You could use a sine wave here to get a more natural curve going.

Sybrand (approved):
I will say - it's a disgusting amount of code to just add another layer, that's pretty much identical to the other layers we've added. We really need to have a more generic approach.

Also - I'll add in Matomo analytics for showRegionalModels

web/src/features/fireWeather/slices/parseModelValuesHelper.ts
```diff
@@ -0,0 +1,34 @@
...
+  const currDate = new Date()
+  allModelValues.forEach(v => {
+    const isFutureModel = new Date(v.datetime) >= currDate
+    if (isFutureModel) {
+      modelValues.push(v)
+    } else {
+      pastModelValues.push(v)
+      if (separate_noon_values && isNoonInPST(v.datetime)) {
+        noonModelValues.push(v)
+      }
```
Kyubinhan, line 24: Just realized this block has to be a separate if statement because we want to include future model values in `noonModelValues` as well

web/src/utils/storybook.ts
andrea-williams: `temperature` is based off `temp`, which uses the sine curve.

web/src/features/fireWeather/slices/parseModelValuesHelper.ts
andrea-williams: gotcha. Fixed in my latest commit

Kyubinhan (approved):

PR #567, opened 2020-11-17
updated location of coverage files

Sybrand:
- Updated location of coverage file for sonarcloud.
- Fixing the path inside coverage report.

(This is happening because I changed how poetry is used to get caching to work nicely)

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -104,6 +104,7 @@ jobs:
           source ~/.poetry/env
           poetry run coverage report
           poetry run coverage xml -o coverage-reports/coverage-report.xml
+          sed -i 's/filename="app/filename="api\/app/' coverage-reports/coverage-report.xml
 
       - name: Cache node modules
         uses: actions/cache@v2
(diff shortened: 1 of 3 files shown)


Kyubinhan (approved):

andrea-williams (approved):

.github/workflows/integration.yml
```diff
@@ -104,6 +104,7 @@ jobs:
           source ~/.poetry/env
           poetry run coverage report
           poetry run coverage xml -o coverage-reports/coverage-report.xml
+          sed -i 's/filename="app/filename="api\/app/' coverage-reports/coverage-report.xml
```
Kyubinhan, line 107: what does this do??

Kyubinhan (approved):

.github/workflows/integration.yml
Sybrand: Replace occurrences of 'filename="app' with 'filename="api/app" in coverage-report.xml - you want I should add a comment?

Kyubinhan: Oh I see, Yeah that would be nice

PR #569, opened 2020-11-17
Display observed & forecast precipitation graph

Kyubinhan:
* Implement precip graph component
* Remove husky hooks
* add `.storybook` under /web


andrea-williams:
would it be possible to create a bigger "target area" for the popup? I was just playing around with the graph in dev and I noticed that it's pretty difficult to get the cursor in just the right location to trigger the popup

web/.storybook/config.ts
```diff
@@ -0,0 +1,20 @@
+import { configure, addDecorator } from '@storybook/react'
+import { ThemeDecorator } from './decorators/ThemeDecorator'
+
+// @ts-ignore
+window.env = {
+  REACT_APP_FIDER_LINK: 'https://psufiderdev.pathfinder.gov.bc.ca/',
+  REACT_APP_KEYCLOAK_AUTH_URL: 'https://sso-dev.pathfinder.gov.bc.ca/auth',
```
andrea-williams, line 7: isn't this the wrong URL? I think this should be oidc.gov.bc.ca

web/src/features/fireWeather/components/graphs/PrecipGraph.tsx
```diff
@@ -0,0 +1,364 @@
...
+      '&--hidden': {
+        visibility: 'hidden'
+      }
+    }
+  }
+})
+
+interface PrecipValue {
+  date: Date
+  precip?: number
```
andrea-williams, line 61: could you please rename this to `observedPrecip` (assuming that's what it is?) I think it would be better to make it explicit, especially once we start adding to this with the different model types, etc.

web/src/features/fireWeather/components/graphs/PrecipGraph.tsx
```diff
@@ -0,0 +1,364 @@
...
+        .append('g')
+        .attr('class', 'legend')
+        .attr('transform', `translate(${margin.left}, ${legendMarginTop})`)
+
+      /* Create scales for x and y axes */
+      const xScale = d3
+        .scaleTime()
+        .domain(xDomain)
+        .range([0, chartWidth])
+      const xScaleCopy = xScale.copy()
```
andrea-williams, line 186: what's the purpose of making `xScaleCopy` rather than just using `xScale`?

andrea-williams (changes requested):
SSO pathfinder URL needs to be replaced with oidc

web/src/features/fireWeather/components/graphs/PrecipGraph.tsx
```diff
@@ -0,0 +1,364 @@
...
+        .attr('y', 0 - margin.left)
+        .attr('x', 0 - chartHeight / 2)
+        .attr('dy', '1.3em')
+        .attr('dx', '0')
+        .attr('class', 'yAxisLabel')
+        .text('Precipitation (mm/cm)')
+        .attr('transform', 'rotate(-90)')
+
+      /* Render legends */
+      // TODO: We're going to have to look at using layouts moving forward to achieve the placement of objects. https://www.d3indepth.com/layouts/
```
Sybrand, line 242: Oh man - here we go again! We really need to prioritise figuring out the layout.

Sybrand (changes requested):
Only issue I have is with the tooltip hover. Looks good otherwise.

web/.storybook/config.ts
Sybrand: Doesn't matter because it's for storybook - should be able to put any junk in ther eno?

Kyubinhan: Okay will put some junk link

web/src/features/fireWeather/components/graphs/PrecipGraph.tsx
Kyubinhan: Will do

web/src/features/fireWeather/components/graphs/PrecipGraph.tsx
Kyubinhan: This might be better to talk about it over a chat

Sybrand (approved):

andrea-williams (approved):

PR #592, opened 2020-11-25
Recognize station codes in URL

Kyubinhan:
* Get the app to react to the station code query present in the URL (both in Percentile Calc and MoreCast)
* Add Cache-Control header in /stations response
* Remove react-testing-library from `package.json` and its tests
* Write unit tests for util files
* Refactor env variable related stuff


web/src/utils/url.ts
```diff
@@ -0,0 +1,13 @@
+export const stationCodeQueryKey = 'codes'
+
+export const getStationCodesFromUrl = (search: string) => {
```
Sybrand, line 3: Add return type as per warning?

Sybrand (approved):

andrea-williams (approved):
Apart from GHA's warning, all good 👍

web/src/utils/url.ts
Kyubinhan: Done

PR #605, opened 2020-12-03
Story/load faster/420

Sybrand:
# Primary
- Switched to using pre-interpolated data instead of "raw" data.
- Fixed bug where summaries from one station would end up in the next when querying multiple stations. (added unit test for multiple stations).
- Change how unit tests get data (mocking out call to crud instead of using sqlalchemy mock).
- Increased memory use for production.

# Incidental
- Improved Makefile so you don't need different calls for when using poetry or not.
- Updated Zap scan version.
- Fixed some Zap warnings.
- Removed redundant endpoint /models/{model}/predictions/historic/most_recent/
- Temporarily disabled cronjobs.


andrea-williams (approved):
Sonarcloud is complaining about 1 code smell, but otherwise I'm happy. Performance is hugely improved

Kyubinhan (approved):

PR #606, opened 2020-12-04
Set default view for the temp & rh graph

Kyubinhan:
* Refactor hell out of TempRHGraph component
* Set the size bar position & size for T-2, T+2 view
* Dynamically set the temp y axis range
* Upgrade Keycloak type definition to match with the newer version of keycloak.js

### DIFF
--- a/web/src/features/fireWeather/components/graphs/TempRHGraph.tsx
+++ b/web/src/features/fireWeather/components/graphs/TempRHGraph.tsx
@@ -44,7 +44,7 @@ const TempRHGraph: React.FunctionComponent<Props> = (props: Props) => {
 
   const graphRef = useRef<SVGSVGElement>(null)
   const legendRef = useRef<SVGSVGElement>(null)
-  const gc = useMemoGraphCalculation(props, chartWidth)
+  const graphCalc = useMemoGraphCalculation(props, chartWidth)
 
   // Effect hook for displaying graphics
   useEffect(() => {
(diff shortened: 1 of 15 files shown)


web/src/features/fireWeather/components/graphs/TempRHGraph.tsx
```diff
@@ -5,280 +5,70 @@ import { ObservedValue } from 'api/observationAPI'
...
-  const svgRef = useRef(null)
+  const utcOffset = PDT_UTC_OFFSET
 
-  // Keep track of the sidebar position & size so that the sidebar can stay
-  // exactly as it is when toggling weather sources
-  // Use mutable ref object that doesn't cause re-render
-  const brushSelection = useRef<[number, number] | null>(null)
+  const graphRef = useRef<SVGSVGElement>(null)
+  const legendRef = useRef<SVGSVGElement>(null)
+  const gc = useMemoGraphCalculation(props, chartWidth)
```
andrea-williams, line 47: I'd prefer to see `gc` renamed to `graphCalc` or something similar to make it more obvious what it is.

andrea-williams (approved):

Sybrand (approved):

Sybrand (changes requested):
Looking at the ticket, "active default toggles should be Actuals and HRDPS"  but I don't get HRDPS selected.

Unrelated to the ticket, but since you've re-factored it - why do we default to +40 degrees celcius?

Sybrand (approved):

PR #607, opened 2020-12-08
updated container diagram

Sybrand:
Updated container diagram


Kyubinhan (approved):
In the Env Canada bot box, maybe add that we do some pre-calculations such as interpolation?

Sybrand:
> In the Env Canada bot box, maybe add that we do some pre-calculations such as interpolation?

@Kyubinhan have a look - I've updated it to say "Periodically fetch weather data, process and store subset" - I'd rather have something generic like "process" than go into detail.

andrea-williams (approved):

PR #635, opened 2020-12-14
Slow response time

Sybrand:
- Removed need for multiple calls to DB (in most cases, one can just look at the previous model run in memory).
- Fixed the query to only look at one station.
- Fixed the join to only join on the same model run.
- Modified unit test to allow for supporting multiple crud mocking.
- Added unit test for edge case.


Kyubinhan (approved):

api/app/weather_models/fetch/predictions.py
```diff
@@ -241,14 +241,23 @@ async def fetch_model_run_predictions_by_station_code(
     # Helper dictionary.
     station_predictions = defaultdict(dict)
 
+    # with open('get_station_model_predictions.json', 'w') as tmp:
+    # dump_sqlalchemy_response_to_json(historic_predictions, tmp)
```
andrea-williams, line 245: delete this?

andrea-williams (approved):
There's one commented-out code block that I suspect could be deleted, but otherwise all good. Takes approx. 3.5 seconds to load and render data for 3 stations 🚀

PR #643, opened 2020-12-18
Display observed wind speed & direction in a graph

Kyubinhan:
* Use Plotly to display wind speed and direction graph
* Allow 'unsafe-eval' in Content Security Policy to be able to use Plotly in production
* Update Node version to 12 in our integration GH action environment to fix `JavaScript heap out of memory` error.
* Fix an oversight on Percentile Calculator (Should have used a constant year - 2020 as the year when the calculation is done).


andrea-williams (commented):
Apart from the half-cropped "Now" in the time range slider and the shaded-out part when "all" is clicked, looks great to me!

Do you think we need a legend for the wind graph?

web/src/features/percentileCalculator/components/TimeRangeSlider.tsx
```diff
@@ -17,8 +17,7 @@ interface Props {
 }
 
 const MIN_YEARS = 0
-
-const MAX_YEARS = new Date().getFullYear() - 1970
+const MAX_YEARS = 50
```
Sybrand, line 20: Why this change? Won't this limit the percentile calculator to 2020? If we are limiting the percentile  calculator to 2020 (which I guess, since we only have data for that - we should) - I'd rather we be explicit about it and use 2020 in the code instead of 50, which seems arbitrary.

Sybrand (approved):
Soooo much nicer than doing it from scratch with d3. Looks really pretty Kyubin. Well done.

I've approved - but please look at the comment re. the magic constant 50 - I think we can do a better there if I understand it correctly.

andrea-williams (approved):

PR #646, opened 2020-12-22
Task/re factor api/268

Sybrand:
- Upgraded packages in poetry.lock.
- Bumped versions in pyproject.toml, and made versions less strict.
- Updated tests to work with latest version of pytest_bdd.
- Improved tests to not use eval, but rather use json.loads.
- Updated api documentation.


api/app/tests/weather_models/test_machine_learning.py
```diff
@@ -41,33 +43,26 @@ def given_an_instance(coordinate: List, points: List):
...
         model_temp: float, timestamp: datetime, bias_adjusted_temp: float):
     """ Assert that the ML algorithm predicts the temperature correctly """
-    result = given_an_instance.predict_temperature(model_temp, timestamp)
+    result = instance.predict_temperature(model_temp, timestamp)
     assert result == bias_adjusted_temp
-# pylint: enable=redefined-outer-name
 
 
-# pylint: disable=redefined-outer-name
 @ then('The <model_rh> for <timestamp> results in <bias_adjusted_rh>')
```
andrea-williams, line 63: since you deleted the spaces for @when, etc, just pointing out that the 'then's still have spaces

api/app/tests/weather_models/test_models_predictions_db.py
```diff
@@ -65,47 +70,40 @@ def mock_get_model_run_predictions(*args):
...
-    num_prediction_values = eval(num_prediction_values)  # pylint: disable=eval-used
-    assert len(given_a_database['response_json']['predictions']
-               [num_prediction_values['index']]['values']) == num_prediction_values['len']
+    assert len(context['response_json']['predictions'][num_prediction_values['index']]
+               ['values']) == num_prediction_values['len']
 
 
-@then('The <expected_response> is matched')
-def assert_response(given_a_database, expected_response):  # pylint: disable=redefined-outer-name
+@ then('The <expected_response> is matched')
```
andrea-williams, line 106: but then you added spaces here. So 🤷‍♀️. 
I'm not actually bothered by it either way

andrea-williams (approved):

api/app/tests/weather_models/test_machine_learning.py
Sybrand: Thanks - sometimes vscode and I fight.

api/app/tests/weather_models/test_models_predictions_db.py
Sybrand: I'm VERY bothered! :) Thanks.

api/app/tests/weather_models/test_machine_learning.py
andrea-williams: haha oops. Just realized I tagged someone.

PR #649, opened 2020-12-23
Task/re factor api 2/268

Sybrand:
- Changed LOGGER to logger.
- Added API routers.
- Renamed endpoints (hourlies -> observed, models -> weather_models, noon_forecasts -> forecasts/noon, noon_forecasts/summaries -> forecasts/noon/summaries).
- Switched to using ephemeral storage in dev. (using jq to strip out pvc, will have to point to Sybrand/openshift-launchpad-deployment@add-jq until upstream accepts PR)
- Loading json data in tests using common function.
- Removed /models/{model}/predictions/ (not being called by UI), and functions it was using all the way down to CRUD.


api/app/routers/weather_models.py
```diff
@@ -0,0 +1,78 @@
...
+    """ Returns 10 day noon prediction based on the specified model,
+    for the specified set of weather stations. """
+    try:
+        logger.info('/weather_models/%s/predictions/', model.name)
+        if model == ModelEnum.GDPS:
+            projection = ProjectionEnum.LATLON_15X_15
+        elif model == ModelEnum.HRDPS:
+            projection = ProjectionEnum.HIGH_RES_CONTINENTAL
+        else:
+            projection = None
```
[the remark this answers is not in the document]
Sybrand: Good catch - how is that even working?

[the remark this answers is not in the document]
andrea-williams: Indeed. I'm curious to know how unit tests passed when RDPS is non-existent

api/app/tests/__init__.py
```diff
@@ -1,12 +1,27 @@
...
 
 
+def _load_json_file(module_path: str, filename: str) -> dict:
+    """ Load json file given a module path and a filename """
+    dirname = os.path.dirname(os.path.realpath(module_path))
+    with open(os.path.join(dirname, filename)) as file_pointer:
+        return json.load(file_pointer)
+
+
+def load_json_file(module_path: str) -> Callable[[str], dict]:
```
Kyubinhan, line 18: The name of this method seems misleading. It's a function that returns a function right?

Sybrand: Any suggestions? If I call it load_json_file_function it remains confusing. I think it's clear with the return type that it's returning a function - but I agree it's a confusing function.

api/app/routers/weather_models.py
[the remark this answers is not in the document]
Sybrand: Nothing is actually calling this code - and the unit tests are just returning mocked data - I'm ripping it out.

api/app/tests/__init__.py
Sybrand: @Kyubinhan I don't know if renaming it has really made it more readable... have a look.

Kyubinhan: Sorry did you change the name? Anyway yeah let's just stick to the old one. Naming is hard

Sybrand: Yeah - I've reverted - adding curry to the mix just confuses things.

Kyubinhan (approved):

PR #661, opened 2021-01-06
Wind Speed and Direction - Model Data

Sybrand:
# Env. Canada & API
- Added wind speed and direction to model download & database.
- Using "closest point" (instead of interpolating)
- Interpolating noon value for GDPS.
- NOT wiring up machine learning (we're only doing temp and rh right now).
- Added tests to check interpolation.
- Updated some test data.
- Wired up to API response.
- Some minor refactoring.
- Removing some redundant/dead code.

# Front end
- All work was already in place.
- Added model run into UI.

# Other
- Double patroni memory in prod to 4Gi request, 8Gi limit (was 2Gi and 4Gi respectively) ; all the pods end up using more than 2Gi, and the leader is maxing out at 4 right now.
- Deleting env-canada pods before cleaning up database.


api/app/weather_models/__init__.py
```diff
@@ -55,6 +57,54 @@ def interpolate_between_two_points(  # pylint: disable=invalid-name
...
+
+    params:
+    :time_a: Time of the first bearing.
+    :time_b: Time of the 2nd bearing.
+    :target_Time: Time for which we desire a value.
+    :direction_a: Bearing (in degrees) at time_a.
+    :direction_b: Bearing (in degrees) at time b.
+    """
+    x_axis = (time_a.timestamp(), time_b.timestamp())
+    if abs(direction_a - direction_b) > 180:
```
Sybrand, line 72: Is there a nicer way to deal with obtuse angles? Anyone?

Kyubinhan: Can you explain what this condition is checking?

api/app/tests/weather_models/test_interpolate_bearings.feature
```diff
@@ -0,0 +1,15 @@
...
+        Then You get <result>
+        Examples:
+            | time_a                           | time_b                           | target_time                      | direction_a | direction_b | result             |
+            # Test acute angle:
+            | 2020-09-03T18:00:00.000000+00:00 | 2020-09-03T21:00:00.000000+00:00 | 2020-09-03T20:00:00.000000+00:00 | 10          | 20          | 16.666666666666668 |
+            # Test obtuse angle:
+            | 2020-09-03T18:00:00.000000+00:00 | 2020-09-03T21:00:00.000000+00:00 | 2020-09-03T20:00:00.000000+00:00 | 10          | 200         | 256.6666666666667  |
+            # Test obtouse angle, interpolation in the other direction:
+            | 2020-09-03T18:00:00.000000+00:00 | 2020-09-03T21:00:00.000000+00:00 | 2020-09-03T20:00:00.000000+00:00 | 200         | 10          | 313.3333333333333  |
+
```
Kyubinhan, line 15: Can you add another case here with two angles being bigger than 180?

api/app/weather_models/__init__.py
Sybrand: You want me to just move the comment on line 74 up?

api/app/tests/weather_models/test_interpolate_bearings.feature
andrea-williams: have you tested what happens when the two angles are equal?

api/app/weather_models/__init__.py
andrea-williams: I can't think of a nicer way. The logic of it makes sense to me

Sybrand: @Kyubinhan I added an explanation - check if it makes sense to you please.

api/alembic/versions/ad4f37763020_add_wind.py
```diff
@@ -0,0 +1,36 @@
...
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ##
+    op.add_column('model_run_grid_subset_predictions', sa.Column(
+        'wdir_tgl_10', postgresql.ARRAY(sa.Float()), nullable=True))
+    op.add_column('model_run_grid_subset_predictions', sa.Column(
+        'wind_tgl_10', postgresql.ARRAY(sa.Float()), nullable=True))
+    op.add_column('weather_station_model_predictions', sa.Column('wdir_tgl_10', sa.Float(), nullable=True))
+    op.add_column('weather_station_model_predictions', sa.Column('wind_tgl_10', sa.Float(), nullable=True))
```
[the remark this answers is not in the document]
Sybrand: It's confusing as hell - I agree - but I want to match up to the env canada names as much as possible to avoid even greater confusion.

api/app/weather_models/machine_learning.py
```diff
@@ -136,7 +136,8 @@ def _add_sample_to_collection(self,
                                   actual: HourlyActual,
                                   sample_collection: SampleCollection):
         """ Take the provided prediction and observed value, adding them to the collection of samples """
-        for model_key, sample_key in zip(MODEL_VALUE_KEYS, SAMPLE_VALUE_KEYS):
+        # TODO: add precip and wind speed/direction to SAMPLE_VALUE_KEYS
```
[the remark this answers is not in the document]
Sybrand: Yep - right now we're only doing temp/rh - and there's no-one I'm aware of asking for adding anything else. It would have to happen in a different ticket.

api/app/tests/weather_models/test_interpolate_bearings.feature
Sybrand: I've added a case for when they're equal.

openshift/scripts/oc_cleanup.sh
```diff
@@ -35,6 +35,7 @@ OC_CLEAN_FORECAST_CRONJOB="oc -n ${PROJ_DEV} ${DELETE_OR_GET} cronjob/bcfw-p1-fo
 OC_CLEAN_EC_GDPS_CRONJOB="oc -n ${PROJ_DEV} ${DELETE_OR_GET} cronjob/env-canada-gdps-${NAME_APP}-${SUFFIX}"
 OC_CLEAN_EC_HRDPS_CRONJOB="oc -n ${PROJ_DEV} ${DELETE_OR_GET} cronjob/env-canada-hrdps-${NAME_APP}-${SUFFIX}"
 OC_CLEAN_EC_RDPS_CRONJOB="oc -n ${PROJ_DEV} ${DELETE_OR_GET} cronjob/env-canada-rdps-${NAME_APP}-${SUFFIX}"
+OC_DELETE_EC_PODS="oc -n ${PROJ_DEV} get pods -o name | grep -E 'env-canada-(gdps|rdps|hrdps)-${NAME_APP}-${SUFFIX}' | xargs oc ${DELETE_OR_GET}"
```
Kyubinhan, line 38:
```suggestion
OC_DELETE_EC_PODS="oc -n ${PROJ_DEV} ${DELETE_OR_GET} pods -o name | grep -E 'env-canada-(gdps|rdps|hrdps)-${NAME_APP}-${SUFFIX}' | xargs oc ${DELETE_OR_GET}"
```

api/alembic/versions/ad4f37763020_add_wind.py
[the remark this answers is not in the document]
andrea-williams: oh I see. fair enough

andrea-williams (approved):

openshift/scripts/oc_cleanup.sh
Sybrand: Nope. The first command, will always be an oc get.
Step #1 - get a list of pods
Step #2 - filter down to only giving env canada pods for this pr
Step #3 - delete each pod (this is the DELETE_OR_GET)

web/src/features/fireWeather/slices/parseModelValuesHelper.ts
```diff
@@ -8,10 +8,15 @@ export const parseModelValuesHelper = (
   const pastModelValues: ModelValue[] = []
   const modelValues: ModelValue[] = []
   const noonModelValues: ModelValue[] = []
-  const allModelValues = model_runs.reduce(
-    (values: ModelValue[], modelRun) => values.concat(modelRun.values),
-    []
-  )
+  const reducer = (values: ModelValue[], modelRun: ModelRun) => {
```
Kyubinhan, line 11: Quick question about the response structure, why do we have them this way?

Kyubinhan: Oh is it because the model timestamp can vary for individual model values?

Sybrand: The structure from the API is
-model run
--predictions

It didn't make sense to repeat the model run information on every single prediction.

openshift/scripts/oc_cleanup.sh
Kyubinhan: Ah I can see `${DELETE_OR_GET}` at the end now

Kyubinhan (approved):

PR #662, opened 2021-01-06
Remove Fider stuff in the project

Kyubinhan:
Wanted to test `window.open('mailto:bcws.predictiveservice@gov.bc.ca')` but doesn't seem to be possible with Cypress


web/src/components/PageHeader.tsx
```diff
@@ -60,15 +60,13 @@ export const PageHeader: React.FunctionComponent<Props> = ({ title }: Props) =>
...
         </div>
-        <a
-          className={classes.contact}
-          href={FIDER_LINK}
-          target="_blank"
-          rel="noopener noreferrer"
+        <div
           id="contact-link"
+          className={classes.contact}
+          onClick={() => window.open('mailto:bcws.predictiveservice@gov.bc.ca')}
```
andrea-williams, line 66:
```suggestion
          onClick={() => window.open('mailto:bcws.predictiveservices@gov.bc.ca')}
```

andrea-williams (approved):
Typo in the mailto link, but otherwise all good

web/src/components/PageHeader.tsx
Kyubinhan: Ah I blame Sybrand for not having a consistent mail address in the ticket, nice catch!

andrea-williams: it's a good approach, always blaming @Sybrand . I hold him responsible for many of my failings and inadequacies

Sybrand, line 66: I think we should add a subject (maybe even a body?) - Something like "Predictive Services Unit - Website" - it's maybe too much trouble to add in "Percentile Calculator" or "Morecast" depending on where you are?
```suggestion
          onClick={() => window.open('mailto:bcws.predictiveservices@gov.bc.ca?subject=Predictive Services Unit - Website')}
```

Sybrand (approved):
Just a suggestion re. adding in a subject. At your discretion.

web/src/components/PageHeader.tsx
Kyubinhan: Okay sounds good I will add the broader subject

Kyubinhan: Change of mind. Turns out it's fairly easy to be more specific

Sybrand: Always happy to take the blame.

PR #680, opened 2021-01-08
fix pods delete

Sybrand:
Sorry - my cleanup code failed when there were no pods to delete.

I'm kinda breaking from the whole 1. create command variable 2. eval command variable 3. print variable methodology for deleting the cronjob pods - i just can't think of a nice way to get that done.

### DIFF
--- a/openshift/scripts/oc_cleanup.sh
+++ b/openshift/scripts/oc_cleanup.sh
@@ -30,12 +30,12 @@ else
 fi
 OC_CLEAN_DEPLOY="oc -n ${PROJ_DEV} ${DELETE_OR_GET} all,cm -o name -l app=${NAME_OBJ}"
 OC_CLEAN_TOOLS="oc -n ${PROJ_TOOLS} ${DELETE_OR_GET} all,cm -o name -l app=${NAME_OBJ}"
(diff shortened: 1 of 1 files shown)


openshift/scripts/oc_cleanup.sh
```diff
@@ -49,7 +49,12 @@ eval "${OC_CLEAN_FORECAST_CRONJOB}"
 eval "${OC_CLEAN_EC_GDPS_CRONJOB}"
 eval "${OC_CLEAN_EC_HRDPS_CRONJOB}"
 eval "${OC_CLEAN_EC_RDPS_CRONJOB}"
-eval "${OC_DELETE_EC_PODS}"
+
+pods=$(eval "${OC_GET_EC_PODS}")
+if [ -n "$pods" ]; then
+	echo $pods | xargs oc -n ${PROJ_DEV} ${DELETE_OR_GET}
+fi
```
Kyubinhan, line 56: How about https://stackoverflow.com/a/8296746/11903963

Sybrand: Because that actually stops the bash script from running. Very frustrating, somehow no-one else seems to have that problem. It's what I tried 1st.

Kyubinhan: I see I see. In the comment section it also says it doesn't work in Mac so nvm

Kyubinhan (approved):

andrea-williams (approved):

PR #700, opened 2021-01-25
C-Haines severity polygons

Sybrand:
# C-Haines 
Download files from env. Canada and generate C-Haines polygons for each model run prediction for an area around B.C.

- web/src/features/cHaines/pages/CHainesPage.tsx is PROOF OF CONCEPT - and not intended to be final. For the most, please disregard this code. It's horrible and needs to be replaced.
- take note of change to web/src/components/Container.tsx - which will affect all pages. (doesn't make sense to squish the map into a small space)
- separate ubuntu based imagestream for c-haines, since the debian imagestream doesn't provide the features needed.

# Related
- session context manager introduced to ensure that session are closed correctly.

# Other
- fixed issue with weather model missing index on geom.
- increased the readinessProbe period to 360 seconds (was defaulting to 10 seconds)


andrea-williams (commented):
Inserted some comments for the first part of my review - I only reviewed the first 12 files

.github/workflows/deployment.yml
```diff
@@ -126,12 +125,6 @@ jobs:
           AUTH_TOKEN: ${{ secrets.OC_TOOL_TOKEN_EDIT }}
           SCRIPT: GIT_BRANCH=${GITHUB_HEAD_REF} DOCKER_FILE="Dockerfile.ocp3" openshift/scripts/oc_build.sh ${SUFFIX} apply
 
-      - name: Deploy to Dev
```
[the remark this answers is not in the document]
Sybrand: correct - this should go away as part of your pr.

.github/workflows/integration.yml
```diff
@@ -63,6 +61,18 @@ jobs:
         run: |
           sudo apt-get update --fix-missing && sudo apt-get -y install libgdal-dev
 
+      - name: Install gdal python (api)
+        # We don't have much control over what version of gdal we're getting, it's pretty much whatever is
+        # available to us. As such, gdal is not installed by poetry, since the version will differ between
+        # platforms.
+        # Also have to be absolutely sure that numpy is installed before gdal (otherwise gdal doesn't install
+        # correctly, and you'll run into ndarray errors.)
```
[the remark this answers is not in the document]
Sybrand: This is a tricky one. Somehow, installing from the poetry cache wasn't good enough to get gdal to install correctly with numpy. (at least, that's why I think went wrong).
Hence the explicit install of numpy, followed by gdal - and then allowing install from cache and then doing the poetry install.
@andrea-williams  should I just add comments to that effect? It's real clunky - but I don't have a cleaner solution.

api/Makefile
```diff
@@ -48,13 +48,13 @@ test:
 	# Run tests
 	# useful params for pytest:
 	# -s show stdout
-	# -n numprocesses - doesn't really seem to run faster with more proccesses at the moment.
-	PYTHONPYCACHEPREFIX=python_cache $(POETRY_RUN) pytest app;
+	# -n numprocesses
+	PYTHONPYCACHEPREFIX=python_cache $(POETRY_RUN) pytest -n 3 app;
```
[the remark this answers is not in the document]
Sybrand: Nope - that's a python variable.

api/Makefile
```diff
@@ -48,13 +48,13 @@ test:
...
 	# ImportMismatchError? run: make clean
 
 test-verbose:
 	# Run tests with verbose output
-	$(POETRY_RUN) pytest app -vv;
+	$(POETRY_RUN) pytest app -n 2 -vv;
```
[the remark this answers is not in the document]
Sybrand: Coverage doesn't actually have -n ; coverage gets confused when you're running in parallel.

I'll change it to -n 3 ; 3 is just what I've found runs fastest for me. 1 (or not specifying parallel) has now just gotten too slow because of my c-haines test.

api/app/.env.example
```diff
@@ -33,4 +33,5 @@ ROCKET_AUTH_TOKEN=sometoken
 ROCKET_USER_ID=someid
 ROCKET_URL_POST_MESSAGE=https://somewhere/api/v1/chat.postMessage
 ROCKET_CHANNEL=somechannel
-PYTHONPYCACHEPREFIX=python_cache
\ No newline at end of file
+PYTHONPYCACHEPREFIX=python_cache
+MAX_AGE_C_HAINES=26
```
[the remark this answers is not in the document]
Sybrand: Suggestions? We probably want to be consistent. There are other things that will have retention ages. E.g. raw environment canada data. e.g. something like MAX_AGE_GDPS_RAW - so what do we do
MAX_WEEKS_C_HAINES=26
MAX_WEEKS_GDPS_RAW=26 ?

web/src/components/Container.tsx
```diff
@@ -4,5 +4,5 @@ import { Container as C, ContainerProps } from '@material-ui/core'
 export const Container: React.FunctionComponent<ContainerProps> = (
   props: ContainerProps
 ) => {
-  return <C maxWidth="md" {...props} />
+  return <C maxWidth="xl" {...props} />
```
Kyubinhan, line 7: Hmm, I don't know changing the component's default property is the right way. You could set `maxWidth` property when using the component.

.github/workflows/deployment.yml
```diff
@@ -225,6 +218,12 @@ jobs:
           AUTH_TOKEN: ${{ secrets.OC4_DEV_TOKEN }}
           SCRIPT: POSTGRES_WRITE_HOST=patroni-wps-${SUFFIX}-leader POSTGRES_READ_HOST=patroni-wps-${SUFFIX}-replica PROJ_TARGET="e1e498-dev" PROJ_TOOLS="e1e498-tools" POSTGRES_USER="wps" POSTGRES_DATABASE="wps" IMAGE_REGISTRY="image-registry.openshift-image-registry.svc:5000" PROJ_DEV="e1e498-dev" openshift/scripts/oc_provision_ec_rdps_cronjob.sh ${SUFFIX} apply
 
+      - name: C-Haines Cronjob
+        uses: bcgov/openshift-launchpad-deployment@v1.2
+        with:
+          AUTH_TOKEN: ${{ secrets.OC4_DEV_TOKEN }}
+          SCRIPT: PROJ_TARGET="e1e498-dev" PROJ_TOOLS="e1e498-tools" PROJ_DEV="e1e498-dev" openshift/scripts/oc_provision_c_haines_cronjob.sh ${SUFFIX} apply
+
```
Kyubinhan, line 226: Do you need to run this twice? here and after building linux image?

Sybrand: Definitely not - thanks for spotting that. Will fix.

api/app/.env.example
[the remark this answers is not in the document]
andrea-williams: yep I like MAX_WEEKS_...

openshift/unicorn-base/openshift/build.yaml
```diff
@@ -21,7 +21,7 @@ parameters:
   - name: GIT_BRANCH
     value: main
 objects:
-  #ImageStream is create if it doesn't already exist
+  #ImageStream is createed if it doesn't already exist
```
conbrad, line 24: Minor: createed -> created

openshift/templates/c_haines.cronjob.yaml
```diff
@@ -0,0 +1,119 @@
...
+  - name: IMAGE_REGISTRY
+    description: Image registry for imagestream.
+    value: image-registry.openshift-image-registry.svc:5000
+    required: true
+  - name: IMAGE_NAME
+    required: true
+    # value: ${NAME}-${SUFFIX}
+  - name: IMAGE_TAG
+    required: true
+    # value: ${SUFFIX}
```
andrea-williams, line 25: did you intend to leave the default values for NAME, IMAGE_TAG, and IMAGE_NAME commented out?

openshift/templates/c_haines.cronjob.yaml
```diff
@@ -0,0 +1,119 @@
...
+                    - name: ROCKET_USER_ID
+                      valueFrom:
+                        secretKeyRef:
+                          name: ${GLOBAL_NAME}-global
+                          key: rocket.chat-user-id-secret
+                    - name: ROCKET_AUTH_TOKEN
+                      valueFrom:
+                        secretKeyRef:
+                          name: ${GLOBAL_NAME}-global
+                          key: rocket.chat-auth-token-secret
```
andrea-williams, line 118: the Rocketchat-related config and secret variables haven't been configured in OCP4 yet - just a heads up in case you were planning on relying on these notifications right now

openshift/ubuntu-base/openshift/build.yaml
```diff
@@ -0,0 +1,58 @@
...
+  - name: NAME
+    value: ubuntu-base
+  - name: SUFFIX
+  - name: VERSION
+    description: Output version
+    value: "20.04"
+  - name: GIT_URL
+    value: https://github.com/bcgov/wps.git
+  - name: GIT_BRANCH
+    value: feature/c-haines-re-factor
```
andrea-williams, line 22: what happens when this branch gets deleted post-merge?

openshift/unicorn-base/openshift/build.yaml
andrea-williams, line 24: classic Sy... introducing a new typo while trying to fix an old typo ;)

sonar-project.properties
```diff
@@ -19,11 +19,15 @@ sonar.javascript.lcov.reportPaths=web/coverage/lcov.info
 # api/alembic/** - database migrations shouldn't have unit tests.
 # api/app/tests/** - test code doesn't have tests.
 # api/app/db/crud.py - unit tests don't talk to a database.
+# api/app/db/database.py - this should mostly be stubbed out, and not ever be covered by unit tests.
 # web - we don't need coverage reports on stories or tests
+# web/src/features/cHaines/** - the UI is alpha, if we're going to keep it, it needs a massive re-factor.
```
andrea-williams, line 24: I don't think this comment still applies, does it?

conbrad (changes requested):

api/app/c_haines/fetch.py
```diff
@@ -0,0 +1,273 @@
...
+    """ Add kml <Style> tag """
+    kml.append('<Style id="{}">'.format(style_id))
+    kml.append('<LineStyle>')
+    kml.append('<width>1.5</width>')
+    kml.append('<color>{}</color>'.format(color))
+    kml.append('</LineStyle>')
+    kml.append('<PolyStyle>')
+    kml.append('<color>{}</color>'.format(color))
+    kml.append('</PolyStyle>')
+    kml.append('</Style>')
```
andrea-williams, line 81: Oof. Writing HTML in python one line at a time. That's rough.

api/app/c_haines/fetch.py
```diff
@@ -0,0 +1,273 @@
...
+        end = model_run_timestamp + timedelta(hours=1)
+    kml = []
+    kml.append('<LookAt>')
+    kml.append('<gx:TimeSpan>')
+    kml.append(f'<begin>{model_run_timestamp.isoformat()}</begin>')
+    kml.append(f'<end>{end.isoformat()}</end>')
+    kml.append('</gx:TimeSpan>')
+    kml.append('<longitude>-123</longitude>')
+    kml.append('<latitude>54</latitude>')
+    kml.append('<range>3000000</range>')
```
andrea-williams, line 109: (a) what is "range"? (b) why are there hard-coded lat/long values here?

api/app/c_haines/fetch.py
```diff
@@ -0,0 +1,273 @@
...
+    # Iterate through the records we get from the database.
+    for poly, severity, prediction_timestamp in result:
+        # If it's a new timestamp.
+        if prediction_timestamp != prev_prediction_timestamp:
+            # Close the previous placemark if needed.
+            if not prev_severity is None:
+                yield close_placemark()
+            prev_severity = None
+            # Close the previous folder if needed.
+            if not prev_prediction_timestamp is None:
```
andrea-williams, line 128: I think this would be easier to read if it were `if prev_prediction_timestamp is not None`

api/app/c_haines/fetch.py
```diff
@@ -0,0 +1,273 @@
...
+                yield f'{FOLDER_CLOSE}\n'
+            prev_prediction_timestamp = prediction_timestamp
+            # Start a new folder.
+            yield f'{FOLDER_OPEN}\n'
+            yield '<name>{} {} {}</name>\n'.format(model, model_run_timestamp, prediction_timestamp)
+        # Close the placemark if needed.
+        if severity != prev_severity:
+            if not prev_severity is None:
```
andrea-williams, line 136: same comment as above - `if prev_severity is not None:` ?

web/src/features/cHaines/slices/cHainesModelRunsSlice.tsx
```diff
@@ -0,0 +1,189 @@
...
+  model_run_timestamp: string
+  prediction_timestamp: string
+  result: FeatureCollection
+}
+
+const initialState: State = {
+  loading: false,
+  error: null,
+  model_runs: [],
+  selected_model_abbreviation: 'GDPS',
```
andrea-williams, line 29: why is this set to GDPS by default?

api/app/c_haines/severity_index.py
```diff
@@ -0,0 +1,453 @@
...
+
+    def __init__(self, projection, geotransform, rows: int, cols: int):
+        self.projection = projection
+        self.geotransform = geotransform
+        self.rows: int = rows
+        self.cols: int = cols
+
+
+def create_in_memory_band(data: numpy.ndarray, source_info: SourceInfo):
+    """ Create an in memory data band """
```
andrea-williams, line 194: This function documentation is a bit vague - could you describe what a data band is?

api/app/db/models/c_haines.py
```diff
@@ -0,0 +1,85 @@
...
+from app.db.models.common import TZTimeStamp
+
+
+class SeverityEnum(enum.Enum):
+    """ Enumerated values for severity
+    """
+    LOW = "<4"
+    MODERATE = "4-8"
+    HIGH = "8-11"
+    EXTREME = ">11"
```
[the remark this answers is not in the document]
Sybrand: 8.0 would be high.

So really what's happening here is:
```
    LOW = "c-haines <4"
    MODERATE = "4 <= c-haines < 8"
    HIGH = "8 <= c-haines < 11"
    EXTREME = "11 < c-haines"
```

@andrea-williams  what do you think - what would be a better way to get the message across without the little white lie, and without confusing people?

api/app/c_haines/severity_index.py
```diff
@@ -0,0 +1,453 @@
...
+        model run, and prediction timestamp. """
+
+        def _download_files(urls: dict,
+                            model: ModelEnum,
+                            tmp_path: str) -> Union[List[str], None]:
+            """ Try to download all the files """
+            filenames = []
+            for key in make_model_levels(model):
+                # Try to download this file.
+                filename = download(urls[key], tmp_path)
```
andrea-williams, line 346: it's not immediately clear whether `tmp_path` is short for "temporary path" or something related to temperature...?

api/app/c_haines/severity_index.py
```diff
@@ -0,0 +1,453 @@
...
+        return 'CMC_hrdps_continental_{level}_ps2.5km_{date}{HH}_P{hhh}-00.grib2'.format(
+            level=level, date=date, HH=model_run_start, hhh=forecast_hour)
+    raise UnhandledPredictionModelType()
+
+
+def make_model_levels(model: ModelEnum):
+    """ Return list of layers. (The layers are named slightly differently for HRDPS) """
+    if model == ModelEnum.HRDPS:
+        return ['TMP_ISBL_0700', 'TMP_ISBL_0850', 'DEPR_ISBL_0850']
+    return ['TMP_ISBL_700', 'TMP_ISBL_850', 'DEPR_ISBL_850']
```
andrea-williams, line 147: could you document what variables these are referring to? (specifically, I'm not sure what DEPR is)

web/src/features/cHaines/pages/CHainesPage.tsx
```diff
@@ -0,0 +1,736 @@
+import React, { useRef, useEffect, useState } from 'react'
+import ReactDOMServer from 'react-dom/server'
+import { selectCHainesModelRuns } from 'app/rootReducer'
+import { useDispatch, useSelector } from 'react-redux'
+import 'leaflet/dist/leaflet.css'
+import L from 'leaflet'
```
andrea-williams, line 6: are we sticking with leaflet or will this be refactored to use OpenLayers at some point in the future?

openshift/templates/c_haines.cronjob.yaml
Sybrand: Removed it.

web/src/features/cHaines/pages/CHainesPage.tsx
```diff
@@ -0,0 +1,736 @@
...
+        'https://openmaps.gov.bc.ca/geo/pub/WHSE_LAND_AND_NATURAL_RESOURCE.PROT_WEATHER_STATIONS_SP/ows?',
+        {
+          format: 'image/png',
+          layers: 'pub:WHSE_LAND_AND_NATURAL_RESOURCE.PROT_WEATHER_STATIONS_SP',
+          styles: 'BC_Wildfire_Active_Weather_Stations_Labels',
+          transparent: true,
+          minZoom: 0,
+          maxZoom: 18
+        }
+      )
```
andrea-williams, line 220: do we need to add attributions for the sources of these layers?

andrea-williams (changes requested):
I was very harsh in this review... sorry about that.

The only red flag for me is the potential for SQL injection in your hand-crafted SQL queries

openshift/templates/c_haines.cronjob.yaml
Sybrand: @andrea-williams weird - I'm seeing it, and deployment of other cronjobs shouldn't be possible if it's not working. Am I missing something?

openshift/ubuntu-base/openshift/build.yaml
Sybrand: Nothing - this build config is run manually to get the ubuntu base image in place. But I have updated it to remove confusion.

api/app/db/models/c_haines.py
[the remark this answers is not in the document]
andrea-williams: It would be succinct to write `MODERATE = [4,8)` and `HIGH = [8,11)` but I'm not sure if that's too math-y.

sonar-project.properties
Sybrand: This section document the reason behind excluding code coverage on some parts of the code - so it will remain relevant until such time as the c-haines front end code has been cleaned up.

api/app/db/models/c_haines.py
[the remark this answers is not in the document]
andrea-williams: It's probably fine to leave it as is for now - I was mostly just concerned that it could cause logical problems in the code if it wasn't specified how 8.0 would be handled, but now that I've read the rest of your code I'm satisfied that it won't be an issue on the code side of things.

openshift/templates/c_haines.cronjob.yaml
andrea-williams: Whoops. I was looking in TEST, where it hasn't been configured. Everywhere else it has been. Silly me.

api/app/c_haines/c_haines_index.py
```diff
@@ -0,0 +1,182 @@
...
+    """
+    # pylint: disable=invalid-name
+
+    # Temperature depression term (this indicates atmospheric instability).
+    # Temperature at 850mb - Temperature at 700mb.
+    ca = (t850-t700)/2-2
+    # Dew point depression term (this indicates how dry the air is).
+    cb = d850/3-1
+
+    # This part limits the extent to which dry air is able to affect the overall index.
```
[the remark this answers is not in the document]
Sybrand: Essentially yes, though probably more accurate to say that the temperature and dew point depression terms are the domain? I've added some more comments, and links to the source material where available to increase clarity.

api/app/c_haines/c_haines_index.py
```diff
@@ -0,0 +1,182 @@
...
+    """ Get the geographical area (the bounding box) that we want to generate data for.
+    Currently hard coded to be an area around B.C. """
+    # S->N 46->70
+    # E->W -110->-140
+    top_left_geo = (-140.0, 70.0)
+    bottom_right_geo = (-110.0, 46.0)
+    return (top_left_geo, bottom_right_geo)
+
+
+class BoundingBoxChecker():
```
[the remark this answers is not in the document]
Sybrand: @conbrad https://www.python.org/dev/peps/pep-0020/

api/app/tests/weather_models/test_env_canada_gdps.py
```diff
@@ -3,12 +3,16 @@
...
+from typing import Generator
+from contextlib import contextmanager
 import datetime
 from datetime import datetime
 import pytest
 import requests
 import shapely.wkt
+from sqlalchemy.orm import Session
 from geoalchemy2.shape import from_shape
+# TODO: get rid of alchemy mock!
```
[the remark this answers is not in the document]
Sybrand: Yep - issue 651 https://app.zenhub.com/workspaces/wildfire-predictive-services-5e321393e038fba5bbe203b8/issues/bcgov/wps/651

openshift/templates/c_haines.cronjob.yaml
```diff
@@ -0,0 +1,119 @@
...
+                          key: app-db-password
+                    - name: POSTGRES_WRITE_HOST
+                      value: ${POSTGRES_WRITE_HOST}
+                    - name: POSTGRES_READ_HOST
+                      value: ${POSTGRES_READ_HOST}
+                    - name: POSTGRES_PORT
+                      value: "5432"
+                    - name: POSTGRES_DATABASE
+                      value: ${POSTGRES_DATABASE}
+                    - name: MAX_AGE_C_HAINES
```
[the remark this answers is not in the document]
Sybrand: I went for MAX_WEEKS_....

api/app/c_haines/c_haines_index.py
```diff
@@ -0,0 +1,182 @@
...
+    # instability, that dry air isn't going anywhere.
+    if cb > 9:
+        cb = 9
+    elif cb > 5:
+        cb = 5 + (cb-5)/2
+
+    # Combine the two terms for the index.
+    ch = ca + cb
+
+    return ch
```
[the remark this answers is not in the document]
Sybrand: @andrea-williams super good point Andrea. I've added citations. I've been going off random things emailed to me and found on the internet thus far. This is a (small) can of worms, since our code doesn't exactly match the source.

api/app/c_haines/c_haines_index.py
```diff
@@ -0,0 +1,182 @@
...
+
+
+def read_scanline(band, yoff):
+    """ Read a band scanline, returning an array of values. """
```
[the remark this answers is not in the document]
Sybrand: I've added links to reference, and given a short description of what a band and a scanline is in the context.

api/app/c_haines/fetch.py
```diff
@@ -0,0 +1,273 @@
...
+
+def _yield_folder_parts(result, model: ModelEnum, model_run_timestamp: datetime) -> Iterator[str]:
+    """ Yield up all the different parts of the folders with placemarks.
+    """
+    prev_prediction_timestamp = None
+    prev_severity = None
+    # Iterate through the records we get from the database.
+    for poly, severity, prediction_timestamp in result:
+        # If it's a new timestamp.
+        if prediction_timestamp != prev_prediction_timestamp:
```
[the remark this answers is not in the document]
Sybrand: I only care that the timestamp has changed, in practice it happens to be ordered, but as long as the timestamps are grouped, this code would work regardless of sequence.

api/app/tests/c_haines/test_c_haines.py
```diff
@@ -0,0 +1,206 @@
...
+        'tmp_850': get_complete_filename(__file__, tmp_850),
+        'dew_850': get_complete_filename(__file__, dew_850)}
+
+
+@when("We generate c-haines")
+def generate_c_haines(collector):
+    """ Generate c-haines data using grib files. """
+    with open_gdal(collector['tmp_700'], collector['tmp_850'], collector['dew_850']) as gdal_data:
+        generator = CHainesGenerator()
+        # Generate c-haines data. It's pretty slow.
```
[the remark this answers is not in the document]
Sybrand: One could generate a grib file that contains a smaller subset of data yes - but then I'd have to figure out how to do that. Also, the code for selecting what area of the grib file to use may change, end then I'd have to re-create that.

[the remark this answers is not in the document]
conbrad: Ok was not aware how grib files were generated.

api/app/db/models/c_haines.py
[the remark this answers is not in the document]
Sybrand: I've lumped this into a ticket for future consideration.

api/app/c_haines/c_haines_index.py
```diff
@@ -0,0 +1,182 @@
...
+
+class BoundingBoxChecker():
+    """ Class used to check if a given raster coordinate lies within a specified
+    geographic bounding box. """
+
+    def __init__(self, padf_transform, raster_to_geo_transformer):
+        self.geo_bounding_box = get_geographic_bounding_box()
+        self.padf_transform = padf_transform
+        self.raster_to_geo_transformer = raster_to_geo_transformer
+        self.good_x = dict()
```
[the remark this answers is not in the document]
Sybrand: It's bad code - I've changed the code to use pythons built in memoization (from functools import lru_cache)

api/app/c_haines/c_haines_index.py
```diff
@@ -0,0 +1,182 @@
...
+            (raster_x, raster_y),
+            self.padf_transform,
+            self.raster_to_geo_transformer)
+        lon0 = self.geo_bounding_box[0][0]
+        lat0 = self.geo_bounding_box[0][1]
+        lon1 = self.geo_bounding_box[1][0]
+        lat1 = self.geo_bounding_box[1][1]
+        if lon0 < lon < lon1 and lat0 > lat > lat1:
+            good_y = self.good_x.get(raster_x)
+            if not good_y:
```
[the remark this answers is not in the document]
Sybrand: See above. Ripped out this horrid code, and rather using lru_cache (should have used an entirely different pattern for this!)

api/app/c_haines/fetch.py
Sybrand: a) Distance from point (added a comment) b) fairly arbitrary position that gives you a nice view of B.C. in google earth (added comment).

api/app/c_haines/fetch.py
Sybrand: My initial thought was to use some kml or xml library - but doing it this way is considerably faster since having a streaming response is much easier to do this way.

api/app/db/crud/c_haines.py
```diff
@@ -0,0 +1,217 @@
...
+    logger.info('deleted %s model runs', model_runs)
+
+
+def get_most_recent_model_run(session: Session, model: ModelEnum) -> CHainesModelRun:
+    """ Return the most recent model run for which we have at least one prediction. """
+    return session.query(CHainesModelRun)\
+        .join(PredictionModel, PredictionModel.id == CHainesModelRun.prediction_model_id)\
+        .join(CHainesPrediction, CHainesPrediction.model_run_id == CHainesModelRun.id)\
+        .filter(PredictionModel.abbreviation == model)\
+        .order_by(desc(CHainesModelRun.model_run_timestamp)).limit(1).first()
```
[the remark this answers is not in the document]
Sybrand: Nope. limit() ensures the database only gives back one result, first() ensures we get a single object instead of some collection.

web/src/features/cHaines/pages/CHainesPage.tsx
Sybrand: There's a ticket for switching to OpenLayer.

web/src/features/cHaines/pages/CHainesPage.tsx
Sybrand: Leaflet automatically pulls in attribution and throws it in the bottom right corner.

api/app/c_haines/severity_index.py
Sybrand: Changed variable name.

Sybrand:
> I was very harsh in this review... sorry about that.
> 
> The only red flag for me is the potential for SQL injection in your hand-crafted SQL queries

Not at all - thank you for so diligently wading through this ridiculously large PR.

andrea-williams (approved):
Apart from the residual typo in the api/app/tests/c_haines/endpoints/test_c_haines_endpoints_crud_mapings.json filename that Conor pointed out (it looks like I now have some competition in my spelling error sniffing!), I think it should all be good now.

Kyubinhan (approved):
I wonder if it's worth it to give up `immutableCheck` & `serializableCheck` to include C-haines data in Redux store (I get that these checks are only included in development but still). As a suggestion maybe when you get a chance to refactor, try this approach? https://reactjs.org/docs/hooks-reference.html#usereducer

conbrad (approved):
There's some unresolved comments that may have been missed (probably Github auto-collapsing them), mostly minor spelling. Most important comment is probably this one: https://github.com/bcgov/wps/pull/700#discussion_r597955568

Otherwise looks good!

api/app/c_haines/severity_index.py
```diff
@@ -0,0 +1,453 @@
...
+    dst_ds.FlushCache()
+    # Explicitly clean up (is this needed?)
+    del dst_ds, data_ds, mask_ds
+
+
+def save_geojson_to_database(session: Session,
+                             source_projection, filename: str,
+                             prediction_timestamp: datetime,
+                             model_run: CHainesModelRun):
+    """ Open geojson file, iterate through features, saving them into the
```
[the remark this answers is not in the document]
Sybrand: In this case, I don't actually want geojson the database - I want geometry objects. Allows us to do all kinds of fun GIS things supported by PostGIS (select features in area, reduce polygon complexity etc. etc.) - currently only using it to generate the KML.
What I'm envisioning though, is throwing all this away - and storing geojson and kml in an object store - and not having much of anything in the database.

api/app/tests/__init__.py
```diff
@@ -1,18 +1,27 @@
...
-    """ Load json file given a module path and a filename """
+def get_complete_filename(module_path: str, filename: str):
+    """ Get the full path of a filename, given it's module path """
     dirname = os.path.dirname(os.path.realpath(module_path))
-    with open(os.path.join(dirname, filename)) as file_pointer:
-        return json.load(file_pointer)
+    return os.path.join(dirname, filename)
+
+
+# TODO: Replace Union[dict, None] with dict | None when python >= 3.10
```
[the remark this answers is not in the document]
Sybrand: There isn't - but - it's not like it's going to stop working - it's just a nicer syntax that's being introduced in 3.10;

web/src/features/cHaines/slices/cHainesModelRunsSlice.tsx
Sybrand: Good question - I'll throw it at people in standup.

Sybrand:
> There's some unresolved comments that may have been missed (probably Github auto-collapsing them), mostly minor spelling. Most important comment is probably this one: [#700 (comment)](https://github.com/bcgov/wps/pull/700#discussion_r597955568)
> 
> Otherwise looks good!

My most humble apologies - addressing them all.

web/src/features/cHaines/slices/cHainesModelRunsSlice.tsx
Sybrand: Changed to HRDPS - but there's other code in the UI that overrides it. Have added this into a ticket.

api/app/c_haines/severity_index.py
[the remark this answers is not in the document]
Sybrand: @conbrad oh wait - I also misunderstood part of the question I think. So - part of the crazy sequence of events  is due to what gdal can and can't do. For example - I can't create the geojson in memory, it only knows how to write it to disk.

api/app/tests/c_haines/test_bounding_box.py
```diff
@@ -0,0 +1,42 @@
...
+    dirname = os.path.dirname(os.path.realpath(__file__))
+    dataset = gdal.Open(os.path.join(dirname, grib_file), gdal.GA_ReadOnly)
+    crs = CRS.from_string(dataset.GetProjection())
+    return {
+        'padf_transform': dataset.GetGeoTransform(),
+        'raster_to_geo_transformer': get_transformer(crs, GEO_CRS)
+    }
+
+
+@then('We expect the coordinate <x_coordinate> <y_coordinate> to be <is_inside>')
```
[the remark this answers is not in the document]
Sybrand: We do actually have that - if you look at the gherkin file - api/app/tests/c_haines/test_bounding_box.feature, two of the scenarios is_inside is False, one of them it's True.

api/app/c_haines/severity_index.py
```diff
@@ -0,0 +1,453 @@
...
+        return 1
+    # 8 - 11 : high
+    if c_haines_index < 11:
+        return 2
+    # 11 + Extreme
+    return 3
+
+
+def get_severity_string(severity: int) -> str:
+    """ Return the severity level as a string, e.g. severity level 3 maps to "11+"
```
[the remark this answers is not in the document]
conbrad: I was thinking of changing the method to accept a `SeverityEnum` so we could just reuse that instead of switching on integers, but then I followed the call chain up to see how the c-haines data is evaluated and it only knows numbers so yeah it's a bit trickier -- feel free to disregard!

api/app/tests/c_haines/test_bounding_box.py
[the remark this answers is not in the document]
conbrad: Ah thought that was just a documentation file! My bad. Reading up on gherkin now...

Sybrand:
> I wonder if it's worth it to give up `immutableCheck` & `serializableCheck` to include C-haines data in Redux store (I get that these checks are only included in development but still). As a suggestion maybe when you get a chance to refactor, try this approach? https://reactjs.org/docs/hooks-reference.html#usereducer

I've commented it out - when doing c-haines development one can always just switch it off. Added a TODO: to investigate changing the approach.

PR #719, opened 2021-02-03
Show weather data based on given time of interest

Kyubinhan:
* Create a separate form component for requesting weather data with time of interest time picker
* Rename weather related variables
* Set PST in all the wx display


api/app/schemas/shared.py
```diff
@@ -0,0 +1,11 @@
...
+"""
+
+from typing import List
+from pydantic import BaseModel
+
+
+class WeatherDataRequest(BaseModel):
+    """ A request for weather data for a given set of stations with a time of interest. """
+    stations: List[int]
+    time_of_interest: str = None
```
Sybrand, line 11: How about defaulting to now as the time of interest? Then you don't need all that code checking for null and inserting it.
```suggestion
    time_of_interest: str = get_utc_now().isoformat()
```

api/app/main.py
```diff
@@ -96,20 +98,29 @@ async def get_health():
...
 
 @api.post('/observations/', response_model=schemas.observations.WeatherStationHourlyReadingsResponse)
-async def get_hourlies(request: schemas.stations.StationCodeList, _: bool = Depends(authenticate)):
+async def get_hourlies(request: schemas.shared.WeatherDataRequest, _: bool = Depends(authenticate)):
     """ Returns hourly observations for the last 5 days, for the specified weather stations """
     try:
         logger.info('/observations/')
-        readings = await hourlies.get_hourly_readings(request.stations)
+
+        if request.time_of_interest is not None:
```
Sybrand, line 117: See my comment in WeatherDataRequest - I think we can get rid of the 4 lines of code that are repeated a few times.

Sybrand (changes requested):
Could you look at moving get_utc_now into the schema? Also - I wonder about the "time of interest" label. Have you checked if that makes sense with Dana? Is he ok with that wording for "now" and for a point in the past?

api/app/schemas/shared.py
Kyubinhan: So the problem with this is that mocking `get_utc_now` doesn't work at all here. I couldn't figure out why so I went the non-elegant way. Thought?

web/src/utils/date.ts
```diff
@@ -1,6 +1,6 @@
 import moment from 'moment'
 
-import { PST_UTC_OFFSET, PDT_UTC_OFFSET } from 'utils/constants'
+import { PST_UTC_OFFSET, PDT_UTC_OFFSET } from './constants'
```
andrea-williams, line 3: I thought we were always using absolute paths for imports?

web/src/features/fireWeather/components/graphs/plotlyHelper.ts
```diff
@@ -379,18 +393,14 @@ export const getDailyAndAccumPrecips = (values: PrecipValue[]) => {
     })
 
     Object.entries(aggregatedPrecips).forEach(([formattedDate, totalPrecip]) => {
-      const midNightOfTheDay = moment(formattedDate)
-        .set({ hour: 0 })
-        .toDate()
+      const midNightOfTheDay = getMidnightDate(formattedDate)
```
andrea-williams, line 396:
```suggestion
      const midnightOfTheDay = getMidnightDate(formattedDate)
```

web/src/features/fireWeather/components/graphs/plotlyHelper.ts
```diff
@@ -379,18 +393,14 @@ export const getDailyAndAccumPrecips = (values: PrecipValue[]) => {
...
       dates.push(midNightOfTheDay)
       dailyPrecips.push(totalPrecip)
     })
   } else {
     values.forEach(({ datetime, total_precipitation }) => {
       if (total_precipitation !== undefined) {
-        const midNightOfTheDay = moment(datetime)
-          .set({ hour: 0 })
-          .toDate()
+        const midNightOfTheDay = getMidnightDate(datetime)
```
andrea-williams, line 403:
```suggestion
        const midnightOfTheDay = getMidnightDate(datetime)
```

andrea-williams (approved):
Some minor changes requested, but otherwise I'm happy. Although I agree with Sy that it would be neater to set the timeOfInterest default value to now rather than having to check for None in a bunch of different places

Sybrand (approved):
I'm happy now - just check the "time of interest" wording with Dana/users.

web/src/utils/date.ts
Kyubinhan: Yeah... but when Cypress references this file, it doesn't like absolute paths

PR #722, opened 2021-02-08
Add disclaimer for time of interest picker

Kyubinhan:
Forgot to add a disclaimer for the time of interest time picker.


Sybrand (approved):

web/src/features/fireWeather/components/TimeOfInterestPicker.tsx
```diff
@@ -26,6 +26,7 @@ const TimeOfInterestPicker = (props: Props) => {
       label="Time of Interest (PST-08:00)"
       type="datetime-local"
       value={formatDateInPST(props.timeOfInterest, 'YYYY-MM-DDTHH:mm')}
+      helperText="Disclaimer: all data may not be available."
```
andrea-williams, line 29: I think what you mean is "not all data may be available" ?

Kyubinhan: Okay will change that

andrea-williams (approved):

PR #730, opened 2021-02-10
Database deployment (pull request and deploy to production)

Sybrand:
- Deploy DB in OCP4 in addition to OCP3.
- Deploy DB in OCP4 via Jenkins, in addition to OCP3.
- Cleanup after closing pr, in both OCP3 and OCP4.
- Changed patroni image to match https://github.com/bcgov/patroni-postgres-container.
- Switched to using health check from bcgov image - so no longer doing disk free check with message to rocket chat.
- Added postgis extension (1st step) as part of alembic migration.


openshift/patroni-postgis/README.md
```diff
@@ -1,27 +1,40 @@
...
+
+#### Common build failures
+
+```text
+E: Version '3.1.0+dfsg-1.pgdg90+1' for 'postgresql-11-postgis-3' was not found
+E: Version '3.1.0+dfsg-1.pgdg90+1' for 'postgresql-11-postgis-3-scripts' was not found
+```
+
+The latest version of the POSTGIS_VERSION available on debian is constantly changing. It is very likely, that when you attempt to build the image,
+it will fail because the version of postgis has changed. The most sure fire way of establishing the which version to us is to:
```
andrea-williams, line 29:
```suggestion
it will fail because the version of postgis has changed. The most sure fire way of establishing which version to use is to:
```

openshift/templates/patroni.yaml
```diff
@@ -353,31 +305,3 @@ parameters:
...
-    value: 5Gi
-  - description: Version of Patroni image to be used.
-    displayName: Patroni image version
-    name: IMAGE_VERSION
-    value: "v10"
-  - description: Pod management policy. For faster startups, use Parallel.
-    displayName: Pod management policy.
-    name: POD_MANAGEMENT_POLICY
-    value: "Parallel"
-  - description: Rocket.Chat channel to use for notifications.
```
andrea-williams, line 372: Have you configured notifications in Silver to warn us when we're approaching our disk space limit? I seem to recall reading a while ago that this was one of the things that OCP4 allows that OCP3 doesn't

andrea-williams (approved):
LGTM, although for the most part I don't know what I'm looking at

openshift/templates/patroni.yaml
Sybrand: @andrea-williams I've decided to just forego it entirely right now, in preference to us just using the bcgov image, no questions asked. (They just fail the health check at 90% - which would be 22.5GB for us right now).
I remember some talk of Prometheus being available with OCP4 - I don't know where that stands.

Kyubinhan (approved):

PR #736, opened 2021-02-11
Task/deploy to dev/623

andrea-williams:
- API+Front end deployment in OCP4.
- Health endpoint updated to work with new OCP4 api endpoint.
- OCP4 database now provisioning with fixed username and password "wps".
- Patroni PostGIS image modified to include creation of postgis extension. (PostGIS installation removed from alembic revision.)
- Poetry packages updated (causing slight change in output).


andrea-williams (commented):
Approving this PR, but GitHub won't let me approve it because I'm the official author of it.

Sybrand (approved):

PR #737, opened 2021-02-12
switch to knp

Sybrand:
- IN PROGRESS: Switch patroni postgres cluster to KNP.
- TODO: nsp for jenkins


Sybrand:
Too many resources being used

PR #741, opened 2021-02-17
Matomo migration

Kyubinhan:
* Separate secret objects from the templates
* Write down Matomo migration guide
* Update matomo url in .env.example


andrea-williams (approved):

openshift/matomo/openshift/templates/matomo-db/matomo-db-deploy.json
```diff
@@ -22,7 +23,8 @@
           "name": "${NAME}",
           "app": "${APP_NAME}",
           "service": "${NAME}",
-          "template": "matomo-template"
+          "template": "matomo-template",
+          "role": "${NAME}"
```
Sybrand, line 27: We're getting a bit tag heavy. I feel like we can drop template, and if we're adding role, we can drop service.

Nothing is actually using these tags right now, so it's just adding noise.

Also see below comment re. ${ROLE} - we should probably be using that instead of ${NAME}

openshift/matomo/openshift/templates/matomo-db/matomo-db-deploy.json
```diff
@@ -294,35 +274,11 @@
...
-      "description": "Password for the MariaDB connection user.  Needs to be base64 encoded/",
-      "required": true,
-      "generate": "expression",
-      "from": "[a-zA-Z0-9]{10}"
-    },
-    {
-      "name": "MYSQL_ROOT_PASSWORD",
-      "displayName": "MariaDB Admin Password",
-      "description": "Password for the MariaDB administrative account.  Needs to be base64 encoded.",
+      "name": "ROLE",
```
Sybrand, line 277: The role label is actually using ${NAME}

openshift/matomo/openshift/templates/matomo/matomo-deploy.yaml
```diff
@@ -130,28 +127,6 @@ objects:
               volumeMounts:
                 - mountPath: /var/www/html
                   name: matomo-source
-  - kind: NetworkSecurityPolicy
```
Sybrand, line 133: Can you maybe add a # TODO : add KNP in here ?

Sybrand (approved):
Some comments re. non use of ${ROLE} variable, and maybe having too many labels - can you please address before merging?

PR #742, opened 2021-02-18
Task/new table view/723

Kyubinhan:
Change table titles and use Accordian component

Test: https://wps-pr-742.pathfinder.gov.bc.ca/morecast


andrea-williams (approved):
You might want to check in with Ambrose on these changes (specifically that the tables are now collapsed by default) before merging into prod, since it might affect his usability testing next week (?)

web/src/features/fireWeather/components/WxDataDisplays.tsx
```diff
@@ -104,23 +104,23 @@ export const WxDataDisplays = React.memo(function _(props: WxDataDisplaysProps)
               <ErrorBoundary>
                 <ObservationTable
                   testId={`observations-table-${code}`}
-                  title="Past 5 days of hourly observations from station: "
+                  title="Past 5 days hourly observation table"
```
Sybrand, line 107: As per discussion - please make sure this wording is OK with UX and with PO.

Sybrand (approved):
Please check with UX and P.O. re. text changes, and starting collapsed. If they're OK with it - then this has my approval.

ambroseyau (approved):

PR #744, opened 2021-02-19
Fix time of interest date picker issue

Kyubinhan:
* Store time of interest in string instead of date
* Initialize time of interest in PST

Test: https://wps-pr-744.pathfinder.gov.bc.ca/morecast


web/src/features/fireWeather/components/TimeOfInterestPicker.tsx
```diff
@@ -37,7 +36,7 @@ const TimeOfInterestPicker = (props: Props) => {
       onChange={e => {
         const value = e.currentTarget.value
         if (value) {
-          props.onChange(new Date(value))
+          props.onChange(`${value}:00-08:00`) // Append seconds and timezone at the end
```
Sybrand, line 39: Is this going to work in all locales?

web/src/features/fireWeather/components/TimeOfInterestPicker.tsx
```diff
@@ -23,11 +22,11 @@ const TimeOfInterestPicker = (props: Props) => {
   return (
     <TextField
       data-testid="time-of-interest-picker"
+      className={classes.datePicker}
       label="Time of Interest (PST-08:00)"
       type="datetime-local"
-      value={formatDateInPST(props.timeOfInterest, 'YYYY-MM-DDTHH:mm')}
+      value={props.timeOfInterestISO.slice(0, 16)} // 'YYYY-MM-DDTHH:mm'
```
andrea-williams, line 28: what if we sliced this at (0, 13) instead? Then we could get rid of the minutes selection, which doesn't have any effect anyway

web/src/features/fireWeather/components/TimeOfInterestPicker.tsx
andrea-williams: is there a nice straightforward way to determine a user's timezone/T offset based on their OS or browser or something like that?

Kyubinhan: At this point yes, because that value doesn't contain the timezone information. But this is actually a valid concern when initializing the time of interest. I need to make a change in `getTimeOfInterestFromUrl`

andrea-williams (commented):
I tested out the UI and wasn't able to find any bugs or break it :)

web/src/features/fireWeather/components/TimeOfInterestPicker.tsx
Kyubinhan: Unfortunately, the datepicker component doesn't accept `YYYY-MM-DDTHH` format

web/src/features/fireWeather/components/TimeOfInterestPicker.tsx
Kyubinhan: @andrea-williams there is https://stackoverflow.com/questions/1091372/getting-the-clients-time-zone-and-offset-in-javascript
In this case tho we are fixating the timezone to PST

andrea-williams (approved):

web/src/utils/url.ts
```diff
@@ -12,7 +13,11 @@ export const getTimeOfInterestFromUrl = (search: string): string => {
     return queryString
   }
 
-  return formatDateInISO(new Date())
+  const currInPST = moment(new Date())
```
Sybrand, line 16:
```suggestion
  const currentDateInPST = moment()
```
Maybe I'm being pedantic in naming of the variable. But also - you don't need to pass new Date - see https://momentjs.com/docs/#/parsing/now/.

Sybrand, line 16: I would point at that moment recommends that projects shouldn't use it anymore:  https://momentjs.com/docs/#/use-it/

web/src/utils/url.ts
```diff
@@ -12,7 +13,11 @@ export const getTimeOfInterestFromUrl = (search: string): string => {
...
+    .utcOffset(PST_UTC_OFFSET)
```
Sybrand, line 17: Am I the only one who finds this confusing? Moments documentation didn't clear anything up for me. If we're going to keep moment, maybe add a comment as to what's happening here - something along the lines of "Take the current date, re-stating it in terms of pacific time (the date/time doesn't change, only the timezone). If you're already in PST, nothing changes."

Sybrand (approved):
I've added a bunch of comments. It does look like we should NOT be using moment - so maybe now is a good time to evaluate the alternatives? I'll leave that up to you, but I think that if a developer asks you not to use it's library - that's a pretty strong reason not to.

web/src/utils/url.ts
Kyubinhan: huh good to know

PR #745, opened 2021-02-19
Add zap scan to OCP4 pipeline

andrea-williams:
- Add Zap scan to Github actions when deploying in OCP4
- Modify rules.tsv to match URLs for OCP4

### DIFF
--- a/.github/workflows/zap_scan.yml
+++ b/.github/workflows/zap_scan.yml
@@ -7,7 +7,7 @@ on:
 jobs:
   zap_scan:
     runs-on: ubuntu-latest
-    name: Scan the webapplication
+    name: Scan the web application in OCP3
     steps:
       - name: Checkout
         uses: actions/checkout@v2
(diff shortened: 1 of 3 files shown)


Kyubinhan (approved):

.github/workflows/zap_scan.yml
```diff
@@ -19,3 +19,17 @@ jobs:
         with:
           target: "https://wps-prod.pathfinder.gov.bc.ca/"
           rules_file_name: ".zap/rules.tsv"
+  zap_scan:
```
Sybrand, line 22: Job "id" (in this case zap_scan) - is supposed to be unique: https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_id

Sybrand (approved):
@andrea-williams please change the job id to be unique - otherwise, all good.

PR #753, opened 2021-02-24
fix matomo url 

Sybrand:
weird - how did this every work, and why does it conflict with what the matomo site gives?


Sybrand:
The problem was with the server, not the client.

PR #761, opened 2021-03-01
Wind Speed+Direction -> Forecast

Sybrand:
- Display wind speed and direction in forecast graph.
- Modified max_user_watches to get rid of warning.

NOTE: Fails on sonarcloud - but we can't test clicks on the plotly graph, so there's no way to resolve that.
NOTE: You'll have to test this locally with storyboard or full db dump - there's no forecast data being created right now.


andrea-williams (approved):
I didn't bother testing out the UI because I'm too lazy to go to the trouble of importing the full DB locally, so I'll take your word for it that it works. Code review passes.

web/src/features/fireWeather/components/graphs/WindGraph.tsx
```diff
@@ -142,7 +156,8 @@ const WindGraph = (props: Props) => {
           gdpsData.windSpdLine,
           rdpsData.windSpdLine,
           hrdpsData.windSpdLine,
-          observationData.windSpdLine
+          observationData.windSpdLine,
+          forecastData.windSpdLine
```
Kyubinhan, line 160: Ordering actually matters here. By putting observation at last it will appear on top of all the graphics, so can you change the order here and down below for arrows??

Kyubinhan (changes requested):
You can actually increase the coverage by checking the number of arrows displayed on the graph.
Have a look at `Wind speed & direction graph should be displayed` in `morecast-page.spec.ts`

web/src/features/fireWeather/components/graphs/WindGraph.tsx
Sybrand: I also forgot to check for min and max values - I've added that too.

Kyubinhan (approved):
Appreciate the additional commits!

PR #775, opened 2021-03-06
Add dewpoint observations to table & graph

andrea-williams:


web/src/utils/storybook.ts
```diff
@@ -194,7 +196,7 @@ const getFutureValues = () => {
           datetime,
           temperature: temp + (Math.random() - 0.5) * 8,
           bias_adjusted_temperature: temp + (Math.random() - 0.5) * 8 - 2,
-          dew_point,
+          dewpoint,
```
Sybrand, line 199: Do we not also want the random variation here?

web/src/features/fireWeather/components/graphs/TempRHGraph.tsx
```diff
@@ -10,12 +10,14 @@ import { ToggleValues } from 'features/fireWeather/components/graphs/useGraphTog
...
   getLayoutConfig,
   populateGraphDataForTempAndRH,
+  TempRHGraphProperties,
   populateTimeOfInterestLineData,
   rangeSliderConfig
 } from 'features/fireWeather/components/graphs/plotlyHelper'
 
 const observedTempColor = '#fb0058'
 const observedRHColor = '#005f87'
+const observedDewpointColor = '#02fa44'
```
Sybrand, line 20: This seems a bit close to HRDPS RH? I'll defer to @ambroseyau on that.

web/cypress/integration/morecast-page.spec.ts
```diff
@@ -112,7 +112,7 @@ describe('MoreCast Page', () => {
     })
 
     it('Temp & RH Graph should be displayed', () => {
-      const checkNumOfTempMarkers = (num: number) => {
+      const checkNumOfTempDewpointMarkers = (num: number) => {
```
Kyubinhan, line 115: I am not sure if combining these two is the right way. We only display dew points for Observations right?

web/src/features/fireWeather/components/graphs/plotlyHelper.ts
```diff
@@ -66,29 +66,36 @@ interface TempRHValue {
...
   relative_humidity?: number | null
+  dewpoint?: number | null
   tmp_tgl_2_5th?: number
   tmp_tgl_2_90th?: number
   rh_tgl_2_5th?: number
   rh_tgl_2_90th?: number
-  tmp_max?: number
-  tmp_min?: number
+  tmp_dp_max?: number
+  tmp_dp_min?: number
```
Kyubinhan, line 75: Don't we need `tmp_max` and `tmp_min` for Forecast?

web/src/features/fireWeather/components/tables/ObservationTable.tsx
```diff
@@ -27,6 +27,12 @@ export const columns: Column[] = [
     align: 'right',
     format: (value: number): number => Math.round(value)
   },
+  {
```
Sybrand, line 30: Adding this has caused the date to wrap to two lines - I think we're better off not having anything wrap.

Sybrand (commented):
Can we please look at the color & date text wrapping?

Kyubinhan (commented):
I notice some changes have been made in `yarn.lock` which is odd because I don't see `package.json` being modified in this PR. Could you help me understand what happened?

web/src/utils/storybook.ts
andrea-williams: This is in the section for GDPS values, which is unrelated to this ticket. I just renamed the variable for consistency here, but it would probably be less confusing if I just deleted it altogether. At this point we're not using dewpoint for GDPS at all, so it's not necessary

web/src/features/fireWeather/components/graphs/plotlyHelper.ts
Kyubinhan: Well, the variable names are confusing. It implies that it has something to do with dew points no? also there is no `tmp_dp_max` property in any weather data type, which means line 173 `if (tmp_dp_min && tmp_dp_max) {` will never get executed

Kyubinhan: I should note that `tmp_max` and `tmp_min` are from ForecastSummary which are used to draw vertical lines. And I see why we didn't see any tests failing since we are not testing this part.

web/src/features/fireWeather/components/tables/ObservationTable.tsx
andrea-williams: Fixed!

web/src/features/fireWeather/components/graphs/plotlyHelper.ts
```diff
@@ -170,9 +170,9 @@ export const populateGraphDataForTempAndRH = (
     }
 
     // From forecast min & max
-    if (tmp_dp_min && tmp_dp_max) {
-      tempDewpointMinMaxDates.push(date)
-      tempDewpointMinMaxValues.push([tmp_dp_min, tmp_dp_max])
+    if (tmp_min && tmp_max) {
```
conbrad, line 173: Is it considered an error if either or both of these values are missing? Or should it be logged for transparency?

Kyubinhan: They are actually being used if you time travel to the time where forecasts are recorded. The toggles are disabled if there is no data to display, so it depends on the situation. 
Check this out: https://wps-web-prod.pathfinder.gov.bc.ca/morecast?codes=322&toi=2020-09-10T10:31:00-08:00

conbrad (approved):

Kyubinhan (approved):

Sybrand (approved):
Looks fine - still that colour issue, but if Ambrose has a better colour for us, we can always make a 1 pointer for that.

PR #790, opened 2021-03-11
Story/map in morecast/365

Kyubinhan:
* Add OpenLayers dependencies `ol`, `@types/ol`
* Add basic components for Map view
* Display weather stations in the map using Geojson
* Add functionality to Map component to render a tooltip when hovering Geojson features
* Create LayerSwitch component for Map view (but won't use it until we find more options on base layers)

Test: https://wps-pr-790.pathfinder.gov.bc.ca/morecast


web/.eslintrc.json
```diff
@@ -23,9 +23,10 @@
     ],
     "react/prop-types": "off",
     "no-useless-catch": "off",
+    "@typescript-eslint/ban-ts-comment": "warn",
     "@typescript-eslint/camelcase": "off",
     "@typescript-eslint/explicit-function-return-type": "off",
-    "@typescript-eslint/ban-ts-comment": "warn"
+    "@typescript-eslint/no-non-null-assertion": "off"
```
conbrad, line 29: We lose a lot of type safety if we turn this off globally. Is there another way or can we scope this down to the specific cases where we need this?

Kyubinhan: It was my laziness that went this route. I will revert this

Kyubinhan: Done

web/src/features/fireWeather/components/maps/WeatherMap.tsx
```diff
@@ -0,0 +1,80 @@
...
+        color: 'black'
+      })
+    })
+  })
+}
+
+const getArcGISAttribution = (service: string): string =>
+  `Tiles © <a href="https://services.arcgisonline.com/ArcGIS/rest/services/${service}/MapServer" target="_blank">ArcGIS</a>`
+const getArcGISMapUrl = (service: string): string =>
+  `https://server.arcgisonline.com/ArcGIS/rest/services/${service}/MapServer/tile/{z}/{y}/{x}`
```
Sybrand, line 31: I don't think we're licensed to use this data - see:  https://developers.arcgis.com/documentation/mapping-apis-and-location-services/licensing/license-location-services/

Sybrand (changes requested):
Looks real pretty. I'm just worried about the licensing. Am I missing something?
Also - I think we can just pull that json directly from source.

web/src/features/fireWeather/components/maps/WeatherMap.tsx
```diff
@@ -0,0 +1,80 @@
+import React, { useState, useCallback } from 'react'
+import { fromLonLat, get } from 'ol/proj'
+import * as olSource from 'ol/source'
+import GeoJSON from 'ol/format/GeoJSON'
+import { Circle as CircleStyle, Fill, Stroke, Style } from 'ol/style'
+import { FeatureLike } from 'ol/Feature'
+
+import WeatherStationsGeoJson from 'features/fireWeather/components/maps/weather-stations-geojson.json'
```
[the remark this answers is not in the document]
Kyubinhan: Where did you get this link? does this give your GeoJson? also you get 404 not found if you go to the link.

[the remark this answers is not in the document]
Sybrand: Sorry - copy paste issues - that's for `curl https://openmaps.gov.bc.ca/geo/pub/wfs\?SERVICE\=WFS\&VERSION\=2.0.0\&REQUEST\=GetFeature\&outputFormat\=json\&typeName\=WHSE_LAND_AND_NATURAL_RESOURCE.PROT_WEATHER_STATIONS_SP\&srsName\=EPSG:4326` 

try without the escape : https://openmaps.gov.bc.ca/geo/pub/wfs?SERVICE=WFS&VERSION=2.0.0&REQUEST=GetFeature&outputFormat=json&typeName=WHSE_LAND_AND_NATURAL_RESOURCE.PROT_WEATHER_STATIONS_SP&srsName=EPSG:4326

I've been trying to learn more about what we can query live bc's resources.

Kyubinhan (commented):
I am going to use a BC base map layer until I figure out the licensing issue with ArcGIS

web/src/features/fireWeather/components/maps/WeatherMap.tsx
```diff
@@ -0,0 +1,92 @@
...
+        color: '#E59982'
+      }),
+      stroke: new Stroke({
+        color: 'black'
+      })
+    })
+  })
+}
+
+const BC_WEATHER_STATIONS_WEB_FEATURE_SERVICE =
```
Sybrand, line 26: We could consider making these configurable - that way we can easily switch out layers.

web/src/features/fireWeather/components/maps/WeatherMap.tsx
```diff
@@ -0,0 +1,92 @@
...
+  'https://openmaps.gov.bc.ca/geo/pub/wfs?SERVICE=WFS&VERSION=2.0.0&REQUEST=GetFeature&outputFormat=json&typeName=WHSE_LAND_AND_NATURAL_RESOURCE.PROT_WEATHER_STATIONS_SP&srsName=EPSG:3857'
+const BC_ROAD_BASE_MAP_SERVER_URL =
+  'https://maps.gov.bc.ca/arcgis/rest/services/province/roads_wm/MapServer'
+
+const center = [-123.3656, 51.4484] // BC
```
Sybrand, line 31: Default view and zoom level we could also make configurable.

Sybrand (approved):
@Kyubinhan can you make sure we have a ticket for getting the other layers? Satellite etc.?

Left some comments re. config - but I think we're good to go. Real pity to be stuck with only the roads layer - but figuring that out is definitely outside the scope of this ticket.

web/.storybook/preview.js
```diff
@@ -2,6 +2,7 @@ import React from 'react'
 import { ThemeProvider } from '@material-ui/core/styles'
 
 import { theme } from '../src/app/theme'
+import '../src/index.css'
```
andrea-williams, line 5: Double-checking because my memory of this issue is foggy... wherever possible we want absolute filepaths for import statements, but Storybook only accepts relative filepaths. Is that correct?

andrea-williams (commented):
Do we still need the link to the ArcGis map now that we have this?

web/.storybook/preview.js
Kyubinhan: We want to use absolute paths in files under `/src` (I believe it's possible because we set `baseurl` to `src` in `tsconfig.json`.

andrea-williams (approved):

web/package.json
```diff
@@ -60,8 +62,8 @@
     "lint": "eslint './src/**/*.{ts,tsx}'",
     "lint:fix": "eslint --fix './src/**/*.{ts,tsx}'",
     "format": "prettier --write \"**/*.+(js|jsx|json|yml|yaml|css|md)\"",
-    "storybook": "start-storybook -p 6006 -s public",
-    "build-storybook": "build-storybook -s public"
+    "storybook": "export $(cat .env | xargs) && start-storybook -p 6006 -s public",
```
conbrad, line 65: Which env variables does storybook need?

conbrad (approved):

web/package.json
Kyubinhan: Probably none. There was one but then I reverted the changes. I kinda want to leave it this way tho just to be precautious.

PR #795, opened 2021-03-12
Bumps cypress from 5.2.0 to 6.6.0

conbrad:
Based on: https://docs.cypress.io/guides/references/migration-guide.html#Migrating-cy-route-to-cy-intercept

### DIFF
--- a/web/package.json
+++ b/web/package.json
@@ -84,7 +84,7 @@
     "@storybook/node-logger": "^6.1.11",
     "@storybook/preset-create-react-app": "^3.1.5",
     "@types/jest": "^24.9.1",
-    "cypress": "^5.2.0",
+    "cypress": "6.6.0",
     "eslint-config-prettier": "^6.10.0",
     "eslint-config-react": "^1.1.7",
     "eslint-plugin-prettier": "^3.1.2",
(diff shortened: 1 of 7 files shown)


Kyubinhan (commented):
It seems that formatting hasn't been run. Can you install Prettier plugin in your IDE and run it against all the modified files?

conbrad:
> It seems that formatting hasn't been run. Can you install Prettier plugin in your IDE and run it against all the modified files?

Done

Kyubinhan (approved):

web/package.json
```diff
@@ -84,7 +84,7 @@
     "@storybook/node-logger": "^6.1.11",
     "@storybook/preset-create-react-app": "^3.1.5",
     "@types/jest": "^24.9.1",
-    "cypress": "^5.2.0",
+    "cypress": "6.6.0",
```
Sybrand, line 87: Is there a reason we want to lock the minor?
```suggestion
    "cypress": "^6.6.0",
```

Sybrand (approved):
Looks pretty straight forward. Just a question/suggestion re. locking the version.

web/package.json
conbrad: Good question. I copied the specific update command from somewhere on the cypress site but don't remember where. I'll update to not lock the minor.

PR #797, opened 2021-03-15
Changed temp colors

conbrad:
Changed the colors according to: https://www.figma.com/file/2VuF3FNyhuf5gKD7jqEbeR/Untitled?node-id=91%3A182

~~Only changed temp colors, not sure if other colors need to be changed.~~
Changed both temp and RH colors according to design

~~Only used top row colors that were labeled, not sure if others need to be used.~~
Top row colors are temp, row below is RH in the figma design.

all
plumes


ambroseyau:
We need to change all the colours in the graph because users find the graphs confusing when users turned on more than one toggle

web/src/features/fireWeather/components/graphs/TempRHGraph.tsx
```diff
@@ -15,25 +15,25 @@ import {
...
+const rdpsTempColor = '#FF64DD'
+const rdpsRHColor = '#285777'
 const rdpsTempPlumeColor = 'rgba(244, 143, 65, 0.4)'
 const rdpsRHPlumeColor = 'rgba(42, 137, 137, 0.4)'
-const gdpsTempColor = '#f56c9c'
-const gdpsRHColor = '#32e7e7'
+const gdpsTempColor = '#7556CA'
+const gdpsRHColor = '#F2994A'
 const gdpsTempPlumeColor = 'rgba(255, 150, 169, 0.4)'
 const gdpsRHPlumeColor = 'rgba(148, 255, 235, 0.4)'
```
andrea-williams, line 34: with Dev down in OCP4 it's kinda hard to tell what's going on, but I wanted to flag this just in case - I would have expected that the plume colours would also be updated? Is that what the squares labelled "20%" are for @ambroseyau ? Also, was there a new colour for observed Dewpoint? I'm just seeing a blank square for that in Figma

conbrad: @andrea-williams dewpoint color change incoming. Not sure about the plume colors...

ambroseyau: The 20% is the color for the area color underneath the line graphs. The reason why we had to use the plume color is because users would be able to tell these are the model colours. Also, In most cases, users will only open all models at once to see if they're aligned. Chances of users using all models at once is quite infrequent.

web/src/features/fireWeather/components/graphs/TempRHGraph.tsx
```diff
@@ -15,25 +15,25 @@ import {
...
```
andrea-williams, line 34: the plume colours should be updated too - opacity should be 0.2 instead of 0.4. Just to make sure, @conbrad when we talk about "plumes" do you know what we're talking about?

andrea-williams (changes requested):
Just need to update the plume colours & opacity, and then should be good to go. Eg. GPDS RH markers are orange, but the plume is light blue - these should be the same colour.

web/src/features/fireWeather/components/graphs/TempRHGraph.tsx
conbrad: Sorry yeah wasn't sure on the figma design if there were plume color changes or not. Just cleared it up with Ambrose. The plume is the area under the lines correct?

andrea-williams: the plume is the lightly shaded areas for each of the weather variables. They indicate the range that each weather variable was forecasted at some point in time - e.g., because a model is run multiple times to predict wx for a given day, we should the range of values that were predicted, up to the 90th percentile. Hope that makes sense - we can always have a quick chat on Teams if that's still confusing

andrea-williams (approved):
Haven't been able to check this in OCP4 Dev because of platform issues, but by bending the rules slightly and looking at it in OCP3 Dev, it looks fine to me :)

Sybrand (approved):
I like the new colours.

PR #804, opened 2021-03-19
Story/new layout morecast/770

Kyubinhan:
Changes
* Introduce a new layout for MoreCast
* Include a map view into the page

Test: https://wps-pr-804.apps.silver.devops.gov.bc.ca/morecast

Note
The duplication complaint is from having two WxStationDropdown components. I started with just one with conditions but then it got messy so I decided to go with this route.


web/src/features/fireWeather/components/WxDataDisplays.tsx
```diff
@@ -89,41 +87,36 @@ export const WxDataDisplays = React.memo(function _(props: WxDataDisplaysProps)
             !observations && !noonForecasts && !gdpsModels && !hrdpsModels && !rdpsModels
 
           return (
-            <Paper key={code} className={classes.paper} elevation={3}>
-              <Typography className={classes.station} variant="subtitle1" component="div">
-                Weather station: {`${station.name} (${station.code})`}
+            <div key={code} className={classes.display}>
```
Sybrand, line 90: What's your reason for the move away from paper? (I know I avoid it when I can get away with just normal html)

web/src/components/Container.tsx
```diff
@@ -6,3 +6,12 @@ export const Container: React.FunctionComponent<ContainerProps> = (
 ) => {
   return <C maxWidth="md" {...props} />
 }
+
+type OptionalContainerProps = ContainerProps & { noContainer?: boolean }
+export const OptionalContainer = (props: OptionalContainerProps) => {
```
Sybrand, line 11: Why not only have the one container? Why the two different classes? Almost seems like this class shouldn't if a if (props.noContainer) ?

web/src/features/map/Map.tsx
```diff
@@ -126,6 +128,15 @@ const Map = ({ children, zoom, center, renderTooltip }: Props) => {
     map.getView().setCenter(center)
   }, [center]) // eslint-disable-line react-hooks/exhaustive-deps
 
+  useEffect(() => {
+    if (!map) return
+
+    // Wait enough time for the side panel to completely collapse then resize the map
+    setTimeout(() => {
+      map.updateSize()
+    }, 700)
```
Sybrand, line 137: Isn't there a better way to do this? I don't like relying on an arbitrary number. What happens on a slow computer where the panel takes 701ms?

Sybrand (commented):
Some comments - nothing critical - though I am concerned about the timeout. Would prefer that update not to be triggered by an arbitrary amount of time.
```
// Wait enough time for the side panel to completely collapse then resize the map
    setTimeout(() => {
      map.updateSize()
    }, 700)
```

Sybrand (approved):
Oops - meant to approve:
Some comments - nothing critical - though I am concerned about the timeout. Would prefer that update not to be triggered by an arbitrary amount of time.
```
// Wait enough time for the side panel to completely collapse then resize the map
    setTimeout(() => {
      map.updateSize()
    }, 700)
```

web/src/features/fireWeather/components/WxDataForm.tsx
```diff
@@ -2,24 +2,41 @@ import React, { useEffect, useState } from 'react'
...
+    },
+    '& .MuiChip-root': {
+      background: '#f0f0f0'
+    }
+  },
   stationDropdown: {
-    marginBottom: 10
+    marginRight: 16
+  },
+  toi: {
```
andrea-williams, line 27: I'd prefer to see this renamed to `timeOfInterest` - it's one less acronym to have to remember

web/src/features/map/Map.tsx
andrea-williams: I'd also like to see a better way of doing this. I noticed that the animation of closing the panel was not very smooth, e.g. for a while there's a delay where half the screen is just blank, and then the map gets redrawn.

andrea-williams (approved):
One minor comment regarding my preference for the naming of a variable, but otherwise looks good!

web/src/features/map/Map.tsx
Kyubinhan: It's slow because of all the graph rendering - and yeah it's going to be slower with slow browsers. I am thinking of removing the animation. That way when I set 0.1s for wait time, it will always work.

PR #811, opened 2021-03-22
Deploy to TEST and re-integrate USE_WFWX variable

andrea-williams:


.github/workflows/cleanup.yml
```diff
@@ -31,8 +31,31 @@ jobs:
           oc login "${{ secrets.OPENSHIFT_CLUSTER }}" --token="${{ secrets.OC4_TOOL_TOKEN }}"
           PROJ_TOOLS=e1e498-tools PROJ_DEV=e1e498-dev bash openshift/scripts/oc_cleanup_tools.sh ${SUFFIX} apply
 
-      - name: Clean up database (OCP4)
+      - name: Clean up api - test (OCP4)
```
Sybrand, line 34: I think that oc_cleanup_test.sh has the potential to become a complete cut and paste of oc_cleanup_dev.sh ; I think we should have a single oc_cleanup.sh (maybe with a better name?), where you specify the target as either dev or test.

It's true that right now test is a subset of dev - but if we have a single thing, we can be sure it will clean up both correctly.

.github/workflows/cleanup.yml
```diff
@@ -31,8 +31,31 @@ jobs:
...
+        shell: bash
+        run: |
+          oc login "${{ secrets.OPENSHIFT_CLUSTER }}" --token="${{ secrets.OC4_TEST_TOKEN }}"
+          PROJ_TEST=e1e498-test PROJ_DEV=e1e498-dev bash openshift/scripts/oc_cleanup_test.sh ${SUFFIX} apply
+
+    name: Delete DB resources associated with PR
+    runs-on: ubuntu-20.04
```
Sybrand, line 41: We could add a needs: [Delete non-DB resources associated with pr] - this could be nice, as it's going to clean up everything that uses the db first, then the db.

api/app/health.py
```diff
@@ -27,6 +27,7 @@ def patroni_cluster_health_check():
     # NOTE: In Openshift parlance "replica" refers to how many of one pod we have, in Patroni, a "Replica"
     # refers to a read only copy of of the Leader.
     # Get the number of pods that are ready:
+    logger.info('Response json: %s', resp_json)
```
Sybrand, line 30: Why do we need this?

openshift/scripts/oc_cleanup_test.sh
```diff
@@ -0,0 +1,44 @@
+#!/bin/sh -l
```
Sybrand, line 1: See comment above re. only having one cleanup script, and specifying the target.

Sybrand (changes requested):

conbrad (approved):
Parroting Sybrand's comments but looks good otherwise (based on my limited environment knowledge)

api/app/health.py
andrea-williams: We don't. Deleted.

.github/workflows/cleanup.yml
andrea-williams: Done

Kyubinhan (approved):

Sybrand (approved):
Woohoo!

PR #837, opened 2021-03-26
Table/graph view toggle on side bar via buttons

conbrad:
Moves table data and graphs into separate tabs in the sidebar.

Screen Shot 2021-03-31 at 12 56 55 PM
Screen Shot 2021-03-31 at 12 57 04 PM


Sybrand (approved):
Looks good.

web/src/features/fireWeather/components/WxDataDisplays.tsx
```diff
@@ -96,42 +102,69 @@ export const WxDataDisplays = React.memo(function _(props: WxDataDisplaysProps)
...
+                  <ErrorBoundary>
+                    <NoonForecastTable
+                      testId={`noon-forecasts-table-${code}`}
+                      title="Weather forecast noon values: "
+                      rows={noonForecasts}
+                    />
+                  </ErrorBoundary>
+                </React.Fragment>
+              ) : (
+                <React.Fragment>
```
Kyubinhan, line 150: You don't have to wrap this with Fragment if there is only one child

Kyubinhan (commented):
Is it just only me thinking the tab display is weird? it gets weirder with multiple stations due to the blank gaps.

web/src/features/fireWeather/components/WxDataDisplays.tsx
conbrad: Oops, copy-pasta, thanks!

conbrad:
> Is it just only me thinking the tab display is weird? it gets weirder with multiple stations due to the blank gaps.

Can you define weird? Are the blank gaps you mention the ones between tables, graphs or both?

andrea-williams (approved):
looks good to me!

Kyubinhan (commented):
Shouldn't the buttons be on the top? Also I think the active button should have a different background to indicate what's actually displaying

conbrad:
> Shouldn't the buttons be on the top? Also I think the active button should have a different background to indicate what's actually displaying

Wireframe shows the buttons below the title of the first station: https://miro.com/app/board/o9J_ktow2OY=/

I can look into active button backgrounds.

Kyubinhan:
> Wireframe shows the buttons below the title of the first station: https://miro.com/app/board/o9J_ktow2OY=/
> 
> I can look into active button backgrounds.

To me, the position of these buttons implies that they belong to the first station. It'll be more apparent what I am trying to say if you display two stations.

Kyubinhan (commented):
How about we make these buttons skinner? https://material-ui.com/components/button-group/#sizes-and-colors
Just a suggestion here so no need to make a change for it.

Kyubinhan (approved):

conbrad:
> How about we make these buttons skinner? https://material-ui.com/components/button-group/#sizes-and-colors
> Just a suggestion here so no need to make a change for it.

Oh thank you haha I was trying to do it myself. Will update.

PR #843, opened 2021-03-30
Reduce number of backups

Sybrand:
We're running out of space, and need to reduce the number of backups. We cannot create new backups at this point in time due to insufficient space.

- Reduce monthly backups from 1 to 0.
- Reduce total backup count from 3 to 2.


conbrad (approved):
How often do backups occur when we specific 0 monthly backups?

Sybrand:
> How often do backups occur when we specific 0 monthly backups?

Backups occur as frequently as the cronjob runs (see openshift/templates/backup-postgres-cronjob.yaml - schedule is currently set to "0 3,9,14,20 * * *". Reason for frequency is that those are the least busy times (in terms of cronjobs processing model runs) - and least likely to fail for that reason.

The settings adjusted in this PR relate to retention - so setting it to 0 monthly means, don't keep a monthly snapshot (see https://github.com/BCDevOps/backup-container#deployment--configuration)

Kyubinhan (approved):

andrea-williams (approved):

PR #846, opened 2021-03-31
patroni replicas 3->2 and increase size

Sybrand:
We no longer have space for 3 replicas.
- reduce replicas to 2 (so we have one leader, one follower).
- increase pvc size to 37Gi leaves us more wiggle room.

### DIFF
--- a/openshift/Jenkinsfile
+++ b/openshift/Jenkinsfile
@@ -43,7 +43,7 @@ pipeline {
                     // The size of the storage is in fact 30Gi ; The volume claim for a
                     // stateful set cannot be altered. The stateful set needs to be re-created
                     // and data needs to be restored from backup.
-                    sh "CPU_REQUEST=100m CPU_LIMIT=2000m MEMORY_REQUEST=4Gi MEMORY_LIMIT=8Gi PVC_SIZE=25Gi PROJ_TARGET=${projProd} ./openshift/scripts/oc_provision_db.sh prod apply"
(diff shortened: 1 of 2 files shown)


conbrad (changes requested):
Apparently Patroni runs the Raft consensus protocol: https://patroni.readthedocs.io/en/latest/SETTINGS.html#raft

Distributed consensus needs at least 3 replicas, a majority, to prevent byzantine faults so I'm not sure this will work. Is our system configured to have all writes go through a single master or can both replicas accept writes?

Here's an issue on the Patroni repo that also says it's not possible with failover: https://github.com/zalando/patroni/issues/748

It may be easier to just run a single postgres instance for now?

OR, the last Q&A answer in the Raft section linked above says you can just run the `patroni_raft_controller` on the 3rd node without PostgresSQL for consensus. This may be the best option.

Sybrand:
> Apparently Patroni runs the Raft consensus protocol: https://patroni.readthedocs.io/en/latest/SETTINGS.html#raft
> 
> Distributed consensus needs at least 3 replicas, a majority, to prevent byzantine faults so I'm not sure this will work. Is our system configured to have all writes go through a single master or can both replicas accept writes?
> 
> Here's an issue on the Patroni repo that also says it's not possible with failover: [zalando/patroni#748](https://github.com/zalando/patroni/issues/748)
> 
> It may be easier to just run a single postgres instance for now?
> 
> _Or_, the last Q&A answer in the Raft section linked above says you can just run the `patroni_raft_controller` on the 3rd node without PostgresSQL for consensus. This may be the best option.

I don't think this applies. I don't know how much of the decision to 

We have a setup where only the leader accepts writes, the replicas are all read-only.

I did a quick test in a two node environment. Of you destroy the leader, the follower takes over - so it seems to work fine.

conbrad (approved):
Discussed offline

Kyubinhan (approved):

PR #854, opened 2021-04-07
Highlight min/max values in data tables

andrea-williams:
https://wps-pr-854.apps.silver.devops.gov.bc.ca/morecast


conbrad (approved):
LGTM - just remove package-lock.json as mentioned offline

web/src/features/fireWeather/components/tables/SortableTableByDatetime.tsx
```diff
@@ -78,6 +136,68 @@ function SortableTableByDatetime<R extends WeatherValue>(props: Props<R>) {
...
+
+  rowsSortedByDatetime.forEach((row, idx) => {
+    if (
+      row.relative_humidity?.toFixed(RH_VALUES_DECIMAL) ===
+      minMaxValuesToHighlight.relative_humidity?.toFixed(RH_VALUES_DECIMAL)
+    ) {
+      rowIds['relative_humidity'].push(idx)
+    }
+    if (
+      row.precipitation?.toFixed(PRECIP_VALUES_DECIMAL) ===
```
Sybrand, line 172: I think we need to exclude 0. If the max precip is 0, then we just have all the rows highlighted. I know it's mathematically correct, but it makes for a weird display i.m.h.o.

Sybrand (approved):
Can we have ambrose, and then one of the weather people give a look before we merge?
Also - I think the multiple highlights on precip. is weird when the max is 0.

conbrad (approved):

conbrad (approved):

Sybrand (approved):
Looks real pretty!

web/src/features/fireWeather/components/tables/SortableTableByDatetime.tsx
```diff
@@ -65,6 +105,24 @@ interface Props<R> {
...
+
+interface RowIdsOfMinMaxValues {
+  relative_humidity: number[]
+  precipitation: number[]
+  wind: number[]
+  max_temp: number[]
+  min_temp: number[]
+}
+
 function SortableTableByDatetime<R extends WeatherValue>(props: Props<R>) {
```
conbrad, line 126: A unit test suite for this component may be in order due to all the conditional branches and switches -- I'm not sure the cypress tests capture that and I think unit tests would be easier to write. WDYT?

conbrad (changes requested):

web/src/utils/table.test.ts
```diff
@@ -30,4 +89,26 @@ describe('Table util functions', () => {
       expect(arr).toEqual(deepCopy)
     })
   })
+
+  describe('getMinMaxValueCalculator', () => {
```
conbrad, line 93: I think we should also add cases for when values are null and max value is 0

web/src/features/fireWeather/components/tables/SortableTableByDatetime.tsx
```diff
@@ -136,8 +186,56 @@ function SortableTableByDatetime<R extends WeatherValue>(props: Props<R>) {
...
+                            if (rowIds['min_temp'].includes(idx)) {
+                              className = classes.minTemperature
+                              testId = `min-temperature-td`
+                            } else if (rowIds['max_temp'].includes(idx)) {
+                              className = classes.maxTemperature
+                              testId = `max-temperature-td`
+                            }
+                            break
+                          }
+                          case 'precipitation':
```
conbrad, line 207: Not sure if this is easier to write as a cypress test or unit test but I think we should add a test that checks that the `className` and `testId` are set for either `precipitation`, `delta_precipitation` or `total_precipitation`.

The easiest way would be to pull out a utility function that accepts the columns and does the switch logic so it's easier to test with jest, then just have the component import and use it.

More component-y way would be to use some like enzyme to shallow render the component and check it with the library assertions: https://devhints.io/enzyme

conbrad (changes requested):
Looks good, just a few more tests I think

conbrad (approved):
🚀

PR #860, opened 2021-04-08
Return only stations we have data for

conbrad:
Changes `WeatherMap` to use the stations endpoint so that it shows the same stations listed in the dropdown.

Formats `/stations/` endpoint to return GeoJSON formatted stations based on the below RFC:
https://tools.ietf.org/html/rfc7946

Note: GeoJSON orders coordinates as `[long, lat]` (https://tools.ietf.org/html/rfc7946#section-3.1.1)


web/src/features/fireWeather/components/maps/WeatherMap.tsx
```diff
@@ -37,25 +38,20 @@ interface Props {
...
+  const { stations } = useSelector(selectFireWeatherStations)
 
   useEffect(() => {
-    fetch(WEATHER_STATIONS_WEB_FEATURE_SERVICE)
-      .then(resp => resp.json())
-      .then(json => {
-        setWxStationsGeoJSON(json)
-      })
-  }, [])
+    dispatch(fetchWxStations())
```
Sybrand, line 46: The station list is being requested (and thus downloaded and set twice). I'm not sure how worried I am about this.... since we're looking at having the stations on the map show all kinds of additional information we're probably going to end up having two api endpoints - one for a simple station list, and one for a list with extended information.

web/src/utils/dropdown.ts
```diff
@@ -0,0 +1,23 @@
+import { Station } from '/Users/conor/projects/wps/web/src/api/stationAPI'
+
+export interface Option {
+  name: string
+  code: number
+}
+
+export const getAutoCompleteOption = (
```
Sybrand, line 8: Can we get a comment explaining what this does? I imagine looking at this code without context would be confusing to me.

web/src/api/stationAPI.ts
```diff
@@ -1,7 +1,7 @@
 import axios from 'api/axios'
 import { FireSeason } from 'api/percentileAPI'
 
-export interface Station {
+export interface WeatherStation {
```
Sybrand, line 4: Having Station and WeatherStation is a bit confusing. The one is for the GeoJSON station list, the other is for the Station object from the other endpoints. They're both weather stations.... We either need better names for them, or be consistent in how the API returns them I think.

Sybrand (changes requested):
My only real concern is having "Station" and "WeatherStation" in web/src/api/stationAPI.ts - I feel like they need better names, or that we should just have one.

web/src/utils/dropdown.ts
conbrad: Done -- also renamed a bunch of stuff to make this more clear.

web/src/features/fireWeather/components/maps/WeatherMap.tsx
conbrad: Should I write a ticket for this?

Sybrand: I think rather just add a note to https://app.zenhub.com/workspaces/wildfire-predictive-services-5e321393e038fba5bbe203b8/issues/bcgov/wps/849 ; we'll probably address it as part of that ticket.

Sybrand (approved):

PR #866, opened 2021-04-09
Cleanup - loose ends

Sybrand:
- Change c-haines label to match app (cleanup was missing it)
- Added ignore-not-found=true - was causing intermittent failure when pods not present.


conbrad (approved):

Sybrand:
- Change c-haines label to match app (cleanup was missing it)
- Added ignore-not-found=true - was causing intermittent failure when pods not present.

PR #871, opened 2021-04-12
Adds audit logging to authenticated routes for Prepski

conbrad:
- Adds migration to add logging table that persists username, api path and timestamp
- Authentication request decorator extracts username from token, path from request
- Adds Audit model and creation crud function called by authentication request decorator
- Catch base JWT decode exception instead of most general exception: https://pyjwt.readthedocs.io/en/latest/api.html#jwt.exceptions.InvalidTokenError


api/alembic/versions/c92ba5b2b3d5_adds_audit_model.py
```diff
@@ -0,0 +1,41 @@
...
+# revision identifiers, used by Alembic.
+revision = 'c92ba5b2b3d5'
+down_revision = '004ac48ffe18'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.create_table('audits',
```
andrea-williams, line 21: I'm inclined to say we should name this table something like `audit_user_login` because as we build out Preppy, we'll also need audit logs for modifications made to preparedness levels and the like, which I'd imagine will be in a separate table from this one. But I'd be interested to hear what @Sybrand has to say about this

Sybrand: Auditing won't be normalized. So the create_date, create_user, update_date, create_user will be oin the revelevant table. E.g. if we have a "prep_level" table, and a user changes it, the audit columns will be in the prep_level table.
You make a good point re naming of the table though, though it's not audit_user_login, since the login happens on keycloak. A better name would tell us that these are audits relating to endpoints accessed by users. Don't know a short way to say that. "access_audits" ? "api_audits" ?

api/app/auth.py
```diff
@@ -1,25 +1,29 @@
...
     # RSA public key format
     keycloak_public_key = '-----BEGIN PUBLIC KEY-----\n' + \
         config.get('KEYCLOAK_PUBLIC_KEY') + '\n-----END PUBLIC KEY-----'
 
     try:
-        jwt.decode(token, keycloak_public_key, algorithm='RS256')
+        decoded_token = jwt.decode(token, keycloak_public_key, algorithm='RS256')
+        username = decoded_token['preferred_username']
+        path = request.url.path
+        create_audit_log(username, path)
```
Sybrand, line 26: I'm busy digging to see if there's a way to separate the audit from the authentication nicely....

Sybrand (changes requested):

api/alembic/versions/c92ba5b2b3d5_adds_audit_model.py
```diff
@@ -0,0 +1,41 @@
...
+                    sa.Column('id', sa.Integer(), nullable=False),
+                    sa.Column('username', sa.String(), nullable=False),
```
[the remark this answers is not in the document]
Sybrand: Of course.... it's not a create user.... hmmm.... maybe just "user" ? I don't know.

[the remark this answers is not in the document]
conbrad: I think you are correct in your first comment and my change was made with your second comment's understanding.

But a closer look at the document describes the `CREATE_USER` field as "The user or proxy account that created the record. **"
I should probably also change the username I'm persisting to this field as just "username" instead of the current "username@idir", but I'm not sure what format BCEID's are in.

api/alembic/versions/c92ba5b2b3d5_adds_audit_model.py
conbrad: "api_access_audits"?

api/alembic/versions/c92ba5b2b3d5_adds_audit_model.py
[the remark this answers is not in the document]
Sybrand: Oh no - don't touch the username - take it exactly as given. One doesn't even need idir for things to work. For example - I can create a username "sybrand" in Keylock, and configure things so that one could log in with keycloak, or with idir or with bceid. So just take the value you get and persist it blindly.

api/app/auth.py
Sybrand: @conbrad maybe I'm thinking ahead too far - but I suggest we create another function entirely:
`async def audit(request: Request, username: str = Depends(authenticate)):
    """ Audit depends on authenticate """
    logger.info('audit %s for %s', request.url.path, username)
    path = request.url.path
    create_audit_log(username, path)`

and then on a route we can specify like this:
`router = APIRouter(
    prefix="/forecasts",
    dependencies=[Depends(authenticate), Depends(audit)],
)`
That way auth and audit are decoupled a bit better.
Not sure about my proposed function names - one could have a "secure_audit" and an "insecure" audit.

api/app/auth.py
```diff
@@ -13,19 +15,38 @@
 
 
 async def authenticate(token: str = Depends(oauth2_scheme)):
-    """ Returns True when validation of the token is successful """
+    """ Returns True when validation of the token is successful. """
```
Sybrand, line 18:
```suggestion
    """ Returns Decoded token when validation of the token is successful. Returns empty dictionary on failure to decode """
```

api/app/auth.py
```diff
@@ -13,19 +15,38 @@
...
         config.get('KEYCLOAK_PUBLIC_KEY') + '\n-----END PUBLIC KEY-----'
 
     try:
-        jwt.decode(token, keycloak_public_key, algorithm='RS256')
-        return True
-    except Exception as exception:
-        detail = 'Could not validate the credential ({})'.format(exception)
+        decoded_token = jwt.decode(token, keycloak_public_key, algorithm='RS256')
+        return decoded_token
+    except InvalidTokenError:
```
Sybrand, line 26:
```suggestion
    except InvalidTokenError as exception:
```

api/app/auth.py
```diff
@@ -13,19 +15,38 @@
...
+        detail = 'Could not validate the credential (Not enough segments)'
```
Sybrand, line 27: Rather log the exception - we don't know that InvalidTokenError always means not enough segments, it could be something else.
```suggestion
        logger.error('Could not validate the credential', exception)
```

api/app/auth.py
```diff
@@ -13,19 +15,38 @@
...
+    """ Audits attempted requests based on bearer token. """
+    path = request.url.path
+
+    if not token:
+        create_api_access_audit_log(str(None), False, path)
+        return False
+
+    username = token['preferred_username']
+    create_api_access_audit_log(username, True, path)
+    return True
```
Sybrand, line 42: Do you know if we have to return true? What happens if you just don't have a return value?

api/app/auth.py
```diff
@@ -13,19 +15,38 @@
...
         logger.error(detail)
+        return {}
+
+
+async def audit(request: Request, token=Depends(authenticate)):
+    """ Audits attempted requests based on bearer token. """
+    path = request.url.path
+
+    if not token:
+        create_api_access_audit_log(str(None), False, path)
```
Sybrand, line 37: str(None) creates a string that says 'None'. We don't want that - if the username is None, we want it to be None, not a string.
```suggestion
        create_api_access_audit_log(None, False, path)
```

api/app/auth.py
```diff
@@ -13,19 +15,38 @@
...
+    except InvalidTokenError:
+        detail = 'Could not validate the credential (Not enough segments)'
         logger.error(detail)
+        return {}
+
+
+async def audit(request: Request, token=Depends(authenticate)):
+    """ Audits attempted requests based on bearer token. """
+    path = request.url.path
+
```
Sybrand, line 35: Could do the same in less lines of code:
```
username = token.get('preferred_username', None)
create_api_access_audit_log(username, True, request.url.path)
```

api/app/c_haines/severity_index.py
```diff
@@ -7,8 +7,11 @@
 import tempfile
 import logging
 import json
+try:
+    import ogr
```
Sybrand, line 11: Do we need this? Isn't everything on a version of osgeo where we can just do a straight up `from osgeo import ogr`?

api/app/db/crud/api_access_audits.py
```diff
@@ -0,0 +1,28 @@
...
+logger = logging.getLogger(__name__)
+
+
+def create_api_access_audit_log(
+        username: str,
+        success: bool,
+        path: str) -> APIAccessAudit:
+    """ Create an audit log. """
+    with app.db.database.get_write_session_scope() as session:
+        now = datetime.now(tz=timezone.utc)
```
Sybrand, line 19: just an f.y.i - there's a app.time_utils.get_utc_now that does exactly this, for ensuring that unit tests give consistent results it's mocked out by default in api/app/routers/forecasts.py ;
Doesn't matter in this context, and there isn't a unit test looking at the date - but just so you know.

api/app/db/models/api_access_audits.py
```diff
@@ -0,0 +1,22 @@
...
+    """ Class representing table structure of 'audits' table in DB.
+        Each record is immutably created for each authenticated API request
+        based on IDIR username, request path and timestamp.
+    """
+    __tablename__ = 'api_access_audits'
+    __table_args__ = (
+        {'comment': 'The audit log of an authenticated request by a user.'}
+    )
+    id = Column(Integer, primary_key=True)
+    create_user = Column(String, nullable=False, index=True)
```
Sybrand, line 19: Pls. make create_user nullable for when we can't decode the token.

api/app/main.py
```diff
@@ -115,7 +115,7 @@ async def get_health():
 
 
 @api.post('/observations/', response_model=schemas.observations.WeatherStationHourlyReadingsResponse)
-async def get_hourlies(request: schemas.shared.WeatherDataRequest, _: bool = Depends(authenticate)):
+async def get_hourlies(request: schemas.shared.WeatherDataRequest, _=Depends(authentication_required)):
```
Sybrand, line 118: Please add audit.
```suggestion
async def get_hourlies(request: schemas.shared.WeatherDataRequest, _=Depends(authentication_required), __=Depends(audit)):
```

api/app/tests/c_haines/test_bounding_box.py
```diff
@@ -1,7 +1,14 @@
 """ Test the bounding box logic.
 """
 import os
-import gdal
+try:
```
Sybrand, line 4: See other question re. from osgeo import gdal being enought? Where is this failing?

Sybrand (changes requested):

api/app/auth.py
conbrad: It works without a return value -- I removed it.

api/app/c_haines/severity_index.py
conbrad: I just do this because I'm assuming `import ogr` works on other machines without `from osgeo import ogr` working but I'll change it

api/app/dependency.py
```diff
@@ -0,0 +1,50 @@
+""" Authentication module for validating access tokens generated by Keycloak
```
andrea-williams, line 1: is `dependency.py` where we envision writing all our dependent modules in the future? I ask because if this isn't the case, the filename seems pretty vague

api/app/db/crud/api_access_audits.py
```diff
@@ -0,0 +1,28 @@
...
+    """ Create an audit log. """
+    with app.db.database.get_write_session_scope() as session:
+        now = get_utc_now()
+        audit_log = APIAccessAudit(create_user=username, path=path, success=success,
+                                   create_timestamp=now)
+        try:
+            session.add(audit_log)
+            session.commit()
+        except IntegrityError:
+            logger.info('Skipping duplicate record for username=%s @ %s for path=%s',
```
andrea-williams, line 26: I'm curious to know if this error handling is in here because it's an error that's already been encountered?

api/app/dependency.py
```diff
@@ -0,0 +1,50 @@
...
+from app import config
+from app.db.crud.api_access_audits import create_api_access_audit_log
+
+logger = logging.getLogger(__name__)
+
+# Parse request header and pass the token
+oauth2_scheme = OAuth2PasswordBearer(tokenUrl="/token")
+
+
+async def authenticate(token: str = Depends(oauth2_scheme)):
```
Sybrand, line 17: It's not logging un-authenticated attempts, since OAuth2PasswordBearer raises a 401 exception. You'll have to wrap the call that's happening in oauth2_scheme. Might be a good idea to add a test case for this scenario - it's really easy to miss.

conbrad: Oof thanks

api/app/tests/c_haines/test_bounding_box.py
```diff
@@ -1,7 +1,8 @@
 """ Test the bounding box logic.
 """
 import os
-import gdal
+
+from osgeo import gdal, ogr
```
Sybrand, line 5: ogr is unused.
```suggestion
from osgeo import gdal
```

Sybrand (changes requested):
Looking good - see comments - just missing the no auth header use case.

You can test using curl, and then look in the DB to validate
```
curl 'http://localhost:8080/api/weather_models/RDPS/predictions/summaries/' \
  --data-raw '{"stations":[322],"time_of_interest":"2021-04-13T12:58:00-08:00"}'
```
It should still give a 401 error, but should log the failed attempt in the db.

api/app/db/crud/api_access_audits.py
Sybrand: Good point, maybe better off just letting this exception happen. There shouldn't be integrity issues - there's no unique key other than the id, so this shouldn't happen?

api/app/dependency.py
conbrad: Yes, open to better names. I will update the top comment missed.

api/app/db/crud/api_access_audits.py
conbrad: I added this because there was a pattern of it's usage in other database creation calls but not sure this will ever happen because the only uniqueness constraint is the auto-incrementing primary key.

api/app/dependency.py
Sybrand: @andrea-williams  Suggesions? I've seen a convention for fastapi just naming in depends.py - but that doesn't seem any clearer to me.

api/app/db/crud/api_access_audits.py
Sybrand: I think drop it - let it raise the exception, let's see if it ever happens.

api/app/dependency.py
andrea-williams: depends.py just reminds me of adult diapers. I'm not sure if a filename change is required (unless people like `dependencies.py`?), probably just updating the file's documentation so that it isn't auth-specific should clear up confusion

api/app/db/crud/api_access_audits.py
andrea-williams: if the other database creation calls you're talking about are related to the Env Canada models (and maybe also in our DB writes for observations and forecasts...?), then yes IntegrityError handling is required there because it's a frequent situation that we pull the same data from external sources multiple times - there's no way to know in advance whether the data is a duplicate or if it's been revised. But I don't think IntegrityError should happen here, unless for some reason there's multiple auth calls being made for the same user at the exact same timestamp??

conbrad: @andrea-williams Yes. @Sybrand Will do.

api/app/dependency.py
conbrad: I don't have a strong opinion either way. I'll wait for a decision to update this.

api/app/dependency.py
conbrad: Apparently it doesn't actually raise a 401 exception: https://fastapi.tiangolo.com/tutorial/security/first-steps/#what-it-does

I think writing a test will flush out what needs to be changed here.

api/app/dependency.py
Sybrand: I say update the module comments and leave filename as is.

api/app/dependency.py
Sybrand: See source code: https://github.com/tiangolo/fastapi/blob/master/fastapi/security/oauth2.py#L157

FastAPI generally follows the pattern of raising exception for HTTP error codes - that then get transformed into the appropriate response.

andrea-williams (approved):
Unlike Sybrand, I am happy :)

api/app/dependency.py
conbrad: Huh, good example of code not lining up with documentation...

api/app/db/crud/api_access_audits.py
```diff
@@ -0,0 +1,23 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+def create_api_access_audit_log(
+        username: str,
+        success: bool,
+        path: str) -> APIAccessAudit:
+    """ Create an audit log. """
+    print("in crud method")
```
Sybrand, line 17: Please remove print

api/app/db/crud/api_access_audits.py
```diff
@@ -0,0 +1,23 @@
...
+
+import app.db.database
+
+logger = logging.getLogger(__name__)
+
+
+def create_api_access_audit_log(
+        username: str,
+        success: bool,
+        path: str) -> APIAccessAudit:
```
Sybrand, line 15: Method does not return anything, void equivalent should be None
```suggestion
        path: str) -> None:
```

api/app/tests/test_auth.feature
```diff
@@ -1,21 +1,25 @@
 Feature: Authentication
 
     Scenario: Handling unauthenticated users
+        Given Unauthenticated access audit logging is spied
```
Sybrand, line 4: We re-use this for authenticated and unauthenticated
```suggestion
        Given access audit logging is spied
```

api/app/tests/test_auth.feature
```diff
@@ -1,21 +1,25 @@
...
-            | just_token   | 401    | Not authenticated                                       | /api/weather_models/GDPS/predictions/summaries/   |
+            | Basic token  | 401    | Could not validate the credential (Not enough segments) | /api/weather_models/GDPS/predictions/summaries/   |
+            | just_token   | 401    | Could not validate the credential (Not enough segments) | /api/weather_models/GDPS/predictions/summaries/   |
             | Bearer token | 401    | Could not validate the credential (Not enough segments) | /api/weather_models/GDPS/predictions/summaries/   |
-            | just_token   | 401    | Not authenticated                                       | /api/weather_models/GDPS/predictions/most_recent/ |
+            | just_token   | 401    | Could not validate the credential (Not enough segments) | /api/weather_models/GDPS/predictions/most_recent/ |
             | Bearer token | 401    | Could not validate the credential (Not enough segments) | /api/weather_models/GDPS/predictions/most_recent/ |
 
     Scenario: Verifying authenticated users
+        Given Unauthenticated access audit logging is spied
```
Sybrand, line 19: We re-use this for authenticated and unauthenticated
```suggestion
        Given access audit logging is spied
```

api/app/tests/test_auth.py
```diff
@@ -11,11 +13,24 @@ def test_auth_1st_scenario():
     """ BDD Scenario #1. """
 
 
+@given("Unauthenticated access audit logging is spied", target_fixture='create_audit_logs')
```
Sybrand, line 16: We can re-use this for both scenarios
```suggestion
@given("access audit logging is spied", target_fixture='create_audit_logs')
```

api/app/tests/test_auth.py
```diff
@@ -36,6 +51,18 @@ def test_auth_2nd_scenario():
     """ BDD Scenario #2. """
 
 
+@given("Authenticated access audit logging is spied", target_fixture='create_audit_logs')
```
Sybrand, line 54: We can remove this one - we only need one given

api/app/auth.py
```diff
@@ -1,31 +1,56 @@
...
+async def authentication_required(token=Depends(authenticate)):
+    """ Raises HTTPExcecption with status code 401 if authentation fails."""
+    if not token:
         raise HTTPException(
             status_code=status.HTTP_401_UNAUTHORIZED,
-            detail=detail,
-            headers={'WWW-Authenticate': 'Bearer'},
-        ) from exception
+            detail="Could not validate the credential (Not enough segments)",
```
Sybrand, line 54: We don't actually know what the detail is at this point do we? We know there's no token - but I wonder if we should even pass that down? Maybe we can just skip detail?

Sybrand (changes requested):
Some minor things - looking good.

Sybrand (approved):

ccrocker1 (approved):

PR #877, opened 2021-04-14
Story/forecast accuracy/849

Sybrand:
# Front-end
Dev (json+db): https://wps-pr-877.apps.silver.devops.gov.bc.ca/morecast
Test (wfwx): https://wps-pr-877-test.apps.silver.devops.gov.bc.ca/morecast
Relative Humidity accuracy colors the map stations:
Screen Shot 2021-04-19 at 3 31 57 PM

# API
https://wps-pr-877.apps.silver.devops.gov.bc.ca/api/docs

- moved stations to own route
- added default values for geojson
- added stations/details/ endpoint
- implemented stations/details/ to work with local db
- implemented stations/details/ to work against wfwx api


api/app/db/crud/stations.py
```diff
@@ -0,0 +1,64 @@
...
+                              month=date_of_interest.month,
+                              day=date_of_interest.day,
+                              hour=20, tzinfo=timezone.utc)
+    if date_of_interest < noon_for_today:
+        # oh - you want noon for the prevous day
+        yesterday = date_of_interest - timedelta(days=1)
+        return datetime(
+            year=yesterday.year, month=yesterday.month, day=yesterday.day, hour=20, tzinfo=timezone.utc)
+    # you want noon for today
+    return noon_for_today
```
andrea-williams: I think the code would be more readable if we changed the variable names `noon_for_today` -> `noon_for_toi` and `yesterday` -> `previous_day`. When I first read this I was confused because I was thinking that "yesterday" literally meant yesterday, and was imagining lots of problems when users are looking up dates of interest from previous years

api/app/stations.py
```diff
@@ -77,22 +84,61 @@ async def get_stations(
     return _get_stations_local()
 
 
+def set_weather_variables(parent, station_union):
+    """
+    TODO: add a comment! and explain this code - it's difficult to read!
```
andrea-williams: ...?

web/src/api/stationAPI.ts
```diff
@@ -15,6 +15,23 @@ export interface GeoJsonStation {
...
   geometry: StationGeometry
 }
+
+export interface DetailedGeoJsonStation {
+  type: string
+  properties: DetailedStationProperties
+  geometry: StationGeometry
+}
+
+export interface DetailedStationProperties extends StationProperties {
```
andrea-williams, line 25: I'm confused - how are weather values a property of the station? I'd like to see a comment explaining this.

web/src/features/fireWeather/components/maps/WeatherMap.tsx
```diff
@@ -4,25 +4,24 @@ import * as olSource from 'ol/source'
...
+import { fetchWxStations } from 'features/stations/slices/stationsSlice'
 
 import Map from 'features/map/Map'
 import TileLayer from 'features/map/TileLayer'
 import VectorLayer from 'features/map/VectorLayer'
 import { useDispatch, useSelector } from 'react-redux'
 
 import { selectFireWeatherStations } from 'app/rootReducer'
+import { getDetailedStations } from 'api/stationAPI'
+import { computeAccuracyColors } from './stationAccuracy'
```
andrea-williams, line 16: naughty naughty... no relative imports!

web/src/features/fireWeather/components/maps/stationAccuracy.test.ts
```diff
@@ -0,0 +1,143 @@
...
+  StationMetrics,
+  neutralColor,
+  noDataColor,
+  tempColorScale,
+  rhColorScale,
+  computePercentageDifference,
+  computeScaleIndex,
+  neutralIndex,
+  differenceToMagnitude
+} from './stationAccuracy'
```
andrea-williams, line 12: please change relative import to absolute

web/src/features/fireWeather/components/maps/stationAccuracy.test.ts
```diff
@@ -0,0 +1,143 @@
...
+
+describe('Station map color accuracy', () => {
+  describe('computeAccuracyColors', () => {
+    it('should return the neutral color code when observed an forecasted are equal', () => {
```
andrea-williams, line 16:
```suggestion
    it('should return the neutral color code when observed and forecasted are equal', () => {
```

web/src/features/fireWeather/components/maps/stationAccuracy.test.ts
```diff
@@ -0,0 +1,143 @@
...
+        observations: {
+          temperature: 0,
+          relative_humidity: 0
+        },
+        forecasts: {
+          temperature: 300,
+          relative_humidity: 300
+        }
+      }
+      expect(computeAccuracyColors(largeDifference)).toEqual({
```
andrea-williams, line 55: this rears its ugly head again... according to the ticket, if the forecast > observed, then the colour should be warm. But I can't remember if that's the logic that the forecasters want (so I'm not sure whether this implementation is correct or if the ticket description is correct).

conbrad: From the ticket: "warm (orange) colours mean the RH was lower than forecasted, cool colours (green) mean the RH was higher than forecasted"

The screenshot above shows forecasted RH at 20, observed at 26, for Churn Creek, so RH was higher than forecasted and thus green.

web/src/api/stationAPI.ts
conbrad: These are weather values measured at the station, I can add that if it makes it clearer

web/src/features/fireWeather/components/maps/stationAccuracy.test.ts
conbrad: Just realized the logic implements this but the test describes it wrong, thanks!

api/app/db/crud/stations.py
Sybrand: @andrea-williams have a look, I re-factored this method before I saw your comment.

api/app/stations.py
Sybrand: :) again - updated it before your comment. please have a look now.

conbrad (approved):

api/app/tests/test_auth.py
```diff
@@ -7,16 +8,21 @@
...
+          example_converters=dict(token=str, status=int, message=str, endpoint=str, verb=str))
 def test_auth_1st_scenario():
     """ BDD Scenario #1. """
 
 
-@given("I am an unauthenticated user <token> when I access a protected <endpoint>", target_fixture='response')
-def given_unauthenticated_user(token: str, endpoint: str):
+@given("I am an unauthenticated user <token> when I <verb> a protected <endpoint>", target_fixture='response')
+def given_unauthenticated_user(token: str, endpoint: str, verb: str):
     """ Make POST {endpoint} request which is protected """
```
andrea-williams, line 18: this function comment is misleading because it's now out of date

api/app/wildfire_one.py
```diff
@@ -129,14 +123,41 @@ async def _fetch_raw_stations(session: ClientSession, headers: dict, query_build
...
+        query_builder: BuildQuery) -> (Dict[int, GeoJsonDetailedWeatherStation], Dict[str, int]):
+    stations = {}
+    id_to_code_map = {}
+    # Put the stations in a nice dictionary.
+    async for raw_station in _fetch_raw_stations_generator(session, headers, query_builder):
+        station_code = raw_station.get('stationCode')
+        station_status = raw_station.get('stationStatus', {}).get('id')
+        # Because we can't filter on status in the RSQL, we have to manually exclude stations that are
+        # not active.
+        if _is_station_valid(raw_station):
```
andrea-williams, line 138: I'm not understanding how this `if` statement is filtering out the inactive stations?

api/app/db/crud/stations.py
```diff
@@ -0,0 +1,65 @@
...
+    """ Return union of forecasts and observations. One could argue this method doesn't belong
+    in the stations crud - but it's only used to create the detailed stations response. """
+    noon_date = _get_noon_date(date_of_interest)
+    # It must be possible to do this using sqlalchemy - but things got a bit complicated, and I opted
+    # for a good old fashioned sql query.
+    query = """
+---
+--- most recent forecasts
+---
+select 'forecasts' as record_type, noon_forecasts.weather_date, noon_forecasts.station_code,
```
andrea-williams, line 36: seeing raw SQL queries in code makes me nervous, but I'll let it slide for now. But I would much prefer if at some point we return to this and find a more secure way to make the query

web/src/features/fireWeather/components/maps/stationAccuracy.ts
```diff
@@ -0,0 +1,127 @@
...
+  '#D6B2ED',
+  neutralColor,
+  '#7AD3CE',
+  '#17B8A7',
+  '#089E83'
+]
+
+/**
+ *  Calculates the percentage difference between observed and forecasted station metrics.
+ *  Uses increments of 3% to determine color code in the associated color arrays.
```
andrea-williams, line 51: I thought the increments of 3% only applied to RH. It was my understanding that temperature should have increments of 2 degrees?

andrea-williams: oh duh, this ticket is only for RH.
Anyway, the code that is in here for temperature is wrong - shouldn't be using %age difference, should be calculating absolute degrees differentials

andrea-williams (approved):
Some issues, but nothing that can't be left for later refactoring

web/src/features/fireWeather/components/maps/stationAccuracy.ts
conbrad: Yep will update that when adding temperature. I'll note this on the next ticket.

api/app/db/crud/stations.py
Sybrand: I fully agree. Less so about security, it should be secure the way it's done here - but it defeats the purpose of having an ORM.

web/src/features/fireWeather/components/maps/stationAccuracy.ts
conbrad: https://app.zenhub.com/workspaces/wildfire-predictive-services-5e321393e038fba5bbe203b8/issues/bcgov/wps/879

PR #905, opened 2021-04-21
updated python using poetry --update

Sybrand:
Before addressing individual dependabot PR's - it's easier to just run a quick ```poetry update```


conbrad (approved):
Some major version updates here but if it works then it works

andrea-williams (approved):

PR #921, opened 2021-04-22
Adds a database size report script and dedups common code

conbrad:
Makes it easier to check the size of database tables

### DIFF
--- a/api/scripts/database_size_report.sh
+++ b/api/scripts/database_size_report.sh
@@ -1,0 +1,38 @@
+#!/usr/bin/env bash
+
+#   Runs a database table size query for all tables. 
+#   Query is from: https://wiki.postgresql.org/wiki/Disk_Usage (General Table Size Information)
+#
+# Usage:
+#
+#   PROJECT=OPENSHIFT_PROJECT_NAME POD=OPENSHIFT_POD DATABASE=POSTGRESQL_DATABASE ${THIS_FILE}
+#
+# Example:
+#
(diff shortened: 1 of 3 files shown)


api/scripts/database_size_report.sh
```diff
@@ -0,0 +1,38 @@
...
+set -euo pipefail
+
+SCRIPT_DIR="$(dirname "$0")"
+
+source "$SCRIPT_DIR/common_oc_checks.sh"
+
+# From https://wiki.postgresql.org/wiki/Disk_Usage
+SIZE_REPORT_QUERY="SELECT *, pg_size_pretty(total_bytes) AS total
+    , pg_size_pretty(index_bytes) AS index
+    , pg_size_pretty(toast_bytes) AS toast
```
andrea-williams, line 23: what are toast bytes?

andrea-williams (approved):

api/scripts/database_size_report.sh
conbrad: Postgres does not support physical row data that crosses the size of a data block. To enable values larger than data blocks it breaks them up into smaller pieces and are stored in the TOAST table that is related to the original table.

https://wiki.postgresql.org/wiki/TOAST

Sybrand: @conbrad could you add that comment into the script please?

Sybrand (approved):
Useful. If you could just add the toast comment into the code that would be nice.

PR #928, opened 2021-04-26
Story/comparison table/655

Sybrand:
https://wps-pr-928.apps.silver.devops.gov.bc.ca/morecast

two stations: https://wps-pr-928.apps.silver.devops.gov.bc.ca/morecast?codes=322,838

- Changed api endpoint /observations to return 10 days of data after the date of interest.
- Show station comparison table.
- Side panel minor re-factor


andrea-williams (commented):
Submitting my pending comments before refreshing the page for new code...

web/src/utils/date.ts
```diff
@@ -26,3 +33,18 @@ export const formatMonthAndDay = (month: number, day: number): string =>
   DateTime.fromObject({ month, day }).toFormat('d LLLL')
 
 export const suppressMilliInISO = (iso: string): string => iso.replace(/\.\d{0,3}/, '') // Using RegExp to remove the "." and milliseconds
+
+export const formatDateInUTC0 = (dtISO: string): string => {
+  // Given an ISO formated datetime string, return a ISO formatted datetime string for NOON UTC of that day.
+  const dtDateTime = DateTime.fromISO(dtISO).setZone(`UTC${PST_UTC_OFFSET}`)
+  const dtJS = dtDateTime.toJSDate()
+  // We want today to stay today (e.g., if it's 11pm in PST, we know it's already tomorrow in UTC, but
+  // we want the noon time for today in PST, not today in UTC, which is tomorrow. (today!)
```
andrea-williams, line 42: Brain broken.

web/src/utils/date.ts
```diff
@@ -26,3 +33,18 @@ export const formatMonthAndDay = (month: number, day: number): string =>
   DateTime.fromObject({ month, day }).toFormat('d LLLL')
 
 export const suppressMilliInISO = (iso: string): string => iso.replace(/\.\d{0,3}/, '') // Using RegExp to remove the "." and milliseconds
+
+export const formatDateInUTC0 = (dtISO: string): string => {
```
andrea-williams, line 37: what's the significance of the `0` at the end of the function name?

api/app/db/crud/weather_models.py
```diff
@@ -102,17 +102,19 @@ def get_or_create_prediction_run(session, prediction_model: PredictionModel,
...
-def get_grid_for_coordinate(session: Session,
-                            prediction_model: PredictionModel,
-                            coordinate) -> PredictionModelGridSubset:
-    """ Given a specified coordinate and model, return the appropriate grid. """
+def get_grids_for_coordinate(session: Session,
+                             prediction_model: PredictionModel,
+                             coordinate) -> PredictionModelGridSubset:
+    """ Given a specified coordinate and model, return the appropriate grids.
+    There should only every be one grid per coordinate - but it's conceivable that there are more than one.
+    """
```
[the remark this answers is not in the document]
Sybrand: Some possible scenarios 1) two cronjobs for the same model run in parallel - they could both create a grid with the same geometry ; this shouldn't happen, but can (it happened on my local machine) 2) the raster data from env canada could change slightly, resulting in a slightly different geometry (unlikely, but possible) 3) some code change to gdal/postgres results in a minor floating point calculation change, moving the grid ever so ever so slightly.

web/src/features/fireWeather/components/tables/StationComparisonTable.tsx
```diff
@@ -0,0 +1,482 @@
...
+  },
+  windDirectionValue: {
+    whiteSpace: 'nowrap'
+  },
+  precipitationValue: {
+    whiteSpace: 'nowrap'
+  },
+  darkColumn: {
+    backgroundColor: '#fafafa',
+    padding: '1px',
```
andrea-williams, line 54: why does `lightColumn` have `padding: '2px'` but `darkColumn` only has `padding: 1px`?

api/app/weather_models/env_canada.py
```diff
@@ -597,46 +596,44 @@ def _process_model_run_for_station(self,
         # Lookup the grid our weather station is in.
         logger.info("Getting grid for coordinate %s and model %s",
                     coordinate, model_run.prediction_model)
-        grid = get_grid_for_coordinate(
+        # There should never be more than one grid per model - but it can happen.
+        # TODO: Re-factor away the need for the grid table entirely.
```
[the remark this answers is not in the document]
Sybrand: I thought I had one, created on now (hope it's not a duplicate): https://app.zenhub.com/workspaces/wildfire-predictive-services-5e321393e038fba5bbe203b8/issues/bcgov/wps/982

web/src/features/fireWeather/components/tables/StationComparisonTable.tsx
```diff
@@ -0,0 +1,482 @@
...
+interface AccumulatedPrecipitation {
+  precipitation: number | undefined
+  modelValues: ModelValue[]
+}
+
+const calculateAccumulatedPrecip = (
+  noonDate: string,
+  collection: ModelValue[] | undefined
+): AccumulatedPrecipitation | undefined => {
+  // We are calculating the accumulated precipitation from 24 hours before noon.
```
andrea-williams, line 92: this is where I'm still fuzzy about the precip calculations, and it might be relevant to the ticket I'm working on now. You're calculating the accum precip for the last 24 hours before noon for a model run (so accum precip for May 4 at noon would be all precip between May 3 1pm - May 4 noon), but human forecasters write forecasts for the accum precip for ... all of May 4? Or just the first 12 hours of May 4?

web/src/features/fireWeather/components/tables/StationComparisonTable.tsx
```diff
@@ -0,0 +1,482 @@
...
+          precip.modelValues.push(value)
+        }
+      }
+    })
+    return precip
+  }
+  return undefined
+}
+
+type TemperatureSourceType = NoonForecastValue | ObservedValue | ModelValue | undefined
```
andrea-williams, line 120: any special reason why you created a special type for `TemperatureSourceType` but not for the other weather variables?

web/src/features/fireWeather/components/tables/StationComparisonTable.tsx
```diff
@@ -0,0 +1,482 @@
...
+): ModelValue | undefined => {
+  return collection?.find((item: ModelValue) => item.datetime === noonDate)
+}
+
+interface AccumulatedPrecipitation {
+  precipitation: number | undefined
+  modelValues: ModelValue[]
+}
+
+const calculateAccumulatedPrecip = (
```
conbrad, line 88: Can we pull this out and write tests for this?

conbrad (approved):
LGTM, just a minor ask and cypress test we spoke about offline.

conbrad:
Oh and non-4K monitor friendliess :D

web/src/features/fireWeather/components/tables/StationComparisonTable.tsx
```diff
@@ -0,0 +1,482 @@
...
+                </TableCell>
+                <TableCell className={classes.lightColumnHeader} colSpan={5}>
+                  Precipitation
+                </TableCell>
+                <TableCell className={classes.darkColumnHeader}>Dew point</TableCell>
+              </TableRow>
+              <TableRow>
+                <TableCell>Weather Stations</TableCell>
+                {['temp', 'rh', 'wind', 'precip'].map((value, index) => {
+                  return SubHeadings(index, classes.darkColumnHeader)
```
andrea-williams: very low-priority and pedantic, but the subheadings always have dark backgrounds, whereas I would've expected the subheading background to match its corresponding heading's background

andrea-williams:
> Oh and non-4K monitor friendliness :D

I don't have a 4K monitor and Observed Dewpoint displayed fine for me 😕

conbrad:
> > Oh and non-4K monitor friendliness :D
> 
> I don't have a 4K monitor and Observed Dewpoint displayed fine for me 😕

Been looking into this cypress tool to see how feasible snapshot testing would be given this has a come up a couple times now with inconsistencies: https://medium.com/norwich-node-user-group/visual-regression-testing-with-cypress-io-and-cypress-image-snapshot-99c520ccc595

web/src/utils/date.ts
Sybrand: I want an ISO string that represents the date in terms of hours offset from UTC, in this case, with an offset of 0 hours.

I agree the function name is no good. Ideas?

web/src/features/fireWeather/components/tables/StationComparisonTable.tsx
Sybrand: I still need to talk to a forecaster about this - it's super confusing. (I also need to check - I think I'm doing observed wrong here)

In terms of observed, we can only report on the last 24 hours (being as we can't observe the future).
This limitation makes me believe that surely you need to compare forecast and model date for the last 24 hours - otherwise what are you comparing?

web/src/features/fireWeather/components/tables/StationComparisonTable.tsx
Sybrand: Sonarcloud or something complained that with that many types it was too long to stick straight into the function definition like the others.

web/src/features/fireWeather/components/tables/StationComparisonTable.tsx
Sybrand: Ambrose where are you!

web/src/utils/date.ts
andrea-williams: if it's an offset of 0 hours, wouldn't `formatDateInUTC` describe the functionality accurately?

Sybrand: I wish. But no. You can omit the 00:00 part, and just say Z. Both are in UTC. I need the one with 00:00.

web/src/features/fireWeather/components/tables/StationComparisonTable.tsx
andrea-williams: Yes, this is directly relevant to my current ticket.

I was working with the understanding (really just an assumption) that forecasters make precip forecasts for the total accum precip for the 24-hour period of the forecast day. But on second thoughts that's kind of contradictory to the premise of the rest of their forecast, which is to describe what's going to be happening at noon. So yeah, I think we need to clarify this ASAP.

web/src/utils/date.ts
andrea-williams: `formatDateInUTC00Suffix`? I don't know. I can't think of an elegant name.

andrea-williams (approved):

web/src/utils/date.ts
Sybrand: I love that name.

web/src/features/fireWeather/components/tables/StationComparisonTable.tsx
Sybrand: Because bad. Fixed.

PR #922, opened 2021-04-22
Story/expand side panel/822

conbrad:
https://wps-pr-922.apps.silver.devops.gov.bc.ca/morecast

- Adds `ExpandableContainer` component that holds the `SidePanel`
- Refactors `SidePanel` to remove concept of open/close and adds it to `ExpandableContainer` which contains the expand/collapse/close controls and functionality
- Updates `npm` packages with `yarn upgrade`

Collapsed:
Screen Shot 2021-04-23 at 2 45 46 PM

Expanded:
Screen Shot 2021-04-23 at 11 51 58 AM

PR #943, opened 2021-04-27
yarn upgrade with latest flag so we stop getting spammed by dependabot

conbrad:
Package updates from https://github.com/bcgov/wps/pull/922 does not include upping minimum version, which is what dependabot is doing. This PR does that so dependabot will be satisfied for now.


Sybrand (approved):
All good with me - just give it a good testing.

conbrad:
Closing this -- too many breaking changes.

PR #951, opened 2021-04-28
Date of interest + forecast accuracy

conbrad:


web/src/features/fireWeather/pages/MoreCastPage.tsx
```diff
@@ -90,9 +91,13 @@ const MoreCastPage = () => {
       ? { redraw: true }
       : undefined
   }
-  const openSidePanel = () => {
-    setShowSidePanel(true)
-    setSidePanelWidth(PARTIAL_WIDTH)
+  const openSidePanel = (openOrClose: boolean) => {
```
andrea-williams, line 94: I'm a little confused by the name `openOrClose` - is it `true` when the side panel is open and `false` when the side panel is collapsed?

conbrad: I'm treating the function name as kind of a question so `openSidePanel(true)` means open the side panel. Opposite for false. I'll try to come up with a more readable term for it.

conbrad: Maybe `shouldOpenSidePanel` is better.

PR #953, opened 2021-04-29
Fix tileLayer excessive re-rendering

conbrad:
Keep one copy of the tile layer source once loaded. Our source does not change.

https://openlayers.org/en/latest/examples/reusable-source.html


conbrad:
Sanity checked this locally with latest data from today.

andrea-williams (approved):
Tested it out in the PR env and all looks good.

Sybrand (approved):
Can you just please add a comment explaining why it has to be a const.

PR #960, opened 2021-04-30
Story/stations from wfwx/918

andrea-williams:
https://wps-pr-960.apps.silver.devops.gov.bc.ca/morecast


api/app/stations.py
```diff
@@ -71,12 +71,12 @@ def _set_weather_variables(station_properties: DetailedWeatherStationProperties,
             setattr(weather_variables, variable_name, value)
 
 
-def _get_detailed_stations(time_of_interest: datetime):
+async def _get_detailed_stations(time_of_interest: datetime):
```
conbrad, line 74: Looks like SonarCloud complains about no tests for this function, but all that was changed is sync -> async and where it gets stations from. I believe this is tested by the endpoint tests implicitly though, right?

web/src/components/Button.tsx
```diff
@@ -6,7 +6,7 @@ import { makeStyles } from '@material-ui/core/styles'
 interface CustomProps {
   loading?: boolean
   hasSpinner?: boolean
-  spinnerColor?: string
+  spinnercolor?: string
```
conbrad, line 9: We can increase test coverage + make this component simpler by just removing `hasSpinner` and making `spinnercolor` non optional, or just removing it and using the theme color. Also make the `loading` non optional.

There are currently no instances of this component where it doesn't have a spinner / loading props -- it doesn't need to be this general until we have a use case for it.

web/src/features/fireWeather/components/graphs/TempRHGraph.tsx
```diff
@@ -88,7 +88,7 @@ const TempRHGraph = (props: Props) => {
   const forecastGraphProperties: TempRHGraphProperties = {
     values: [...noonForecasts, ...NoonForecastSummaries],
     tempName: 'Forecast Temp',
-    rhName: 'Forecash RH',
```
conbrad, line 91: :(

conbrad (approved):

api/app/stations.py
andrea-williams: yes, it's implicitly tested because we have at least 1 test for the /stations/details endpoint

web/src/components/Button.tsx
andrea-williams: The epiphany I had just now while making my coffee is that I can revert this file change in this PR, because I already made the change in my unblocked PR. Should help with increasing test coverage on this PR.

Sybrand (approved):

PR #976, opened 2021-05-03
Station data loads when page loads if code specified

andrea-williams:


Sybrand (approved):
Thanks

Sybrand (changes requested):
Things be crazy - need to re-look this in the tomorrow.

PR #983, opened 2021-05-04
modify useeffect to listen correctly

Sybrand:
Duplicating `const codesFromQuery = getStationCodesFromUrl(location.search)
    const toiFromQuery = getTimeOfInterestFromUrl(location.search)`, but it seems to be the cleanest way to listen correctly for changes.


conbrad (approved):

andrea-williams (approved):
I'm happy even if SonarCloud isn't.

PR #985, opened 2021-05-05
Display observed values in Forecasts table

andrea-williams:


conbrad (changes requested):
Code looks good just might want to fix the spacing to be uniform with existing tables

Screen Shot 2021-05-21 at 8 30 12 AM

web/src/app/store.ts
```diff
@@ -1,19 +1,19 @@
...
   // when doing c-haines development you may wish to disable immutableCheck
   // and serializableCheck (see below)
   // TODO: see if a better solution can be found: https://reactjs.org/docs/hooks-reference.html#usereducer
   // import { configureStore, Action, getDefaultMiddleware } from '@reduxjs/toolkit'
-  // middleware: () =>
-  //   getDefaultMiddleware({
-  //     immutableCheck: false,
-  //     serializableCheck: false
-  //   })
+  middleware: () =>
```
Sybrand, line 12: You may want to exclude this when you push your source up.

Sybrand (approved):
Looks good - comment re. middleware but that's it (might be the Moderna speaking here...)

web/src/features/fireWeather/components/tables/ComparisonTableRow.tsx
```diff
@@ -207,218 +218,221 @@ const formatAccumulatedPrecipitation = (
...
-            switch (source) {
-              case DataSource.Observed: {
-                switch (variable) {
-                  case WeatherVariable.Temperature: {
-                    cellContent = formatTemperature(props.observation)
-                    break
+        const colStyle = idx % 2 === 0 ? classes.darkColumn : classes.lightColumn
+        return props.subheaders[idx].map((source: DataSource) => {
+          let cellContent = undefined
+          switch (source) {
```
conbrad, line 228: Lots of cyclomatic complexity with all these switch statements (I'd argue we should turn this on as a lint check: http://pylint.pycqa.org/en/latest/technical_reference/extensions.html#design-checker)

A simpler solution I'd propose is creating a map of `[WeatherVariable, formatFn]` and just lookup the function to call with the props and class. Lmk if you want more of an explanation.

conbrad (changes requested):

Sybrand (approved):

conbrad (approved):

PR #987, opened 2021-05-05
Cronjob: Observations from API

conbrad:


api/app/wildfire_one.py
```diff
@@ -368,14 +360,15 @@ async def fetch_hourlies(
         session: ClientSession,
         raw_station: dict,
         headers: dict,
-        time_of_interest: datetime) -> WeatherStationHourlyReadings:
+        start_timestamp: int,
+        end_timestamp: int) -> WeatherStationHourlyReadings:
     """ Fetch hourly weather readings for the past 5 days for a give station
     TODO: rename this function, or specify the time range.
```
andrea-williams, line 366: looks to me like we can now delete this TODO, since you are now specifying the time range

andrea-williams (changes requested):
Completed initial review w/ some feedback. 

Ofc the `x_new` linear interpolation issue needs to be addressed before this PR can be merged

api/app/tests/fireweather_bot/test_hourly_actuals.py
```diff
@@ -1,33 +1,55 @@
...
+import asyncio
 import os
 import logging
 import pytest
 from pytest_mock import MockerFixture
 from app.fireweather_bot import hourly_actuals
+from app.schemas.observations import WeatherReading, WeatherStationHourlyReadings
+from app.schemas.stations import WeatherStation
+import nest_asyncio
+nest_asyncio.apply()
```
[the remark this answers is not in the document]
conbrad: It patches the event loop to be nested, which I need to run async tests since event loop is already used in the call stack where this is called: https://pypi.org/project/nest-asyncio/

api/app/fireweather_bot/hourly_actuals.py
```diff
@@ -44,6 +47,34 @@ def _fix_datetime(source_date):
     return python_date
 
 
+def parse_hourly_actual(station_code: int, hourly_reading: WeatherReading):
+    """ Maps WeatherReading to HourlyActual """
+    temp_valid = hourly_reading.temperature is not None
+    rh_valid = hourly_reading.relative_humidity is not None
+    wdir_valid = hourly_reading.wind_direction is not None
+    wspeed_valid = hourly_reading.wind_speed is not None
+    precip_valid = hourly_reading.precipitation is not None
```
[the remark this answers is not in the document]
conbrad: Yeah WFWX is not adding these in their response. I'll add the validations you suggested. Maybe this will fix the runtime exceptions?

api/app/fireweather_bot/hourly_actuals.py
```diff
@@ -44,7 +46,44 @@ def _fix_datetime(source_date):
     return python_date
 
 
-class HourlyActualsBot(BaseBot):
+def parse_hourly_actual(station_code: int, hourly_reading: WeatherReading):
+    """ Maps WeatherReading to HourlyActual """
+    temp_valid = hourly_reading.temperature is not None
+    rh_valid = hourly_reading.relative_humidity is not None and validate_metric(
```
Sybrand, line 52: Is there actually data that is coming through with crazy values?

api/app/fireweather_bot/hourly_actuals.py
```diff
@@ -63,58 +102,37 @@ def _get_end_date(self) -> int:
...
+            logger.info("Station hourly readings: %s", station_hourly_readings)
+
+            with app.db.database.get_write_session_scope() as session:
+                for station_hourly_reading in station_hourly_readings.result():
+                    for hourly_reading in station_hourly_reading.values:
+                        save_hourly_actual(session, parse_hourly_actual(
+                            station_hourly_reading.station.code, hourly_reading))
+
+
+async def main():
```
Sybrand, line 128: Main isn't awaited - so this script can't actually run right now. Having this be async also makes your unit tests more complicated than they need to be.

I think rather wrap the async bits in an event loop where needed and keep main normal.

Have a look in api/app/stations.py get_stations_synchronously - there's an example - basically looks like this:
```
loop = asyncio.new_event_loop()
asyncio.set_event_loop(loop)
the_result = loop.run_until_complete(run_my_async_thing_here)
```

api/app/fireweather_bot/hourly_actuals.py
```diff
@@ -63,58 +102,37 @@ def _get_end_date(self) -> int:
...
-    """ Makes the appropriate method calls in order to submit a query to the BC FireWeather Phase 1 API
-    to get hourly values for all weather stations, downloads the resulting CSV file, writes
-    the CSV file to the database, then deletes the local copy of the CSV file.
+    async def run_wfwx(self):
+        """ Entry point for running the bot """
+        with Session() as session:
+            # Authenticate with idir.
+            authenticate_session(session)
+
+            station_codes = get_station_names_to_codes()
```
Sybrand, line 111: This is a crazy rabbit spaghetti hole - get_station_names_to_codes->get_stations_synchronously->get_stations->wildfire_one.get_stations()OR_get_stations_local ; in this case it's actually loading from local, which is not what we want. I suggest unwrapping this puppy, make an explicit call to wildfire_one.get_stations(), and then do the hack we do in get_station_names_to_codes - or something along those lines. Maybe have get_station_names_to_codes_wfwx and get_station_names_to_codes_local so one know what's going on?

api/app/hourlies.py
```diff
@@ -57,15 +57,26 @@ async def fetch_hourly_readings_from_db(
     return result
 
 
+def _get_five_day_interval(time_of_interest: datetime):
```
Sybrand, line 60: Missing function comment (linter should have caught that! weird)

api/app/hourlies.py
```diff
@@ -57,15 +57,26 @@ async def fetch_hourly_readings_from_db(
...
+
+
 async def get_hourly_readings(
         station_codes: List[int],
         time_of_interest: datetime) -> List[WeatherStationHourlyReadings]:
     """ Get the hourly readings for the list of station codes provided.
     Reading 5 days before, and 10 days after the time of interest are returned.
     Depending on configuration, will read from WF1 or from local database.
     """
+    start_time_stamp, end_time_stamp = _get_five_day_interval(time_of_interest)
```
Sybrand, line 77: This behaviour is inconsistent with the method comments and what fetch_hourly_readings_from_db is doing (it's grabbing 5 days past and 10 days future)

api/app/tests/fireweather_bot/test_hourly_actuals.py
```diff
@@ -1,46 +1,115 @@
...
 import os
 import logging
 import pytest
 from pytest_mock import MockerFixture
 from app.fireweather_bot import hourly_actuals
+from app.schemas.observations import WeatherReading, WeatherStationHourlyReadings
+from app.schemas.stations import WeatherStation
+from app import wildfire_one
+import nest_asyncio
+nest_asyncio.apply()
```
Sybrand, line 13: You can lose the need for this by not making main async (see comments above)

api/app/tests/fireweather_bot/test_hourly_actuals.py
```diff
@@ -1,46 +1,115 @@
...
+    station_2 = WeatherStation(code=1, name="one", lat=0, long=0)
+    readings_1 = WeatherStationHourlyReadings(values=[reading_1, reading_2], station=station_1)
+    readings_2 = WeatherStationHourlyReadings(values=[reading_1, reading_2], station=station_2)
+
+    future_hourly = asyncio.Future()
+    future_hourly.set_result([readings_1, readings_2])
+    mocker.patch('app.wildfire_one.get_hourly_readings', return_value=future_hourly)
+
+
+@pytest.mark.asyncio
```
Sybrand, line 34: As above, I think we should simplify the test and remove dependencies.

api/app/weather_models/__init__.py
```diff
@@ -55,7 +55,14 @@ def interpolate_between_two_points(  # pylint: disable=invalid-name
     # Create interpolation function.
     function = interp1d(x_axis, y_axis, kind='linear')
     # Use iterpolation function to derive values at the time of interest.
-    return function(xn)
+
+    try:
+        result = function(xn)
+    except Exception as exception:
+        logger.error('Error interpolating, x_axis: %s, y_axis: %s, xn: %s, exception: %s',
+                     x_axis, y_axis, xn, str(exception))
```
Sybrand, line 63: logger.error(exception, exc_info=True) should give you a better stack trace if that helps?

api/pyproject.toml
```diff
@@ -29,6 +29,8 @@ jinja2 = "^2"
 sklearn = "^0"
 pytest-mock = "^3"
 numpy = "^1.19.5"
+pytest-asyncio = "^0.15.1"
```
Sybrand, line 32: We can lose these by making main not be async - then we have less dependenices.

Sybrand (changes requested):
Some concerns re. async and the spaghetti db/local/api maze.

api/app/fireweather_bot/hourly_actuals.py
Sybrand: @conbrad I've been thinking - and I think it's worth hitting out the WF1 team and confirming with them how they handle invalid values. I think in the past they've used some strange approaches, like having one invalid reading mean that other readings are also invalid. So it may be the case, that if one of the readings are wrong, we need to disregard all other readings.

andrea-williams (changes requested):

api/app/fireweather_bot/hourly_actuals.py
```diff
@@ -63,58 +94,39 @@ def _get_end_date(self) -> int:
...
-            'End_Date': self._get_end_date(),
-            'Format': 'CSV',
-            'cboFilters': config.get('BC_FIRE_WEATHER_FILTER_ID'),
-            'rdoReport': 'OSBH',
-        }
+    def run_wfwx(self):
+        """ Entry point for running the bot """
+        with Session() as session:
+            # Authenticate with idir.
+            authenticate_session(session)
```
conbrad, line 101: Not sure this is necessary anymore

api/app/fireweather_bot/common.py
```diff
@@ -28,7 +30,7 @@ class AuthenticationException(Exception):
     """ Exception thrown if there's any issue authenticating """
 
 
-def _authenticate_session(session: Session) -> Session:
+def authenticate_session(session: Session) -> Session:
```
Sybrand, line 33: You shouldn't be touching this code - this is for using idir on the old site. The API uses an entirely different mechanism.

api/app/fireweather_bot/common.py
```diff
@@ -92,6 +94,15 @@ def get_station_names_to_codes() -> dict:
     return station_codes
 
 
+def get_station_names_to_codes_wfwx() -> dict:
+    """ Helper function to create dictionary of (station_name: station_code) key-value pairs
+        specifically from WFWX API
+    """
+    loop = asyncio.new_event_loop()
+    asyncio.set_event_loop(loop)
+    return loop.run_until_complete(wildfire_one.get_stations())
```
Sybrand, line 103: I'd prefer this function indicate that it's synchronous/blocking so that someone doesn't accidentally use it in the API. Also - it's worth re-writing this code so that the auth token is only created once, and then passed to the station list as well as the hourly observations.

api/app/fireweather_bot/hourly_actuals.py
Sybrand: Most effeminately not. See comment above. I would replace this with code that gets the WFX token, and then passed it on to the station list and the hourly readings function.

api/app/fireweather_bot/hourly_actuals.py
```diff
@@ -63,58 +94,39 @@ def _get_end_date(self) -> int:
...
-
-    def construct_request_body(self):
-        return {
-            'Start_Date': self._get_start_date(),
-            'End_Date': self._get_end_date(),
-            'Format': 'CSV',
-            'cboFilters': config.get('BC_FIRE_WEATHER_FILTER_ID'),
-            'rdoReport': 'OSBH',
-        }
+    def run_wfwx(self):
```
Sybrand, line 97: If you make this method async, then you can await on station and observation list - then you only need one section of code with an event loop.

Sybrand (changes requested):
Pls. remove the call to idir auth, and re-factor code to re-use api auth token. Make things async as much as possible, and then just add an event loop in main.

api/app/fireweather_bot/hourly_actuals.py
```diff
@@ -21,100 +23,116 @@
...
-        # and rename the column appropriately
-        data_df['display_name'].replace(station_codes, inplace=True)
-        data_df.rename(columns={'display_name': 'station_code'}, inplace=True)
-
-        # write to database using _session's engine
+        return self.now
+
+    async def run_wfwx(self):
+        """ Entry point for running the bot """
+        async with ClientSession(connector=TCPConnector(limit=100)) as session:
```
Sybrand, line 106: We shouldn't need a limit this high? There should only be one connection at a time, unless we're doing parallel calls?

api/app/fireweather_bot/hourly_actuals.py
```diff
@@ -21,100 +23,116 @@
...
+        fwi=hourly_reading.fwi,
+    )
+
+
+def validate_metric(value, low, high):
+    """ Validate metric with it's range of accepted values """
+    return low <= value <= high
+
+
+def get_hourly_readings_synchronously_wfwx(station_codes, start_date, end_date):
```
Sybrand, line 71: This code isn't being called anywhere and can be deleted.

api/app/hourlies.py
```diff
@@ -57,15 +57,30 @@ async def fetch_hourly_readings_from_db(
...
         time_of_interest: datetime) -> List[WeatherStationHourlyReadings]:
     """ Get the hourly readings for the list of station codes provided.
     Reading 5 days before, and 10 days after the time of interest are returned.
     Depending on configuration, will read from WF1 or from local database.
     """
+    start_time_stamp, end_time_stamp = _get_time_interval(time_of_interest)
 
     if wildfire_one.use_wfwx():
-        return await wildfire_one.get_hourly_readings(station_codes, time_of_interest)
+        # Limit the number of concurrent connections.
```
Sybrand, line 80: I guess we're reaching a point where we can just drop this flag (not this PR! - but keep in mind).

api/app/schemas/observations.py
```diff
@@ -19,6 +19,8 @@ class WeatherReading(BaseModel):
     ffmc: float = None
     isi: float = None
     fwi: float = None
+    observation_valid_ind: bool = None
```
Sybrand, line 22: I guess we're just using the "ind" thing from wfwx? I wonder if we can't just have observation_valid - I mean - obviously it's an indicator? What do you think? Or do we try to match the wfwx convention?

api/app/weather_models/__init__.py
```diff
@@ -55,7 +55,13 @@ def interpolate_between_two_points(  # pylint: disable=invalid-name
     # Create interpolation function.
     function = interp1d(x_axis, y_axis, kind='linear')
     # Use iterpolation function to derive values at the time of interest.
-    return function(xn)
+
+    try:
+        result = function(xn)
+    except Exception as exception:
+        logger.error(exception, exc_info=True)
```
Sybrand, line 62: Do we still need this logging here? This should be caught and logged higher up no?

api/app/wildfire_one.py
```diff
@@ -393,26 +385,54 @@ async def fetch_hourlies(
...
 
-        # Iterate through "raw" station data.
-        iterator = _fetch_raw_stations_generator(
-            session, header, BuildQueryByStationCode(station_codes))
-        async for raw_station in iterator:
+    # Run the tasks concurrently, waiting for them all to complete.
+    return await asyncio.gather(*tasks)
+
+
+async def get_hourly_readings_all_stations(
```
Sybrand, line 414: Uh oh - this is going to create hundreds of simultaneous connections to the WF1 api - I think we have to rethink this. That's probably why you're hitting those limits - we need to reduce the number of calls here.

api/app/fireweather_bot/hourly_actuals.py
```diff
@@ -21,102 +25,124 @@
...
-            'Format': 'CSV',
-            'cboFilters': config.get('BC_FIRE_WEATHER_FILTER_ID'),
-            'rdoReport': 'OSBH',
-        }
+            for station_hourly_reading in station_hourly_readings:
+                for hourly_reading in station_hourly_reading.values:
+                    hourly_actual = parse_hourly_actual(
+                        station_hourly_reading.station.code, hourly_reading)
+                    if hourly_actual is not None:
+                        save_hourly_actual(session, hourly_actual)
```
Sybrand, line 123: You need to handle duplicates, right now it's failing on duplicates, killing the cronjob and logging to rocket.

Sybrand (changes requested):
Lots of comments - 2 biggish issues. 1) failing on duplicate 2) too many concurrent tcp connections flooding the wf1 api.

api/app/wildfire_one.py
conbrad: Actually this prevented issues, otherwise we run into sessions being closed.

api/app/schemas/observations.py
conbrad: I'm good either way, observation_valid is a bit more readable.

api/app/fireweather_bot/hourly_actuals.py
conbrad: We're using this session across all fetch hourly calls

api/app/tests/fireweather_bot/test_hourly_actuals.py
```diff
@@ -44,3 +93,68 @@ def mock_process_csv(self, filename: str):
...
+        isi=0.0,
+        fwi=0.0
+    )
+
+    hourly_actual = hourly_actuals.parse_hourly_actual(1, weather_reading)
+    assert hourly_actual.rh_valid is True
+    assert hourly_actual.temp_valid is True
+    assert hourly_actual.wdir_valid is True
+    assert hourly_actual.precip_valid is True
+    assert hourly_actual.wspeed_valid is True
```
andrea-williams, line 119: I don't believe these assertions of the `_valid` flags are relevant anymore?

api/app/fireweather_bot/hourly_actuals.py
```diff
@@ -21,104 +25,125 @@
...
+        loop.run_until_complete(bot.run_wfwx())
+
         # Exit with 0 - success.
         sys.exit(os.EX_OK)
-    # pylint: disable=broad-except
-    except Exception as exception:
+    except IntegrityError as exception:
+        # UniqueViolation error
+        if isinstance(exception.orig, errors.lookup('23505')):
+            logger.warning("Attempt to save duplicate hourly actual", exc_info=exception)
```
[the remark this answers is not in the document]
conbrad: Yeah, fixed in https://github.com/bcgov/wps/pull/987/commits/8c1e78dc9505494a56b16d79981c9fe805ff5534

api/app/tests/fireweather_bot/test_hourly_actuals.py
conbrad: It's possible these contradict with wf1's observation_valid_ind unless we really trust their output

andrea-williams: my understanding was that we will no longer have _valid flags for each individual wx variable anymore?

api/app/fireweather_bot/hourly_actuals.py
conbrad: Changed it to a limit of 10 which seems to work.

andrea-williams (approved):
Approving because from my perspective, all is well. I don't have enough familiarity with what's going on with the number of simultaneous connections to WFWX.

api/app/schemas/observations.py
```diff
@@ -19,6 +19,8 @@ class WeatherReading(BaseModel):
     ffmc: float = None
     isi: float = None
     fwi: float = None
+    observation_valid: bool = None
```
Sybrand, line 22: Ooooh - can you give this a try? Don't know if it will work (you have to add from typing import Union) - but if that works, then we will get rid of Sonarcloud complaining (if it works, can you do it for all the other values?) 
```suggestion
    observation_valid: Union(bool, None) = None
```

api/app/schemas/observations.py
```diff
@@ -19,6 +19,8 @@ class WeatherReading(BaseModel):
...
+    observation_valid_comment: str = None
```
Sybrand, line 23:
```suggestion
    observation_valid_comment: Union(str, None) = None
```

Sybrand (changes requested):
Code coverage and some minor comments - please reach out if you want to discuss anything. I think this is looking good now - I would like it to sit and run for a bit before we merge it, which is why I haven't just blanket approved it.

api/app/data/ecodivision_seasons.py
```diff
@@ -0,0 +1,75 @@
+""" Defines singleton class that keeps ecodivisions in memory for reuse """
+import os
+import json
+import logging
+import geopandas
+from shapely.geometry import Point
+
+
+class Singleton:
```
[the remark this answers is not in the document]
conbrad: Woops forgot the credit, thanks

Sybrand (changes requested):
We need observed dew point.

Sybrand (approved):
LGTM

PR #1003, opened 2021-05-11
Fix partial db copy and database size scripts

conbrad:
api/scripts/copy_partial_db_from_pod_to_local.sh: 
- add missing variable
- exclude large `c_haines_polygons` table
- Use custom format for `pg_dump` (tar format has 4GB limit). Custom format is suitable for `pg_restore`: https://www.postgresql.org/docs/9.4/app-pgdump.html

api/scripts/database_size_report.sh
- Remove accidental plain text link

### DIFF
--- a/api/scripts/copy_partial_db_from_pod_to_local.sh
+++ b/api/scripts/copy_partial_db_from_pod_to_local.sh
@@ -18,6 +18,8 @@
 
 set -euo pipefail
 
+SCRIPT_DIR="$(dirname "$0")"
+
 source "$SCRIPT_DIR/common_oc_checks.sh"
 
 #####################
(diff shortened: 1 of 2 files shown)


api/scripts/database_size_report.sh
```diff
@@ -25,7 +25,6 @@ source "$SCRIPT_DIR/common_oc_checks.sh"
 #
 # See https://wiki.postgresql.org/wiki/TOAST for more info.
 
-https://wiki.postgresql.org/wiki/TOAST
```
Sybrand, line 28: Can you keep the comment there please?

api/scripts/copy_partial_db_from_pod_to_local.sh
```diff
@@ -38,7 +40,7 @@ FILENAME="dump_db.tar"
 # --clean clean (drop) database objects before recreating
 # -Ft output file format (custom, directory, tar, plain text (default))
 # --exclude-table-data get the table definitions, but not the data
-PG_DUMP="pg_dump --file=/tmp/${FILENAME} --clean -Ft ${DATABASE} --exclude-table-data=model_run_grid_subset_predictions --exclude-table-data=weather_station_model_predictions"
+PG_DUMP="pg_dump --file=/tmp/${FILENAME} --clean -Fc ${DATABASE} --exclude-table-data=model_run_grid_subset_predictions --exclude-table-data=weather_station_model_predictions --exclude-table-data=c_haines_polygons"
```
Sybrand, line 43: Can you add -Fc to the comment section above?

Sybrand (approved):
Some minor changes please - also - be aware, there's a python version that actually works. The .sh file is duplicate functionality.

PR #1032, opened 2021-05-31
reduce daily backups

Sybrand:
running out of space - AGAIN


andrea-williams (approved):
I'm satisfied with the changes but it would appear that GitHub Actions is not.

PR #1044, opened 2021-06-10
C-Haines : cleanup

Sybrand:
Test: https://wps-pr-1044.apps.silver.devops.gov.bc.ca/c-haines

- drop the now redundant database tables (and associated crud).
- remove the now redundant migration code (and code coverage exclusions!).
- get rid of minio
- make a few things more async and parallel


andrea-williams (approved):
I didn't examine it closely, but whatever.

PR #1051, opened 2021-06-15
Graph popover - only show closest

Sybrand:
Test: https://wps-pr-1051.apps.silver.devops.gov.bc.ca/morecast/?codes=322

- Changed hovermode to only show closest


conbrad (approved):

andrea-williams (changes requested):
the change only applies to the TempRH Graph - from what I'm understanding of the ticket, the change should be applied to the precip & wind graphs too?

andrea-williams (approved):
As discussed, other changes to come later...

PR #1052, opened 2021-06-15
I want to do this...

Sybrand:
I want to change how we do the database stuff...

Looks like async support has just been added:
https://docs.sqlalchemy.org/en/14/orm/extensions/asyncio.html


Sybrand:
Looks like async support has just been added - rather go that route:
https://docs.sqlalchemy.org/en/14/orm/extensions/asyncio.html

PR #1053, opened 2021-06-15
Break like the wind

Sybrand:
https://youtu.be/fiaWwjZS50s?t=32

Test here: https://wps-pr-1053.apps.silver.devops.gov.bc.ca/morecast

- unit of measure now in heading
- wind speed and direction split
- re-factored to use common formatting function across tables and in graph where possible
- tinkered with spacing/maxlength/heading sizes to get unit tests to pass.


web/src/features/fireWeather/components/tables/ComparisonTableRow.tsx
```diff
@@ -93,24 +96,43 @@ const useStyles = makeStyles({
...
+  )
+}
+
+const formatWindSpeed = (
+  source: ObservedValue | ModelValue | undefined,
+  valueClassName: string[]
+): ReactElement => {
+  return (
+    <div>
+      {typeof source?.wind_speed === 'number' && (
```
conbrad, line 120: Can you simplify to remove the type check?

`formatWindSpeedValue(source?.wind_speed)}` should just work for undefined values.

web/src/features/fireWeather/components/tables/NoonForecastTable.tsx
```diff
@@ -143,17 +148,20 @@ const NoonForecastTable = (props: NoonForecastTableProps) => {
                     </TableSortLabel>
                   </TableCell>
                   {/* Temperature */}
-                  <TableCell className={classes.darkColumnHeader}>Forecast</TableCell>
-                  <TableCell className={classes.darkColumnHeader}>Observed</TableCell>
+                  <TableCell className={classes.darkColumnHeader}>FCST</TableCell>
```
conbrad, line 151: Why are these being shortened here?

web/src/features/fireWeather/components/tables/StationComparisonTable.tsx
```diff
@@ -102,26 +102,33 @@ const StationComparisonTable = (props: Props) => {
               <TableRow>
                 <TableCell></TableCell>
                 <TableCell className={classes.darkColumnHeader} colSpan={5}>
-                  Temperature
+                  Temperature &deg;C
```
conbrad, line 105: Other places use the unicode version of degree, should we be using the html symbol codes everywhere?

conbrad (changes requested):

web/src/features/fireWeather/components/tables/ComparisonTableRow.tsx
Sybrand: Depends where one does the check. formatWindSpeedValue is very simple if it doesn't have to deal with undefined. If I do source.?wind_speed - then I have to handle the case where formatWindSpeedValue accepts an undefined - and decide how to format an undefined.

web/src/features/fireWeather/components/tables/NoonForecastTable.tsx
Sybrand: Because I split speed and direction & because I've moved the unit of measure into the headings, the tables have gotten so wide they were causing tests to fail.

web/src/features/fireWeather/components/tables/StationComparisonTable.tsx
Sybrand: I honestly don't know what's best. I find that &deg; reads easier than just some obscure unicode character.

Sybrand: Oh - or do you mean it's straight up got the degree symbol in there? I kind of dislike that as well - since I suspect there's more than one unicode character that looks like a degree symbol - so how do we know which one it is?

web/src/features/fireWeather/components/tables/ComparisonTableRow.tsx
conbrad: I'd argue most of that logic should live in the formatting function since this is a component returning the view. Also makes it easier to unit test.

web/src/features/fireWeather/components/tables/NoonForecastTable.tsx
```diff
@@ -130,7 +132,10 @@ const NoonForecastTable = (props: NoonForecastTableProps) => {
                     Relative Humidity
                   </TableCell>
                   <TableCell className={classes.darkColumnHeader} colSpan={2}>
-                    Wind Speed + Direction
+                    Wind Speed
+                  </TableCell>
+                  <TableCell className={classes.darkColumnHeader} colSpan={2}>
+                    Wind Direction
```
andrea-williams, line 138: all these column headers need units now

web/src/features/fireWeather/components/tables/NoonWxValueTables.tsx
```diff
@@ -29,26 +31,29 @@ const sharedColumns: Column[] = [
...
     label: 'RH (%)',
     align: 'right',
-    format: (value: number): number => Math.round(value)
+    format: formatRelativeHumidity,
+    maxWidth: 70
   },
   {
     id: 'wind_direction',
-    label: 'Wind Dir (10m)',
+    label: 'Wind Dir (10m) °',
```
andrea-williams, line 46:
```suggestion
    label: 'Wind Dir (10m) (°)',
```

web/src/features/fireWeather/components/tables/StationComparisonTable.tsx
conbrad: I also do not know, but whatever way is chosen I think should be consistent. Some places have the straight up degree symbol, other places like here have the HTML code.

web/src/features/fireWeather/components/tables/NoonWxValueTables.tsx
andrea-williams: I was irrationally irritated by the fact that this was the only unit of measure not placed inside `( )`

web/src/features/fireWeather/components/tables/ObservationTable.tsx
```diff
@@ -25,40 +26,43 @@ export const columns: Column[] = [
...
     align: 'right',
-    maxWidth: 90,
-    format: (value: number): string => value.toFixed(TEMPERATURE_VALUES_DECIMAL)
+    maxWidth: 70,
+    format: formatTemperature
   },
   {
     id: 'wind_direction',
-    label: 'Wind Dir',
+    label: 'Wind Dir °',
```
andrea-williams, line 48:
```suggestion
    label: 'Wind Dir (°)',
```

web/src/features/fireWeather/components/tables/StationComparisonTable.tsx
Sybrand: I think the bit's where it's straight up a degree symbol isn't HTML - I'll check.

andrea-williams:
```suggestion
                  Temperature (&deg;C)
```

web/src/features/fireWeather/components/tables/StationComparisonTable.tsx
```diff
@@ -102,26 +102,33 @@ const StationComparisonTable = (props: Props) => {
...
-                  Relative Humidity
+                  Relative Humidity (%)
                 </TableCell>
                 <TableCell className={classes.darkColumnHeader} colSpan={5}>
-                  Wind Speed + Direction
+                  Wind Speed (km/h)
                 </TableCell>
                 <TableCell className={classes.lightColumnHeader} colSpan={5}>
-                  Precipitation
+                  Wind Direction &deg;
```
andrea-williams, line 114:
```suggestion
                  Wind Direction (&deg;)
```

web/src/features/fireWeather/components/tables/StationComparisonTable.tsx
```diff
@@ -102,26 +102,33 @@ const StationComparisonTable = (props: Props) => {
...
+                </TableCell>
+                <TableCell className={classes.darkColumnHeader} colSpan={5}>
+                  Precipitation (mm)
+                </TableCell>
+                <TableCell className={classes.lightColumnHeader}>
+                  Dew point &deg;C
```
andrea-williams, line 120:
```suggestion
                  Dew point (&deg;C)
```

andrea-williams (approved):
I'd like to see consistency in using `( )` with the units of measure in the column headers, and as noted, there's one table that's entirely missing the units of measure.

But I'm too lazy to request changes now and return later to Approve. So, Approved pending requested changes ;)

web/src/features/fireWeather/components/tables/ComparisonTableRow.tsx
Sybrand: Done - code is a bit cleaner then as well.

web/src/features/fireWeather/components/tables/StationComparisonTable.tsx
Sybrand: @conbrad / @andrea-williams and ideas on how to standardize on &deg; - we can't use the degree character straight up in react, because it's not valid html - how do I make the label: 'blah' section take label: 'blah &deg;' ?

andrea-williams: I think `&deg;` is the easiest-to-read HTML-compliant option. There's no way I'm going to memorize unicode numbers.

conbrad (approved):

web/src/features/fireWeather/components/tables/StationComparisonTable.tsx
Sybrand: Oh sorry  - things got confused with escaping. In places we are doing &amp;deg; -  in other places we straight up just use &deg; (see web/src/features/fireWeather/components/tables/ObservationTable.tsx ; we have code like `label: 'Dew Point (°C)',` - I don't know how to make it consistent so we use &amp;deg; in all places.

conbrad: @Sybrand have you tried any of these options?
https://stackoverflow.com/a/13094667

Sybrand: No dice. None apply.

conbrad: Strange. Well we can defer this I think if it's taking too much time / not critical.

Issue #1008, opened 2021-05-12
Rate of Spread displayed 


jakemorr:
**As a** *zone wildfire coordinator* **I want** *to see the calculated rate of spread for a given station* **So That** *I can understand the danger of a potential fire in this location better*

**Additional Context**
- Rate of Spread (m/min) calculated using R Code

**Acceptance Criteria**
- [ ] Given I am on the page, When I look at a weather station, Then I can see the rate of spread

PR #1068, opened 2021-06-22
Base image with R (#1008)

Sybrand:
Test: https://wps-pr-1068.apps.silver.devops.gov.bc.ca/hfi-calculator

- renamed our base api templates and corresponding images
- added R to docker build
- added R package 'cffdrs' in docker build
- added diagram to explain imagestream flow


openshift/wps-api-base/docker/Dockerfile
```diff
@@ -0,0 +1,55 @@
...
+RUN apt-get -y update
+RUN apt-get -y install libgdal-dev
+
+# Install R
+RUN apt-get update --fix-missing && apt-get -y install r-base
+
+# Install cffdrs
+RUN R -e "install.packages('cffdrs')"
+
+# RUN apt-get update --fix-missing && apt-get -y install r-base-dev texlive-full
```
conbrad, line 32: Commented out commands

api/Dockerfile
```diff
@@ -13,7 +13,13 @@ RUN apt-get -y install unixodbc-dev
 # Install old (2.4.*; current debian) version of gdal
 RUN apt-get -y install libgdal-dev
 
-# Update pip  
+# Install R
```
conbrad, line 16: Is this docker file intended to just be build and used locally or is it part of the openshift image build pipeline? If the latter are we installed R twice? It also gets installed in openshift/wps-api-base/docker/Dockerfile

conbrad (approved):

api/Dockerfile
Sybrand: You'll see the same command in 3 docker files - but for any one image, it's only installed once.
1) api/Dockerfile - this docker file is for local development only
2) openshift/ubuntu-base/docker/Dockerfile - this docker file only used for the c-haines cronjob (needs ubuntu for most recent gdal)
3) openshift/wps-api-base/docker/Dockerfile - this is the base image that all our PR's use. It's the fastapi docker image with our magic juice added on top.

api/.dockerignore
```diff
@@ -7,4 +7,9 @@ grib
 .vscode
 **/__pycache__
 **/*.pyc
-python_cache
\ No newline at end of file
+python_cache
+junk/**/*
+junk
```
andrea-williams, line 12: Do we want to include `junk` in this list? I presume it's a Sy-specific local folder?

.dockerignore
```diff
@@ -1,3 +1,6 @@
 **/.gitignore
 **/node_modules
-**/Dockerfile*
\ No newline at end of file
+**/Dockerfile*
+**/openshift
+**/junk
```
andrea-williams, line 5: Do we want to include junk in this list? I presume it's a Sy-specific local folder?

openshift/ubuntu-base/README.md
```diff
@@ -1,29 +1,29 @@
-# Ubuntu base image.
+# Ubuntu base image
 
 This image is currently only needed for the c-haines cronjob, as it uses a more up to date version of gdal
 that is available in the api.
```
andrea-williams, line 4:
```suggestion
than is available in the api.
```
Should "that" -> "than" ?

andrea-williams (approved):

api/.dockerignore
Sybrand: It is Sy specific. I strongly advocate that we ignore my junk folder.

.dockerignore
Sybrand: I'm happy to rename my junk folder to something that seems less Sy specific - but I do need a place to put the many gigs of data I seem to accumulate.

PR #1069, opened 2021-06-22
Story/r rate of spread/1008

conbrad:
- Hooks into R `cffdrs` library to compute rate of spread for HFI Calculator dailies.
- Assumes default values for PC, PDF, CC CBH uniformly for all fuel types for now. Once these are confirmed it should be moved to the FuelTypes table.
- Breaks up `wildfire_one` into its own module with related functions together.
- Adds install steps for installing R and `cffdrs` on mac

Sonarcloud: https://sonarcloud.io/dashboard?id=bcgov_wps&pullRequest=1069
Dev Deploy: https://wps-pr-1069.apps.silver.devops.gov.bc.ca/hfi-calculator


api/app/schemas/hfi_calc.py
```diff
@@ -23,6 +23,7 @@ class StationDaily(BaseModel):
     danger_class: Optional[int] = None
     observation_valid: Optional[bool] = None
     observation_valid_comment: Optional[str] = None
+    ros: Optional[float] = None
```
Sybrand, line 26: Just thinking of future developers - what about?:
```suggestion
    rate_of_spread: Optional[float] = None
```

api/app/utils/cffdrs.py
```diff
@@ -0,0 +1,109 @@
+""" This module contains functions for computing fire weather metrics.
+"""
+import logging
+from rpy2.robjects.packages import importr
+cffdrs = importr('cffdrs')
```
Sybrand, line 5: This may be a good candidate for a singleton pattern, if only to ensure that it's only loaded when actually needed.

api/app/utils/cffdrs.py
```diff
@@ -0,0 +1,109 @@
...
+                    }
+
+
+def rate_of_spread(fuel_type: str, isi: float, bui: float, fmc: float, sfc: float):
+    """ Computes ROS by delegating to cffdrs R package """
+    if fuel_type is None or isi is None or bui is None or sfc is None:
+        logger.error(
+            "%s fuel_type: %s. isi: %s, bui: %s, sfc: %s",
+            PARAMS_ERROR_MESSAGE, fuel_type, isi, bui, sfc)
+        return None
```
Sybrand, line 59: How about raising an exception instead of returning None? I like this level of testing that's being added, the R code is pretty loosy goosy.

api/app/utils/cffdrs.py
```diff
@@ -0,0 +1,109 @@
...
+
+def surface_fuel_consumption(fuel_type: str, bui: float, ffmc: float):
+    """ Computes SFC by delegating to cffdrs R package
+        Assumes a standard GFL of 3.5 kg/m^2.
+    """
+    if ffmc is None:
+        logger.error(
+            "%s fuel_type: %s. bui: %s, ffmc: %s",
+            PARAMS_ERROR_MESSAGE, fuel_type, bui, ffmc)
+        return None
```
Sybrand, line 86: Same comment re. throwing an exception.

web/src/api/hfiCalculatorAPI.ts
```diff
@@ -16,6 +16,7 @@ export interface StationDaily {
   bui: number
   fwi: number
   danger_class: number
+  ros: number
```
Sybrand, line 19: Same comment as above - re. acronyms
```suggestion
  rate_of_spread: number
```

Sybrand (commented):
I like it - few small comments; only strong feelings I have are around throwing exceptions in addition to logging errors instead of returning none

api/app/fireweather_bot/hourly_actuals.py
```diff
@@ -42,12 +43,12 @@ def _get_end_date(self) -> datetime:
     async def run_wfwx(self):
         """ Entry point for running the bot """
         async with ClientSession() as session:
-            header = await wildfire_one.get_auth_header(session)
+            header = await wildfire_api.get_auth_header(session)
```
andrea-williams, line 46: I'd suggest that if we renamed `wildfire_api` to `wfwx_api`, it would be clearer that we're referring to an external API, not our own.

api/app/tests/common.py
```diff
@@ -130,6 +130,28 @@ async def close(self, *args, **kwargs):
...
+    # pylint: disable=invalid-name, no-method-argument, no-self-use
+
+    def _FMCcalc(**kwargs):
+        return 1
+    # pylint: disable=invalid-name, no-method-argument, no-self-use
+
+    def _SFCcalc(**kwargs):
+        return 1
+
+# pylint: disable=unused-argument
```
andrea-williams, line 147: could you please re-format these lines so that it's clear what the pylint disables apply to? e.g., I'd expect the pylint: disable=invalid-name to be on the line immediately before the function definition, not at the end of the previous function definition. Also, I can't tell what pylint: disable=unused-argument applies to

api/app/utils/cffdrs.py
```diff
@@ -0,0 +1,122 @@
...
+        raise CFFDRSException(message)
+    # pylint: disable=protected-access, no-member, line-too-long
+    result = CFFDRS.instance().cffdrs._SFCcalc(FUELTYPE=fuel_type, BUI=bui, FFMC=ffmc, PC=1, GFL=3.5)
+    return result[0]
+
+  # Args:
+  #   LAT:    Latitude (decimal degrees)
+  #   LONG:   Longitude (decimal degrees)
+  #   ELV:    Elevation (metres)
+  #   DJ:     Day of year (offeren referred to as julian date)
```
andrea-williams, line 108:
```suggestion
  #   DJ:     Day of year (often referred to as julian date)
```

api/app/tests/common.py
conbrad: I'd love to but my autosave formatting does this!

api/app/fireweather_bot/hourly_actuals.py
conbrad: Neither are clear to me that it's external is that what `wfwx` means?

api/app/tests/common.py
conbrad: I'll try to turn off plugins to see which one is causing it.

conbrad: I opted for putting the pylint exception rules on the same line to prevent this issue. We do this in some other places in the codebase.

api/app/routers/hfi_calc.py
```diff
@@ -47,7 +46,8 @@ async def get_daily_view(response: Response,
 
     async with ClientSession() as session:
         header = await get_auth_header(session)
-        wfwx_stations = await get_wfwx_stations_from_station_codes(session, header, station_codes)
+        wfwx_stations = await app.wildfire_one.wfwx_api.get_wfwx_stations_from_station_codes(
```
andrea-williams, line 49:
```suggestion
        wfwx_stations = await get_wfwx_stations_from_station_codes(
```

andrea-williams: Minor: could make this a bit cleaner by just importing `get_wfwx_stations_from_station_codes` from `wfwx_api` (as you're already doing with a bunch of other functions in that file)

api/app/tests/fireweather_bot/test_hourly_actuals.py
```diff
@@ -62,10 +64,10 @@ def build_mock_wfwx_response(station_id: str) -> MockWFWXHourlyResponse:
     future_station_codes = asyncio.Future()
     future_station_codes.set_result([station_1, station_2])
 
-    mocker.patch('app.wildfire_one.wfwx_station_list_mapper', return_value=future_station_codes)
-    mocker.patch('app.wildfire_one.get_hourly_actuals_all_stations',
+    mocker.patch('app.wildfire_one.wfwx_api.wfwx_station_list_mapper', return_value=future_station_codes)
```
andrea-williams, line 67:
```suggestion
    mocker.patch('wfwx_api.wfwx_station_list_mapper', return_value=future_station_codes)
```

api/app/tests/fireweather_bot/test_hourly_actuals.py
```diff
@@ -62,10 +64,10 @@ def build_mock_wfwx_response(station_id: str) -> MockWFWXHourlyResponse:
...
+    mocker.patch('app.wildfire_one.wfwx_api.get_hourly_actuals_all_stations',
```
andrea-williams, line 68:
```suggestion
    mocker.patch('wfwx_api.get_hourly_actuals_all_stations',
```

andrea-williams: Minor: should be able to cut down these a little bit since you're already importing `wfwx_api`

api/app/tests/test_get_observations.py
```diff
@@ -70,7 +70,8 @@ def mock_get_session_scope(*_) -> Generator[Session, None, None]:
     else:
         logger.info('running test with WFWX set to False')
         monkeypatch.setenv("USE_WFWX", 'False')
-        monkeypatch.setattr('app.wildfire_one.get_stations_by_codes', lambda _: build_mock_stations(codes))
+        monkeypatch.setattr('app.wildfire_one.wfwx_api.get_stations_by_codes',
```
andrea-williams, line 73:
```suggestion
        monkeypatch.setattr('wfwx_api.get_stations_by_codes',
```

andrea-williams (commented):
Review in progress, just saving the comments I have so far...

api/app/tests/test_hfi_dailies.py
```diff
@@ -38,15 +45,42 @@ def mock_get_fire_centre_station_codes(__):
 def test_hfi_daily_metrics():
     """ BDD Scenario. """
 
-# pylint: disable=line-too-long
+# pylint: disable=line-too-long, unused-argument
```
andrea-williams, line 48: what part of the code does this disable belong to?

api/app/utils/cffdrs.py
```diff
@@ -0,0 +1,122 @@
...
+
+
+# Computable: SFC, FMC
+# To store in DB: PC, PDF, CC, CBH (attached to fuel type, red book)
+PARAMS_ERROR_MESSAGE = "One or more params passed to R call is None."
+
+DEFAULT_METRICS = {"PC": 1, "PDF": 1, "CC": 1, "CBH": 1}
+
+# TODO: Once PC, PDF, CC, CDH are confirmed, replace these defaults
+# and potentially store then in the DB as columns in FuelType
```
andrea-williams, line 46: are we still waiting to get these values? and how much of an effect do they have on ROS? (i.e., what I'm asking is whether the ROS we calculate and display in HFI calc is completely off because we're only using these default metrics)

api/app/wildfire_one/wfwx_api.py
```diff
@@ -0,0 +1,255 @@
...
+    for station_code in station_codes:
+        wfwx_station = station_code_dict.get(station_code)
+        if wfwx_station is not None:
+            requested_stations.append(wfwx_station)
+        else:
+            logger.error("No WFWX station id for station code: %s", station_code)
+
+    return requested_stations
+
+# pylint: disable=too-many-locals
```
andrea-williams, line 221: who does this floating disable comment belong to?

api/app/wildfire_one/schema_parsers.py
```diff
@@ -0,0 +1,203 @@
...
+        ffmc=hourly.get('fineFuelMoistureCode', None),
+        isi=hourly.get('initialSpreadIndex', None),
+        fwi=hourly.get('fireWeatherIndex', None),
+        observation_valid=hourly.get('observationValidInd'),
+        observation_valid_comment=hourly.get('observationValidComment')
+    )
+
+
+def parse_daily(raw_daily, station: WFWXWeatherStation, fuel_type: str) -> StationDaily:
+    """ Transform from the raw hourly json object returned by wf1, to our hourly object.
```
andrea-williams, line 102:
```suggestion
    """ Transform from the raw daily json object returned by wf1, to our daily object.
```

andrea-williams (approved):

andrea-williams:
My approval was premature - technical code review OK, but doesn't seem to be working in dev properly.

api/app/utils/cffdrs.py
conbrad: Still waiting. No response from SMEs yet most likely because they're literally fighting fires.

andrea-williams (approved):
Approving. Issues encountered in dev seem to be the result of WFWX, not us.

api/app/utils/cffdrs.py
```diff
@@ -0,0 +1,122 @@
...
+import logging
+import app.utils.r_importer
+from app.utils.singleton import Singleton
+
+
+logger = logging.getLogger(__name__)
+
+
+@Singleton
+class CFFDRS():
```
Sybrand, line 12: As mentioned, I don't know that this has to be a singleton, but given time constrains I don't mind if it stays this way.

Sybrand (approved):
Concern: Be very careful when merging in main to not lose changes made to wildfire one code by other PR's.

Comment is re. singleton;

PR #1073, opened 2021-06-24
Temp & RH - timestamp missing

Sybrand:
# Test
url: https://wps-pr-1073.apps.silver.devops.gov.bc.ca/morecast?codes=322
instructions: click on graph, hover over different values (look for "hovermode" option on far right to see different options")

# Changes
- added timestamp to each hover
- put timestamp and heading in transparent section
- added hover mode selection for demonstrating different options
- using redis cache for get_stations_by_codes and auth


web/src/features/fireWeather/components/graphs/TempRHGraph.tsx
```diff
@@ -1,6 +1,6 @@
 /* eslint-disable @typescript-eslint/ban-ts-comment */
 import React, { useEffect, useState } from 'react'
-import Plot from 'react-plotly.js'
+import Plot, { PlotParams } from 'react-plotly.js'
```
andrea-williams, line 3:
```suggestion
import Plot from 'react-plotly.js'
```

andrea-williams: SonarCloud is complaining that PlotParams is never used. But then I don't understand how linting passed...

web/src/features/fireWeather/components/graphs/TempRHGraph.tsx
```diff
@@ -161,13 +161,35 @@ const TempRHGraph = (props: Props) => {
 
   // Update plotly revision to trigger re-drawing of the plot
   const setRevision = useState(0)[1]
+  const [hoverMode, setHoverMode] = useState<
+    'closest' | 'x' | 'y' | 'x unified' | 'y unified' | false
```
andrea-williams, line 165: do we need `y` and `y unified` at all? We don't handle them and I wouldn't expect that they'll be at all useful in our application.

web/src/features/fireWeather/components/graphs/TempRHGraph.tsx
```diff
@@ -161,13 +161,35 @@ const TempRHGraph = (props: Props) => {
...
+
+  const handleHoverModeChange = (
+    event: React.ChangeEvent<{ name?: string; value: string }>
+  ) => {
+    switch (event.target.value) {
+      case 'closest':
+      case 'x':
+      case 'x unified':
+        setHoverMode(event.target.value)
+    }
```
andrea-williams, line 176: do we want to use Matomo to see which hover mode is used most often?

andrea-williams (approved):
LGTM

tessmgood (changes requested):
**CSS — styling and positioning** 

Can we apply the same classes to the 'hover mode' select component that are used in the MoreCast nav bar and in 90% tool? 
Inspect tool at a glance indicates we've got consistent `MuiInput...` and `MuiForm...` classes in play across MoreCast and 90% select components (yay). LMK if you need more input from me — hoping we can just copy those classes across for consistency.  

Positioning of select field — tricky!! I like your instinct but would love to try a 2 column approach (do we speak in columns and rows here?). 
Desired behaviour: large screens - 'models' and 'hover mode' on same row (positioned left), separate columns so that on smaller screens (or narrow browser width for side-by-side tasks, as JE demonstrated to me) the hover mode component sits below models. 

1073 temp rh

**Labels/microcopy**
Leave the drop-down labels with me (Hover Mode, Closest, X Unified), I will find out on Monday if those make sense to the forecasters AND will try to get input from Matt on what to in/exclude from the dropdown (Brett away) TY!!!!

conbrad (approved):

web/src/features/fireWeather/components/graphs/TempRHGraph.tsx
Sybrand: In theory, but really our sample size is so small.....

Sybrand: I'm going to skip adding instrumentation.

web/src/features/fireWeather/components/graphs/TempRHGraph.tsx
Sybrand: Yeah - I'll drop that. I added them all because they're all valid, but since we don't use them...

andrea-williams (approved):
Agree that the hover mode text might possibly be a little too big, but I don't think it's super obvious, obnoxiously in-your-face big, so whatever. Maybe our users will like having that text a little bigger so it's easier to find?

conbrad (approved):

api/app/wildfire_one.py
```diff
@@ -117,15 +117,44 @@ def use_wfwx():
...
     """ Fetch an access token for WFWX Fireweather API
     """
     logger.debug('fetching access token...')
     password = config.get('WFWX_SECRET')
     user = config.get('WFWX_USER')
     auth_url = config.get('WFWX_AUTH_URL')
-    async with session.get(auth_url, auth=BasicAuth(login=user, password=password)) as response:
-        return await response.json()
+    cache = _create_redis()
+    key = f'{auth_url}?{urlencode({user: user, password: password})}'
```
conbrad, line 135: Should we be storing these in plaintext in Redis?

conbrad: Keying by `auth_url` should be enough, no?

api/app/wildfire_one.py
```diff
@@ -117,15 +117,44 @@ def use_wfwx():
...
+    else:
+        logger.info('redis cache miss %s', auth_url)
+        async with session.get(auth_url, auth=BasicAuth(login=user, password=password)) as response:
+            response_json = await response.json()
+            try:
+                if response.status == 200:
+                    # We expire when the token expires, or 10 minutes, whichever is less.
+                    # NOTE: only caching for 10 minutes right now, since we aren't handling cases
+                    # where the token is invalidated.
+                    expires = min(response_json['expires_in'], 600)
```
conbrad, line 153: Can we put this 10 minute literal in our config? I'm afraid of us adding lots of timed eviction entries and not being able to audit them easily in the future.

api/app/wildfire_one.py
Sybrand: Re. plaintext - we're using secure redis, so I think it's ok - not sure. Easy enough to hash it though, so maybe I'll do that?
Re. auth_url - I'm considering the weird edge case where the username/password changes - I think I'll go the hashing option.

api/app/wildfire_one.py
Sybrand: Sure thing.

tessmgood (approved):
Cool! I hoped to use the exact same select style with label sitting inside top-left of select field border but I can see now that we're gonna have some weiiiird sizing issues if we re-use the CSS. 
Screen Shot 2021-06-29 at 4 38 14 PM

Let's not hold up this ticket! It works great. 
I need to work through some of these design patterns - Material.io, current tools, design file.

api/app/wildfire_one.py
Sybrand: @conbrad hashing password - probably overkill, could probably just not use username and password in key.... but just in case.

PR #1076, opened 2021-06-25
Bugfix/missing observations/1059

andrea-williams:
Modified our business logic so that an HourlyReading retrieved from WFWX API will be written to our database even if one of the weather variables observed is invalid. This is to address the situation where a number of stations weren't having their observations stored because their barometric pressure was being flagged as invalid by WFWX API. Barometric pressure is not relevant to our app, so we still want those observations.

However, if EVERY wx variable observed is invalid (temp, precip, RH, wind speed, and wind direction), the observation won't be written to our DB (since there's no info to write). 

SonarCloud report for this PR: https://sonarcloud.io/dashboard?id=bcgov_wps&pullRequest=1076
PR site: https://wps-pr-1076.apps.silver.devops.gov.bc.ca


api/app/wildfire_one.py
```diff
@@ -667,3 +673,22 @@ def parse_hourly_actual(station_code: int, hourly_reading: WeatherReading):
 def validate_metric(value, low, high):
     """ Validate metric with it's range of accepted values """
     return low <= value <= high
+
+
+def parse_observation_valid_comment(comment):
```
conbrad, line 678: I'm not sure it's wise to trust an unstructured comment message to define the structured fields of concern. I think it's best to verify against the specific valid boolean fields the API supplies.

conbrad (changes requested):

api/app/wildfire_one.py
andrea-williams: but the API no longer supplies specific valid boolean fields?

conbrad: It supplies `observationValidInd`, any other validation on our end for specifics is fine. Relying on their comment would be building on shaky grounds, it could change out from under us at any time as there's nothing defined as to what it has to be besides a string.

conbrad (approved):
👍

PR #1111, opened 2021-07-09
switch to using wfwx for station list

Sybrand:
We should be using the WF1 api to populate the station list.


conbrad (approved):
What are the cases where we still need local_storage and unspecified parameters?

Sybrand:
> What are the cases where we still need local_storage and unspecified parameters?

At this point it should only be the percentile calculator.

PR #1115, opened 2021-07-09
FireBAT - refactor (in preparation for relative error tests)

Sybrand:
# Re-factor

Oops. Slightly bigger than anticipated.

Goal was to simplify interface for "calculate_fire_behavour_advisory" in order to do unit tests. That is done, but a lot of code ended up getting moved around.

Test: https://wps-pr-1115.apps.silver.devops.gov.bc.ca/fire-behaviour-calculator

## Changes of note:
- Contract modified, date is now on base of request, instead of per station.
- API now handles multiple stations correctly (can handle duplicate station codes)
- Handle stations that come back without observations/forecasts.


api/app/utils/fba_calculator.py
```diff
@@ -12,17 +15,14 @@ class CannotCalculateFireTypeError(Exception):
...
     def __init__(self,  # pylint: disable=too-many-arguments
-                 wfwx_id: str, code: int, elevation: int, fuel_type: str,
+                 elevation: int, fuel_type: str,
                  time_of_interest: date, percentage_conifer: float,
                  percentage_dead_balsam_fir: float, grass_cure: float,
-                 crown_base_height: int, lat: float, long: float, name: str):
-        self.wfwx_id = wfwx_id
-        self.code = code
-        self.name = name
+                 crown_base_height: int, lat: float, long: float, bui: float, ffmc: float, isi: float,
```
andrea-williams, line 24: didn't we want to keep the name in the response? Doesn't that make things easier on the FE?

api/app/routers/fba_calc.py
```diff
@@ -18,26 +23,144 @@
...
+def prepare_response(
+        requested_station: StationRequest,
+        wfwx_station: WFWXWeatherStation,
+        raw_daily: dict,
+        fire_behavour_advisory: FireBehaviourAdvisory,
+        time_of_interest: date) -> StationResponse:
+    """ Construct a response object combining information from the request, the station from wf1,
+    the daily response from wf1 and the fire behaviour advisory. """
+
+    # Exract values from the daily
```
conbrad, line 35: Minor: exract -> extract

openshift/templates/global.config.yaml
```diff
@@ -93,6 +93,18 @@ parameters:
     description: How long to cache auth response for.
     required: True
     value: "600"
+  - name: REDIS_STATION_CACHE_EXPIRY
```
andrea-williams, line 96: could you also add these variables to the .env.example file please?

andrea-williams (commented):
I spent about 15 minutes reviewing it, so I wasn't looking super closely. Don't know how intense a review you want right now considering it's not a final refactor?

api/app/utils/cffdrs.py
```diff
@@ -238,7 +237,9 @@ def total_fuel_consumption(  # pylint: disable=invalid-name
   #   FI:   Fire Intensity (kW/m)
 
 
-def head_fire_intensity(station: FBACalculatorWeatherStation,
+def head_fire_intensity(fuel_type: str,
```
conbrad, line 240: Why decompose into more parameters?

api/app/utils/cffdrs.py
```diff
@@ -248,25 +249,32 @@ def head_fire_intensity(station: FBACalculatorWeatherStation,
...
-    logger.info('Calculated HFI %s', result[0])
+    # set to debug, otherwise it causes a lot of noise in the log
+    logger.debug('Calculated HFI %s', result[0])
     return result[0]
 
 
-def get_ffmc_for_target_hfi(
-        station: FBACalculatorWeatherStation, bui: float,
-        ffmc: float, ros: float, cfb: float, cfl: float, target_hfi: float):
+def get_ffmc_for_target_hfi(fuel_type: str,
```
conbrad, line 264: Why decompose into more parameters?

api/app/utils/hfi_calculator.py
```diff
@@ -6,6 +6,45 @@
 from app.db.database import get_read_session_scope
 
 
+# PC, PDF, CC, CDH from the Red Book. Assumes values of 1 CBH.
```
conbrad, line 9: An old assumption of CBH by yours truly

conbrad (approved):
LGTM just minor stuff that is somewhat subjective.

There is a lot of (IMO) overly commented code, e.g.

`            # get station information from the wfwx api
            wfwx_stations = await get_wfwx_stations_from_station_codes(session, header, unique_station_codes)`

`# get the dailies for all the stations
dailies = await get_dailies(session, header, wfwx_stations, time_of_interest)`

Some examples in `routes/fba_calc.py` where the comment describes exactly what reads from the function name.

None of the above is strongly held so I'm approving.

andrea-williams (approved):

api/app/utils/fba_calculator.py
Sybrand: It's still in the response. It's now set in the "prepare_response" function.

api/app/utils/cffdrs.py
Sybrand: Many reasons 1) I'd rather give the hfi function exactly what it needs - it's more clear what the inputs are that way. passing the entire object, one doesn't know what needs to be set or not (yes - I know some of those can be null - but it's just adding to more confusion 2) there was a circular dependency on that object, so I decided to rather just rip it out.

api/app/utils/cffdrs.py
Sybrand: Same as above.

api/app/utils/hfi_calculator.py
Sybrand: Added this comment and others to tech. debt ticket.

PR #1123, opened 2021-07-13
Story/critical hours hfi/1093

andrea-williams:
https://wps-pr-1123.apps.silver.devops.gov.bc.ca/fire-behaviour-calculator

Note this PR is just for tinkering with wind speed inputs - diurnal FFMC/critical hours has opened a gigantic can of worms that I'll address in a separate PR. Wanted to get wind speed inputs deployed ASAP.


tessmgood:
RAD!  

**Pros/cons of adding this now:**
- **pros:** EK will be happy and understand this pattern - his workflow will speed up
- **cons:** folks not using EK's workflow might get confused. By placing this in the top table, as an empty field, they might feel they have to provide a wind speed and go looking for one — they won't know that a FCST value will be provided when they hit calculate. Risk of confusion, friction, trust. Even if they've used the table before, we can't expect them to remember and connect these dots (usability heuristic: [recognition over recall](https://www.nngroup.com/articles/recognition-and-recall/)) 
  -- One solution would be to load the FCST value, but this will slow things down (performance tradeoff) 

**questions:**
- wait to launch this feature when the tables are merged? (tradeoff: not helping EK now) 
  --merged, inline table: load FCST values first, allow customization (or duplicate weather station rows to run scenarios) 
- make this field editable in the bottom, calculated table now?  
- how much will performance be impacted by loading the FCST value as a default in the top table? 

**prioritization:**
- Reach, Impact, Effort, Confidence - [RICE Prioritization](https://www.intercom.com/blog/rice-simple-prioritization-for-product-managers/)  
- Is it more important to cater to EK's workflow, or focus on broader reach, buy-in, and impact for more people across FC's? 
Note: I don't have anyone (non-EK) available for usability testing to make a quick, informed decision this morning

**product strategy:**
- DH hasn't officially launched this tool (as far as I know) - if that won't change in the short-term, we can focus on EK's workflow and use this as a sandbox?

andrea-williams:
@tessmgood a quick and easy (possible) solution I can think of is to have a little i (info) icon strategically placed somewhere to explain that wind speed is an optional input that can be used if a user wants to experiment with the wind speed's effect on the calculations, or that if left empty wind speed will auto-populate with the forecast/observed wind speed value from WFWX. Thoughts?

I technically have already added the "(Optional)" in the wind speed column label, but it's getting cut off. Will try to make the input table wider so that the full column label is visible

tessmgood:
> @tessmgood a quick and easy (possible) solution I can think of is to have a little i (info) icon strategically placed somewhere to explain that wind speed is an optional input that can be used if a user wants to experiment with the wind speed's effect on the calculations, or that if left empty wind speed will auto-populate with the forecast/observed wind speed value from WFWX. Thoughts?
> 
> I technically have already added the "(Optional)" in the wind speed column label, but it's getting cut off. Will try to make the input table wider so that the full column label is visible

@andrea-williams Perfect - if we increase width and inject a tooltip that helps - I completely missed the ellipses with optional on hover :D 
Long-term I'm reluctant to rely too heavily on tooltips (my last software company did this excessively and it became a crutch for weak UI - if you have to explain everything, your UI is broken!) but it's a fantastic interim solution. I can help with the copy - will post separately

tessmgood:
[Info icon ](https://fonts.google.com/icons?selected=Material%20Icons%20Outlined%3Ainfo)
"Leave this empty to calculate with forecasted/observed wind speed. Add a custom wind speed to influence the calculations."

andrea-williams:
awesome - thanks Tess!

I'm also curious if you have any thoughts on the UI of the results table when a user has specified wind speed. E.g., https://wps-pr-1123.apps.silver.devops.gov.bc.ca/fire-behaviour-calculator?s=322&f=c3&c=0&w=10 - the "status" says "Forecasted" but that's a lie - the wind speed isn't based on forecast, it's from user input. Do you think we need to do anything right now about the UI to indicate that, or is it a lower priority?

tessmgood:
> awesome - thanks Tess!
> 
> I'm also curious if you have any thoughts on the UI of the results table when a user has specified wind speed. E.g., https://wps-pr-1123.apps.silver.devops.gov.bc.ca/fire-behaviour-calculator?s=322&f=c3&c=0&w=10 - the "status" says "Forecasted" but that's a lie - the wind speed isn't based on forecast, it's from user input. Do you think we need to do anything right now about the UI to indicate that, or is it a lower priority?

Yep, it's a risk. My first thought: can we add a third label: forecasted, observed, **adjusted** in the status column once wind speed has been modified. Perhaps also highlighting the cell that has been modified. 
Status: "adjusted" (cell background color: grey #EDEDED) - wind speed: [custom value] (cell background color: grey #EDEDED)

tessmgood:
Lovin it, thank you!!

api/app/routers/fba_calc.py
```diff
@@ -94,16 +97,37 @@ def process_request(
...
     bui = raw_daily.get('buildUpIndex', None)
-    ffmc = raw_daily.get('fineFuelMoistureCode', None)
-    isi = raw_daily.get('initialSpreadIndex', None)
-    wind_speed = raw_daily.get('windSpeed', None)
+    temperature = raw_daily.get('temperature', None)
+    relative_humidity = raw_daily.get('relativeHumidity', None)
+    precipitation = raw_daily.get('precipitation', None)
+    # if user has specified wind speed as part of StationRequest, will need to
+    # re-calculate FFMC & ISI with modified value of wind speed
+    if requested_station.wind_speed is not None:
```
Sybrand, line 105: Flip this around please - "if requested_station.wind_speed is None" reads easier.

Sybrand (changes requested):
The add row button is broken

api/app/utils/cffdrs.py
```diff
@@ -296,11 +401,120 @@ def get_ffmc_for_target_hfi(fuel_type: str,
...
+
+
+def get_critical_hours_end(critical_ffmc: float, solar_noon_ffmc: float, critical_hour_start: float,
+                           temperature: float, relative_humidity: float, wind_speed: float, precip: float):
+    """ Returns the hour of day (on 24H clock) at which the hourly FFMC drops below
+    the threshold of critical_ffmc.
+    Should only be called if critical_hour_start is not None.
+    If diurnally-adjusted FFMC never drops below critical_ffmc in the day, will return 23.5 (11:30 pm).
+    """
+    assert critical_hour_start is not None
```
conbrad, line 459: Is this assert here just for testing? Can it be removed?

web/src/features/fireBehaviourCalculator/components/FBCResultTable.tsx
```diff
@@ -103,6 +106,20 @@ const FBCResultTable = (props: Props) => {
                 <TableCell>BUI</TableCell>
                 <TableCell>FWI</TableCell>
                 <TableCell>HFI</TableCell>
+                {/* <TableCell>
+                  Critical
+                  <br />
+                  Hours
+                  <br />
+                  (4000 HFI)
```
conbrad, line 114: Can we delete this or is it temporarily commented out?

web/src/features/fireBehaviourCalculator/components/FBCResultTable.tsx
```diff
@@ -163,6 +196,8 @@ const FBCResultTable = (props: Props) => {
                     <TableCell>
                       {stationResult.head_fire_intensity?.toFixed(DECIMAL_PLACES)}
                     </TableCell>
+                    {/* <TableCell>{stationResult.critical_hours_hfi_4000}</TableCell>
+                    <TableCell>{stationResult.critical_hours_hfi_10000}</TableCell> */}
```
conbrad, line 200: Delete or temporarily commented out?

conbrad (approved):

tessmgood (approved):
Looks awesome. 

One note - not a blocker, & I can create a ticket, but the responsive behaviour of the labels in the top table is a bit odd now. I sense it's because of the custom work we did to give wind speed an icon. There is a use case of folks using two browser windows side by side (or browser + Word) so this could happen. I think this will be fixed in the next iteration of the table anyway. 
responsive lables FireBAT

oops - the grass cure is right aligned as well, the other two labels are left aligned

web/src/features/fireBehaviourCalculator/components/FBCResultTable.tsx
andrea-williams: just temporary - will be added back in soon

web/src/features/fireBehaviourCalculator/components/FBCResultTable.tsx
andrea-williams: temp

api/app/utils/cffdrs.py
andrea-williams: it's for testing/troubleshooting purposes - will be removed in next PR once we get diurnal FFMC calcs working correctly

api/app/routers/fba_calc.py
```diff
@@ -26,6 +27,7 @@
 def prepare_response(
         requested_station: StationRequest,
         wfwx_station: WFWXWeatherStation,
+        fba_station: FBACalculatorWeatherStation,
```
Sybrand, line 30: It's so confusing tracing where values are coming from. E.g. the FFMC/ISI could be either from the raw_daily, or calculated by cffdrs using the wind speed. I don't like this at all - but don't have a good idea right.

api/app/routers/fba_calc.py
```diff
@@ -71,39 +74,68 @@ def prepare_response(
...
+    # if user has not specified wind speed as part of StationRequest, use the
+    # values retrieved from WFWX in raw_daily
+    if requested_station.wind_speed is None:
+        ffmc = raw_daily.get('fineFuelMoistureCode', None)
+        isi = raw_daily.get('initialSpreadIndex', None)
+        wind_speed = raw_daily.get('windSpeed', None)
+        if raw_daily.get('recordType').get('id') == 'ACTUAL':
+            status = 'Observed'
+        else:
+            status = 'Forecasted'
```
Sybrand, line 107: There's actual more than Observed and forecast - can you do an explicit check so that it's
`
if raw_daily.get('recordType').get('id') == 'ACTUAL':
            status = 'Observed'
elif raw_daily.get('recordType').get('id') == 'I thing it's FORECAST':
            status = 'Forecasted'
else:
           status = raw_daily.get('recordType').get('id')
`
?
Maybe it's even better just doing:
`
status = raw_daily.get('recordType').get('id')
`

Sybrand: Do we really have to match the exact spelling and capitalization? Can't we just use what the wfwx gives us?

api/app/schemas/fba_calc.py
```diff
@@ -52,16 +53,8 @@ class StationResponse(BaseModel):
...
-    ffmc_for_hfi_4000: Optional[float]
-    # HFI when FFMC is equal to value stored in ffmc_for_hfi_4000
-    # (this is used bc max. FFMC is 101. In some cases, HFI will never reach 4000 even with FFMC=101)
-    hfi_when_ffmc_equals_ffmc_for_hfi_4000: Optional[float]
-    # FFMC corresponding to an HFI of approx. 10,000
-    ffmc_for_hfi_10000: Optional[float]
-    # HFI when FFMC is equal to value stored in ffmc_for_hfi_10000
-    # (this is used bc max. FFMC is 101. In some cases, HFI will never reach 10,000 even with FFMC=101)
-    hfi_when_ffmc_equals_ffmc_for_hfi_10000: Optional[float]
+    critical_hours_hfi_4000: Optional[Tuple[float, float]] = None
```
Sybrand, line 56: f.y.i. On the API the type is string

api/app/utils/cffdrs.py
```diff
@@ -296,11 +401,120 @@ def get_ffmc_for_target_hfi(fuel_type: str,
...
                                                percentage_dead_balsam_fir,
-                                               bui, experimental_ffmc, ros, cfb, cfl, experimental_sfc)
+                                               bui, experimental_ffmc, experimental_ros,
+                                               cfb, cfl, experimental_sfc)
         error_hfi = (target_hfi - experimental_hfi) / target_hfi
 
     return (experimental_ffmc, experimental_hfi)
+
+
+def get_critical_hours_start(critical_ffmc: float, solar_noon_ffmc: float, temperature: float,
```
Sybrand, line 419: I think this method (and related) shouldn't be in this module. I think it rather belongs in app.utils.fba_calculator or something else ; I think we should be clear that this module is all about making calls to cffdrs - and not a place where we do out own calculations.

api/app/routers/fba_calc.py
Sybrand: I guess I'm saying I'd prefer we not have fba_station: FBACalculatorWeatherStation here - but I don't have a better answers. It's just all so confusing having the input to the calculation also be the input to the thing that formats the response.

Sybrand (changes requested):
Some refactoring please - see comments.

Sybrand (approved):

PR #1132, opened 2021-07-14
Add remaining tests for relative error

Sybrand:
- Test cases for all fuel types.
- Currently passing with custom limits for each metric - this should be changed in future to an absolute acceptable level.

Test:
https://wps-pr-1132.apps.silver.devops.gov.bc.ca/fire-behaviour-calculator?s=317&f=o1a&c=&w=0,s=267&f=c4&c=0&w=0


api/app/utils/redapp.py
```diff
@@ -407,6 +438,7 @@ def FBPCalculateStatisticsCOM(elevation: float,  # pylint: disable=invalid-name,
         fbp.crownBase = crown_base_height
 
         fbp.useSlope = False
+        fbp.useCrownBaseHeight = use_crown_base_height
 
         # TODO: Sybrand figure out timezone
```
andrea-williams, line 443: does Sybrand still need to figure out timezone as part of this ticket, or is that a separate PR?

andrea-williams (changes requested):
Unless it's waaayyy too much work, I'd like to see us running the tests on more than 1 station

web/src/features/fireBehaviourCalculator/FireBehaviourCalculatorPage.tsx
```diff
@@ -122,7 +122,7 @@ export const FireBehaviourCalculator: React.FunctionComponent = () => {
       />
       <Container maxWidth={'xl'}>
         <h1>
-          Fire Behavior Calculator <b style={{ color: 'Red' }}>Prototype</b>
+          Fire Behavior Analysis Tool (🔥🦇) <b style={{ color: 'Red' }}>Prototype</b>
```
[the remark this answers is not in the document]
conbrad: Thoughts on redirecting `/fire-behavior-calculator` to `/fire-behaviour-calculator`? My hand keeps typing the former to save a keypress.

api/app/utils/redapp.py
```diff
@@ -318,6 +318,30 @@ def _fbp_fuel_type_map(fuel_type: str):
     20"S-2:  White Spruce / Balsam Slash"
     21"S-3:  Coastal Cedar / Hemlock / Douglas-Fir Slash"
     """
+    # fuel_type_map = {
```
conbrad, line 321: Commented out code

api/app/tests/fba/test_fba_error.py
```diff
@@ -96,71 +114,117 @@ def given_input(elevation: float,  # pylint: disable=too-many-arguments, invalid
...
+def relative_error(metric: str, actual: float, expected: float, precision: int = 2):
     """ calculate the relative error between two values - default to precision of 2"""
     actual = round(actual, precision)
     expected = round(expected, precision)
+    if actual == expected:
+        return 0
+    if expected == 0:
+        raise RelativeErrorException(
+            f'unable to calculate relative error for {metric}; actual:{actual};expected:{expected}')
     return abs((actual-expected)/expected)
```
conbrad, line 131: Is it possible that we're dividing by zero here?

conbrad (changes requested):

api/app/tests/fba/test_fba_error.py
conbrad: I guess it doesn't matter since this is a test

api/app/tests/fba/test_fba_error.feature
```diff
@@ -4,18 +4,60 @@ Feature: /fbc/
...
-# | 780       | 50.6733333 | -120.4816667 | 2021-07-12       | 6.2        | 3              | 100                | None                       | None       | 2                 | C3        | 11.5 | 186.8 | 94.8 | 126.1 | 900.3 | 0.07                 | 0.18                     | 8.21            | 13097.75        | 0.7             |
+            | fuel_type | elevation | latitude   | longitude    | time_of_interest | wind_speed | wind_direction | percentage_conifer | percentage_dead_balsam_fir | grass_cure | crown_base_height | isi  | bui   | ffmc | dmc   | dc    | r_h1_em | r_ros_em | r_hfi_em | r_cfb_em | spreadsheet_ros | s_ros_em | spreadsheet_hfi | s_hfi_em | spreadsheet_cfb | s_cfb_em | spreadsheet_1hr | s_h1_em | note                               |
+            | C1        | 780       | 50.6733333 | -120.4816667 | 2021-07-12       | 6.2        | 3              | 100                | None                       | None       | 2                 | 11.5 | 186.8 | 94.8 | 126.1 | 900.3 | 0.09    | 0.01     | 0.02     | 0.07     | 5.22            | 0.01     | 2590.68         | 0.18     | 0.6             | 0.051    | 5.14            | 0.26    |                                    |
```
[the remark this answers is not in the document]
Sybrand: Details below - short story: I agree, but I want to get the error rate < 1% on this set first before adding more tests.

100% agree we need more representative unit tests. Right now we pretty much have a single scenario per fuel type which is definately insufficient.

HOWEVER, 2 big issues:
- Right now, we don't know what the values are supposed to be. If we could trust REDapp entirely, then we could happily use an auto-generated range of numbers and check against REDapp with some tolerance (say, required margin or error < 1% ).
- Auto-generated ranged against excel spreadsheet is probably not going to happen.

api/app/tests/fba/test_fba_error.py
```diff
@@ -96,71 +114,117 @@ def given_input(elevation: float,  # pylint: disable=too-many-arguments, invalid
...
+        'fuel_type': fuel_type
     }
 
 
-def relative_error(actual: float, expected: float, precision: int = 2):
+def relative_error(metric: str, actual: float, expected: float, precision: int = 2):
     """ calculate the relative error between two values - default to precision of 2"""
     actual = round(actual, precision)
     expected = round(expected, precision)
+    if actual == expected:
```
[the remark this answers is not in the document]
Sybrand: When actual == expected == 0, relative error is 0. BUT - actual/expected is ZeroDivision - so we do need this check.

api/app/tests/fba/test_fba_error.py
Sybrand: Check for 0 is happening above.

api/app/utils/redapp.py
Sybrand: Yeah - there's this weird lookup that's happening in REDapp, not sure we need it right now - but would like to keep the code here because it was a pain to copy.

conbrad (approved):

web/src/features/fireBehaviourCalculator/FireBehaviourCalculatorPage.tsx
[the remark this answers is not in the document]
Sybrand: People will mostly be linking in, so I don't know about adding in an extra route for fat fingered developers. That said, this comment took more typing that adding in that additional route - so if you want to add it, I won't complain.

api/app/utils/redapp.py
Sybrand: I'll do it as part of this one.

andrea-williams (approved):
Sy has successfully convinced me that we can hold off on the more extensive testing until we've figured out why some results from RedApp are waaaayyyy off from what we get.

PR #1146, opened 2021-07-19
Reimplement using react-table to fix empty state, number input, formatting issues.

conbrad:
https://wps-pr-1146.apps.silver.devops.gov.bc.ca/fire-behaviour-calculator

- Uses `react-table` for the FBC calc inputs, instead of `DataGrid` from MaterialUI -- better formatting
- Input field clear button works
- Default label prompt instead of first station in list when adding a row
- Single click field edits, even for the number arrows for number fields


web/src/features/fireBehaviourCalculator/FireBehaviourCalculatorPage.tsx
```diff
@@ -103,6 +105,15 @@ export const FireBehaviourCalculator: React.FunctionComponent = () => {
     setRowId(lastId + 1)
   }, [location]) // eslint-disable-line react-hooks/exhaustive-deps
 
+  const disableCalculateButton = some(
```
Sybrand, line 108: Maybe throw grass cure check in here on grass types?

web/src/features/fireBehaviourCalculator/fuelTypes.ts
```diff
@@ -6,7 +8,10 @@ export interface FBCFuelType {
   crown_base_height: number | undefined
 }
 export class FuelTypes {
-  static lookup(key: string): FBCFuelType {
+  static lookup(key: string | undefined): FBCFuelType | null {
+    if (isUndefined(key)) {
```
Sybrand, line 12: Cool. It's always bothered me that there's no straight forward built in method for this!

Sybrand (changes requested):
We need that grass cure check on grass types - currently failing silently if not set.

web/src/features/fireBehaviourCalculator/FireBehaviourCalculatorPage.tsx
conbrad: Grass cure is optional on the backend request model.

conbrad: Totally misunderstood your comment. Will do.

Sybrand: Unfortunately there's no way around marking grass cure as optional, as it's only required in combination with grass fuel types.
But your right contract wise. We can have the back-end return an error for that entry, instead of just breaking. That's probably even be a better way of doing it. We did use to have that check on the front end, and I think the ideal would probably be addressing it on both ends.
However we do it, I think we should have it as part of this ticket.

tessmgood (commented):
Awesome! 
Agree with Sy's comments, we need some feedback in the UI about grass cure when it's required. Can we easily implement the MaterialUI select - error style? (red)
https://material-ui.com/components/selects/#simple-select

One tradeoff is that we don't have an empty table/container now — it's potentially more confusing to get started. 
We have 3 primary buttons competing for our attention (add, delete, calculate) - unclear where to start. Can we load the page with an empty table container (or empty starter row), or just disable delete and calculate until at least one row is added? Usability concerns.

web/src/features/fireBehaviourCalculator/tableRowBuilder.tsx
```diff
@@ -0,0 +1,124 @@
+import { Checkbox, TextField } from '@material-ui/core'
+import { Autocomplete } from '@material-ui/lab'
+import { find, isEqual } from 'lodash'
+import React, { ChangeEvent } from 'react'
+import { GridMenuOption, FBCInputRow } from './components/FBCInputGrid'
+import { FBCInputGridProps } from './components/FBCInputGrid'
+import { grassCureNotSetForGrassType } from './validation'
```
andrea-williams, line 7: please change the relative imports to absolutes

andrea-williams (approved):
Couple of minor complaints, but otherwise, LGTM

tessmgood (approved):

Sybrand (approved):

PR #1148, opened 2021-07-19
C1 - error evaluation

Sybrand:
- C1 now matching REDapp, by setting date of minimum foliar content to 144 (except for fire size)
- C1 Fire size now matches REDapp
- C1 does not match coastal spreadsheet, the coastal spreadsheet calculations differ from CFFDRS - so it just won't match - ever.

Test:
https://wps-pr-1148.apps.silver.devops.gov.bc.ca/fire-behaviour-calculator?s=322&f=c1&c=0&w=NaN


conbrad (approved):

api/app/tests/fba/test_fba_error.feature
```diff
@@ -28,35 +35,35 @@ Feature: /fbc/
...
+            | fuel_type | elevation | latitude   | longitude    | time_of_interest | wind_speed | wind_direction | percentage_conifer | percentage_dead_balsam_fir | grass_cure | crown_base_height | isi  | bui   | ffmc | dmc   | dc    | r_h1_em | r_ros_em | r_hfi_em | r_cfb_em | spreadsheet_ros | s_ros_em | spreadsheet_hfi | s_hfi_em | spreadsheet_cfb | s_cfb_em | spreadsheet_1hr | s_h1_em | note                                                          |
+            | C1        | 780       | 50.6733333 | -120.4816667 | 2021-07-12       | 6.2        | 3              | 100                | None                       | None       | 2                 | 11.5 | 186.8 | 94.8 | 126.1 | 900.3 | 0.09    | 0.01     | 0.01     | 0.01     | 5.22            | 0.01     | 2590.68         | 0.16     | 0.6             | 0.017    | 5.14            | 0.26    | Spreadsheet calculations are fault, thus causing discrepancy. |
```
andrea-williams, line 39:
```suggestion
            | C1        | 780       | 50.6733333 | -120.4816667 | 2021-07-12       | 6.2        | 3              | 100                | None                       | None       | 2                 | 11.5 | 186.8 | 94.8 | 126.1 | 900.3 | 0.09    | 0.01     | 0.01     | 0.01     | 5.22            | 0.01     | 2590.68         | 0.16     | 0.6             | 0.017    | 5.14            | 0.26    | Spreadsheet calculations are faulty, thus causing discrepancy. |
```

api/app/utils/fba_calculator.py
```diff
@@ -76,11 +77,20 @@ def __init__(self,  # pylint: disable=too-many-arguments
...
     """ Transform from the raw daily json object returned by wf1, to our fba_calc.StationResponse object.
     """
-    # time of interest will be the same for all stations
+    # time of interest will be the same for all stations.
     time_of_interest = get_hour_20_from_date(station.time_of_interest)
 
-    fmc = cffdrs.foliar_moisture_content(station.lat, station.long, station.elevation,
-                                         get_julian_date(time_of_interest))
+    if station.fuel_type == 'C1':
+        # It seems REDapp uses a date of 144 for C1. The source of this value is unknown.
```
andrea-williams, line 84: were you able to determine that 144 value just by looking at the redapp code, or did you have to play around with inputs to deduce it?

andrea-williams (approved):

api/app/utils/fba_calculator.py
Sybrand: Play with inputs. I couldn't find the magic value anywhere. It could be that it only works for this test case - so really I should add a few more test cases for C1 as part of this PR. I'll do that on Tuesday.

`
FMC_Calc = jnius.autoclass('ca.cwfgm.fuel.FMC_Calc')
for i in range(0, 365):
    fmc_calc = FMC_Calc()
    fmc_calc.setAttributeValue(16801, i)
    result = fmc_calc.fmc(lat, long, elv, day_of_year)
    print(f'REDapp: {i} fmc:{result}')
`

api/app/utils/fba_calculator.py
```diff
@@ -135,8 +143,30 @@ def calculate_fire_behavour_advisory(station: FBACalculatorWeatherStation) -> Fi
...
+                                                   bui=station.bui, ws=station.wind_speed, waz=wind_azimuth,
+                                                   gs=ground_slope,
+                                                   saz=slope_azimuth, fmc=fmc, sfc=sfc,
+                                                   pc=station.percentage_conifer,
+                                                   cc=station.grass_cure,
+                                                   pdf=station.percentage_dead_balsam_fir,
+                                                   cbh=station.crown_base_height,
+                                                   isi=station.isi)
+
+    bros = cffdrs.back_rate_of_spread(fuel_type=station.fuel_type, ffmc=station.ffmc, bui=station.bui,
```
conbrad, line 160: 😄 😎

api/app/utils/fba_calculator.py
```diff
@@ -135,8 +143,30 @@ def calculate_fire_behavour_advisory(station: FBACalculatorWeatherStation) -> Fi
...
+                                      wsv=wsv,
+                                      fmc=fmc, sfc=sfc,
+                                      pc=station.percentage_conifer,
+                                      cc=station.grass_cure,
+                                      pdf=station.percentage_dead_balsam_fir,
+                                      cbh=station.crown_base_height)
+
+    sixty_minute_fire_size = get_fire_size(station.fuel_type, ros, bros, 60, cfb, lb_ratio)
```
conbrad, line 168: Does this duplicate `get_30_minutes_fire_size` and `get_60_minutes_fire_size`? Do we need all these fire size functions?

conbrad (approved):
LGTM. Is there a way to describe where all the values in the test JSON files come from? They're all magic numbers to me.

api/app/utils/fba_calculator.py
Sybrand: Updated the comment here after feedback from D.

api/app/utils/fba_calculator.py
Sybrand: It does - those ones use the spreadsheet method - I'll delete them on the next round. I've added some comments to them that I'd like in our commit history.

Sybrand: I've added comments to that effect.

Sybrand:
> LGTM. Is there a way to describe where all the values in the test JSON files come from? They're all magic numbers to me.

I'll put a comment at the top of the test_fba.py - it's really trial an error. On failure, you can get the output of the test, and evaluate if the change is correct or not, then manually update it.

PR #1163, opened 2021-07-22
Story/FBC Inline table/1141/Part 1

conbrad:
https://wps-pr-1163.apps.silver.devops.gov.bc.ca/fire-behaviour-calculator

- Merged input and output fields in the same row
- Removed the dependency on `react-table` for implementation simplicity/flexibility 
    - had to rely on JS runtime keeping object property order, since `react-table` is designed around cell value arrays, not row object fields
- Moves input fields into their own components to encapsulate logic
- Table auto-calculates on change and exit focus, as well as on enter key press
    
 TODO:
 - [x] Bug: windspeed from API does not set the input field
 - [x] Bug: Removing items keeps calculated output fields around when a new row is added
 - [x] Calculate results on input field change


web/src/features/fireBehaviourCalculator/validation.ts
```diff
@@ -7,7 +7,10 @@ import { FBCInputRow } from './components/FBCInputGrid'
  * @param row the row to validate
  * @returns false if valid, true otherwise
  */
-export const grassCureNotSetForGrassType = (row: FBCInputRow) => {
+export const grassCureNotSetForGrassType = (row: FBCInputRow): boolean => {
+  if (isUndefined(row)) {
+    return false
+  }
   if (row.fuelType === 'o1a' || row.fuelType === 'o1b') {
```
andrea-williams, line 14: at some point (I'm not sure how high a priority it is in relation to the other things we're working on right now), but we should add more validation about grass curing values. There's a hover-over text message on the grass cure field saying that values must be between 0-100, but this isn't actually enforced. 

By setting grass curing to be a negative value (with a grass fuel type), the resulting HFI is 0.0
Similarly, I tested out what the HFI result is for a grass cure of 100%, and then calculated again with grass cure of 110%, and the HFI was higher.

web/src/features/fireBehaviourCalculator/components/FBCInputGrid.tsx
```diff
@@ -74,106 +121,268 @@ const FBCInputGrid = (props: FBCInputGridProps) => {
...
-      <TableBody>
-        {rows.map((row, ri) => {
-          prepareRow(row)
-          return (
-            <TableRow {...row.getRowProps()} key={ri}>
-              {row.cells.map((cell, ci) => {
+    <div className={classes.display} data-testid={props.testId}>
+      <Paper className={classes.paper} elevation={1}>
+        <TableContainer className={classes.tableContainer}>
+          <Table stickyHeader aria-label="daily table view of HFI by planning area">
```
andrea-williams, line 161: this aria-label is incorrect (we don't use planning areas), but since we generally have been ignoring the use of aria, I'm not sure how important this is...

web/src/features/fireBehaviourCalculator/components/FBCInputGrid.tsx
```diff
@@ -74,106 +121,268 @@ const FBCInputGrid = (props: FBCInputGridProps) => {
...
+                </TableCell>
+                <TableCell>
+                  Wind
+                  <br />
+                  Dir
+                  <br />
+                  (&deg;)
+                </TableCell>
+                <TableCell className={classes.windSpeed}>
+                  {'Wind Speed (km/h) (Optional)'}
```
andrea-williams, line 219: do we still need the "(Optional)" part of the label anymore?

web/src/features/fireBehaviourCalculator/components/FBCInputGrid.tsx
```diff
@@ -74,106 +121,268 @@ const FBCInputGrid = (props: FBCInputGridProps) => {
...
+                <TableCell>ISI</TableCell>
+                <TableCell>BUI</TableCell>
+                <TableCell>FWI</TableCell>
+                <TableCell>HFI</TableCell>
+                <TableCell className={classes.criticalHoursCell}>
+                  Critical
+                  <br />
+                  Hours
+                  <br />
+                  (4000 kW/m) &dagger;
```
andrea-williams, line 241: we have &dagger; s in the column labels but we don't have the corresponding disclaimer box about critical hour calculations anymore. (Was it intentional to remove that disclaimer? If so, we should also delete the &dagger; s)

andrea-williams (commented):
I think there's some minor clean-up required (see my comment about daggers). I also have some concerns about input validation, specifically on the grass cure field, but perhaps that can wait.

Also, very minor, but how about also adding (kW/m) units on the HFI column label (we have the units for HFI in the critical hours column labels, but not in HFI)

tessmgood (approved):
YAY

web/src/features/fireBehaviourCalculator/components/FBCInputGrid.tsx
conbrad: I can update it to "Fire Behaviour Analysis table"

web/src/features/fireBehaviourCalculator/components/FBCInputGrid.tsx
conbrad: Probably not, I can remove it

web/src/features/fireBehaviourCalculator/validation.ts
andrea-williams: on further thought, I think it makes more sense to just create a bug ticket for this. We should be adding the validation on the API, too

web/src/features/fireBehaviourCalculator/components/FBCInputGrid.tsx
conbrad: Oops yeah looks like prod has removed the disclaimers altogether, I'll remove the daggers.

andrea-williams: well let's check with @tessmgood - was there an agreement made yesterday that the disclaimer box for critical hours is no longer required?

web/src/features/fireBehaviourCalculator/validation.ts
conbrad: I'll limit it to max = 100

andrea-williams (approved):

Sybrand (changes requested):
No issue with code other than addressing code smells (https://sonarcloud.io/project/issues?id=bcgov_wps&pullRequest=1163&resolved=false&types=CODE_SMELL) - but I'd expect the page to load correctly when following a bookmarked link.

Sybrand (approved):

PR #1169, opened 2021-07-27
Parse zone code for WFWX stations

andrea-williams:
Added backend logic to construct our best guess at the zone code based on info received from WFWX API. 

The /stations response object from WFWX API has a nested `zone` dictionary, including values for `alias` and `fireCentre`. (I thought that WFWX API just gave us the zone code outright, but I was wrong about that). 

Eric has provided us with the mappings of fire centre to letter, which is used as a zone "prefix", and then the zone alias is a single-digit number that comes from WFWX API.


conbrad (approved):
LGTM if assumptions around zone codes are correct.

Sybrand (approved):

api/app/wildfire_one/util.py
```diff
@@ -17,3 +17,21 @@ def is_station_valid(station) -> bool:
...
+    """
+    fire_centre_to_zone_code_prefix = {
+        25: 'K',            # Kamloops Fire Centre
+        8: 'G',             # Prince George Fire Centre
+        42: 'R',            # Northwest Fire Centre
+        2: 'C',             # Cariboo Fire Centre
+        34: 'N',            # Southeast Fire Centre
+        50: 'V'             # Coastal Fire Centre
+    }
+    if fire_centre_id not in fire_centre_to_zone_code_prefix.keys():
```
Sybrand, line 35: You can skip this check - just do
`return fire_centre_to_zone_code_prefix.get(fire_centre_id, None)`
to return None if key not there.

conbrad (approved):

PR #1171, opened 2021-07-28
Evaluate C3 accuracy

Sybrand:
- introduce "test_fba_error_random_sample" - N random iterations of FBA inputs.
- re-factor "test_fba_error" (had too much code, and very difficult to trace changes).
- fix path bug in api/app/utils/fba_calculator.py.
- attempt to fix static analysis.
- fix some deprecation warnings.
- tinker with sonarcloud trying to fix checks [failed :(]
- changed singleton to not use exception in instantiation (error logging was confounding me! - beter not to use exceptions anyway)
- fix wildfire one test
- fix FMC longitude issue!


api/app/tests/fba/c3_response.json
```diff
@@ -19,13 +19,13 @@
...
             "rate_of_spread": 7.0102999937431365,
             "fire_type": "IC",
-            "percentage_crown_fraction_burned": 0.6078811483509187,
-            "flame_length": 5.440800137427278,
-            "sixty_minute_fire_size": 7.12167902519349,
-            "thirty_minute_fire_size": 1.0808419414685175,
+            "percentage_crown_fraction_burned": 0.607295415601325,
+            "flame_length": 5.440366168631874,
+            "sixty_minute_fire_size": 7.1167278346877,
+            "thirty_minute_fire_size": 1.0795545716384733,
```
andrea-williams, line 28: just curious - what's the reason why some of these values have changed very slightly?

api/app/tests/fba/test_fba_error.feature
```diff
@@ -32,34 +29,59 @@ Feature: /fbc/
         # s_cfb_em :- Spreadsheet Crown Fraction Burned Error Margin
         # s_hfi_em :- Spreadsheet Head Fire Intensity Error Margin
```
andrea-williams, line 30: delete these 2 lines of comments? Doesn't look like the spreadsheet-related stuff is in the test values table anymore

api/app/tests/fba/test_fba_error_random_sample.py
```diff
@@ -0,0 +1,239 @@
...
+    # logging with %s became unreadable:
+    # pylint: disable=logging-fstring-interpolation
+    if comparison_value < 0:
+        logger.warning('Skipping %s! (%s) - note: %s', metric, comparison_value, note)
+    else:
+        assert python_value >= 0
+        error = relative_error(python_value, comparison_value)
+        absolute_error = abs(python_value - comparison_value)
+        # logger.info(
+        #     f'{fuel_type}: Python {python_value}, {metric} {comparison_value} ; error: {error}')
```
andrea-williams, line 64: did you intend to keep this commented out?

api/app/tests/fba/test_fba_error_random_sample.py
```diff
@@ -0,0 +1,239 @@
...
+        if error > acceptable_margin_of_error:
+            logger.error('%s %s relative error %s > %s! (python: %s, expected: %s)',
+                         fuel_type, metric, error, acceptable_margin_of_error, python_value, comparison_value)
+        if metric_error_margin > 0.01:
+            logger.debug('%s: The acceptable margin of error (%s) for %s is set too high',
+                         fuel_type, metric_error_margin, metric)
+        if error > metric_error_margin:
+            # ok - fine, but for small numbers this isn't a big deal, so let's check the absolure difference.
+            if absolute_error < 0.01:
```
andrea-williams, line 73: what's your rationale for 0.01 in `absolute_error < 0.01`? I would've more likely expected 0.1, since in the frontend we round to only 1 decimal place

api/app/tests/fba/test_fba_error_random_sample.py
```diff
@@ -0,0 +1,239 @@
...
+
+@given("""<fuel_type>, <percentage_conifer>, <percentage_dead_balsam_fir>, <grass_cure> and """
+       """<crown_base_height> for <num_iterations>""",
+       target_fixture='results')
+def given_input(fuel_type: str, percentage_conifer: float, percentage_dead_balsam_fir: float,
+                grass_cure: float, crown_base_height: float, num_iterations: int):
+    """ Take input and calculate actual and expected results """
+
+    # get python result:
+    random.seed(42)
```
andrea-williams, line 103: 42 is NOT the answer to everything!

api/app/tests/fba/test_fba_error_random_sample.py
```diff
@@ -0,0 +1,239 @@
...
+    for index in range(num_iterations):
+        # pylint: disable=invalid-name
+        elevation = random.randint(0, 4019)
+        latitude = random.uniform(45, 60)
+        longitude = random.uniform(-118, -136)
+        time_of_interest = _random_date()
+        # NOTE: For high wind speeds, the difference between REDapp and FireBAT starts exceeding
+        # tolerances. REDapp calculates it's own ISI (doesn't take the one provided by the system),
+        # but uses a different formula from CFFDRS, so the results start getting more pronounced
+        # with higher wind speeds. For that reason we limit our wind speeds to 40m, since anything
```
andrea-williams, line 114:
```suggestion
        # with higher wind speeds. For that reason we limit our wind speeds to 40 km/h, since anything
```

andrea-williams: I think you mean km/h?

andrea-williams (commented):
Review still ongoing - just saving my progress

api/app/tests/fba/test_fba_error_random_sample.py
```diff
@@ -0,0 +1,239 @@
...
+                                                   grass_cure=grass_cure,
+                                                   crown_base_height=crown_base_height,
+                                                   lat=latitude,
+                                                   long=longitude,
+                                                   bui=bui,
+                                                   ffmc=ffmc,
+                                                   isi=isi,
+                                                   wind_speed=wind_speed,
+                                                   wind_direction=wind_direction,
+                                                   temperature=temperature,  # temporary fix so tests don't break
```
andrea-williams, line 154: I don't _think_ the "temporary fix so tests don't break" comment applies anymore?

api/app/tests/fba/test_fba_error_random_sample.py
```diff
@@ -0,0 +1,239 @@
...
+            'input': {'isi': isi, 'bui': bui, 'wind_speed': wind_speed, 'ffmc': ffmc},
+            'python': python_fba,
+            'java': java_fbp,
+            'fuel_type': fuel_type}
+        )
+
+    return results
+
+
+@ then("ROS is within <ros_margin_of_error> compared to REDapp")
```
andrea-williams, line 187: file has `@given` but `@ then` - would prefer to see consistency in formatting. (My vote is for no space)

api/app/utils/fba_calculator.py
```diff
@@ -133,6 +138,9 @@ def calculate_fire_behaviour_advisory(station: FBACalculatorWeatherStation) -> F
         # The day 144 is the average date for the minimum foliar moisture content in the boreal regions of
         # Canada. It is usually pretty close maybe 7 days in either direction.
         date_of_minimum_foliar_moisture_content = 144
+    elif station.fuel_type == 'C3':
+        # TODO: Confirm if this is valid
```
andrea-williams, line 142: does this TODO still apply? I thought you had already validated it?

api/app/tests/fba/c3_response.json
Sybrand: Date of minimum foliar moisture content was set to 0, now using value from REDapp for C3 - 189

api/app/tests/fba/test_fba_error_random_sample.py
Sybrand: Will remove.

api/app/tests/fba/test_fba_error_random_sample.py
Sybrand: Arbitrary decision really; I just want to not fail when even though the relative error is big, we're actually talking tiny numbers. 
That said, with CFB, 0.01 would be a full percentage point difference. We multiply CFB by 100, so 0.01 works for CFB, but maybe not as well for others.

api/app/tests/fba/test_fba_error_random_sample.py
Sybrand: :)

api/app/tests/fba/test_fba_error_random_sample.py
Sybrand: ggrrr. stupid autoformat.

api/app/utils/fba_calculator.py
Sybrand: No - didn't get an answer from Dana, and haven't reached out to DP.

api/app/tests/fba/test_fba_error_random_sample.py
andrea-williams: ahhh right. I forgot that we're working with such a small range in CFB

andrea-williams (approved):

api/app/tests/fba/test_fba_error_random_sample.py
conbrad: I think it would be better not supplying a seed, so that it seeds based on current system time: https://docs.python.org/3/library/random.html#random.seed

Otherwise we'll have the same series of random numbers every time this is run. With a different seed every time these tests are run we're more likely to run into a test case failure.

conbrad (changes requested):

api/app/tests/fba/test_fba_error_random_sample.py
Sybrand: Not using a seed makes it practically un-reproducible though, and will result in intermittent failure. I would much rather have a deterministic set of unit tests. I certainly see the merits, but also see someone just re-running the test to get it to pass with a different set of random conditions.

conbrad: We should get the specific value that we can reproduce the failure with when the test fails.

conbrad: But if we don't want to investigate any new failures it makes sense.

conbrad (approved):

api/app/tests/fba/test_fba_error_random_sample.py
Sybrand: We could totally log the seed, so that someone can go back, plug the same seed in, and re-produce the error. Let's chat today - it's a good idea. I'm just worried that no-one is going to follow up and rather just pres "re-run".

Sybrand: Either way - I've got another PR coming in right after this one - we'll chat and then I can put a changing seed in there.

conbrad: Ok sounds good!

api/app/utils/fba_calculator.py
Sybrand: :) completely removed. happy times!

PR #1172, opened 2021-07-28
Display zone code in Firebat results table /1129

andrea-williams:
https://wps-pr-1172.apps.silver.devops.gov.bc.ca/fire-behaviour-calculator


conbrad (approved):

Sybrand (approved):

tessmgood (approved):
Random Wx testing looks good — characters and numbers matching up. Cross referenced with BCWS Weather Station Map (it doesn't confirm the zone code, so still making a few assumptions based on EK's guidance). Will ask testers for feedback.

Noticing that for stations with an N/A forecast, the table doesn't display the zone code. I'll see how testers handle this. Ok for now.

PR #1181, opened 2021-07-29
Enable codecov

conbrad:
Enables codecov for our repo: https://app.codecov.io/gh/bcgov/wps/

- Manually ran coverage report for main so branches can be compared right off the bat.
- VSCode auto removes the spaces in `yaml` files please don't make me undo
- We won't get checks until merged into `main` so it has a baseline to compare PR branches to.

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -20,155 +20,152 @@ jobs:
           # For sonar-scanner to work properly we can't use a shallow fetch.
           fetch-depth: 0
 
+      - name: Install libgdal-dev (api)
+        # The python gdal and R component relies on libgdal-dev being installed.
+        run: |
+          sudo apt-get update --fix-missing && sudo apt-get -y install libgdal-dev
+      # it would be nice to cache this!
+      # - name: Cache libgdal-dev (api)
+      #   uses: actions/cache@v2
+      #   id: cache-libgdal-dev
+      #   with:
+      #     path: "~/libgdal-dev"
+      #     key: "3.0.4"
+
       # - name: Install libgdal-dev (api)
       #   # The python gdal and R component relies on libgdal-dev being installed.
+      #   env:
+      #     CACHE_HIT: ${{steps.cache-libgdal-dev.outputs.cache-hit}}
       #   run: |
-      #     sudo apt-get update --fix-missing && sudo apt-get -y install libgdal-dev
+      #     if [[ "$CACHE_HIT" == 'true' ]]; then
+      #       sudo cp --verbose --force --recursive ~/libgdal-dev/* /
+      #     else
+      #       sudo apt-get update --fix-missing && sudo apt-get -y install libgdal-dev
+      #       mkdir -p ~/libgdal-dev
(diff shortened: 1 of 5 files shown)


codecov.yml
```diff
@@ -0,0 +1,2 @@
+comment: false
+threshold: 5%
```
andrea-williams, line 2: just curious - what are these settings for?

conbrad: https://docs.codecov.com/docs/commit-status#branches

andrea-williams (commented):
Assuming we're committing to replacing Sonarcloud with CodeCov, we should delete the Sonarcloud badge from our project readme (and replace it with CodeCov if there is a badge?)

Also would be handy to insert links to CodeCov docs in our readme

andrea-williams (approved):

README.md
```diff
@@ -82,7 +87,3 @@ Resources:
 - [ZenHub Board](https://app.zenhub.com/workspaces/wildfire-predictive-services-5e321393e038fba5bbe203b8/board?repos=235861506,237125626,237125691)
 - [PEP8](https://github.com/python/peps/blob/master/pep-0008.txt) and [PEP20](https://github.com/python/peps/blob/master/pep-0020.txt) coding conventions, but with 110 character line breaks
 - [Code of Conduct](https://github.com/bcgov/wps/blob/master/CONDUCT.md)
-
```
Sybrand, line 85: Don't drop this - we'll still keep using sonarcloud, just not for coverage.

.github/workflows/integration.yml
```diff
@@ -197,13 +194,8 @@ jobs:
         working-directory: ./web
         run: yarn run test:ci
 
-      - name: Fix code coverage paths (web)
-        # Due to the way sonarscanner mounts thing when running in docker, we need to change the coverage file.
-        working-directory: ./web/coverage
-        run: sed -i 's/\/home\/runner\/work\/wps\/wps\/web\//\/github\/workspace\//g' lcov.info
-
-      - name: SonarCloud Scan
```
Sybrand, line 205: We want to keep sonarcloud, just not the code coverage - we can move it to a different job so it happens in parallel though.

.github/workflows/integration.yml
```diff
@@ -6,7 +6,7 @@ on:
       - main
 jobs:
   lint-and-test:
-    name: Lint, Test with Coverage & SonarQube
+    name: Lint, Test with CodeCov
```
Sybrand, line 9:
```suggestion
    name: Lint, Test, Coverage with CodeCov
```

Sybrand (changes requested):

README.md
conbrad: What do we want to keep sonarcloud for?

Sybrand: Does static analysis. Catches security issues, code smells etc. etc.

Sybrand (changes requested):
You have to do the same in static_analysis.yml - that action runs on main after code has been merged.

Sybrand (approved):

PR #1189, opened 2021-07-30
Task/deploy on label

Sybrand:
# Label based deployment (a.k.a. get the hell rid of Jenkins)

So the idea is: when you set the label "pr/approved", and then re-run the "Production Deployment" jobs, it deploys to production, no problem.

Take care, everything is in place, so it will attempt to deploy!

With this in place - we won't need Jenkins at all.


.github/workflows/production_deployment.yml
```diff
@@ -0,0 +1,100 @@
+name: Production Deployment
+
+on:
+  pull_request:
+    branches:
+      - main
+jobs:
+  accept:
+    # We want to be sure that an outside collaborator can't deploy using a fork, without us marking
```
conbrad, line 9: Can a label be applied on this repo by someone not on the `wps` team?

Sybrand: Nope. People outside the team don't have rights to apply labels.

conbrad (approved):

andrea-williams (approved):

PR #1198, opened 2021-08-03
Frontend Tech Debt

conbrad:
- Rename `fbc` -> `fba` to unify fire behaviour advisory naming convention from backend
- Adds tests
- Test coverage is at 88.16% overall, 90.64% for `web/src/features/fbaCalculator`: https://codecov.io/gh/bcgov/wps/tree/0f46298688781d461f8fa39ccedd432867cdcc9c/web/src/features


web/.babelrc
```diff
@@ -0,0 +1,3 @@
+{
+  "presets": ["@babel/preset-env", "@babel/preset-react"]
```
conbrad, line 2: For `react-testing-library` to render components

web/src/features/fbaCalculator/components/GrassCureCell.tsx
```diff
@@ -54,11 +48,12 @@ const GrassCureProps = (props: GrassCureCellProps) => {
...
-    props.fbcInputGridProps.inputRows[props.rowId]
+    props.fbaInputGridProps.inputRows[props.rowId]
   )
 
   return (
     <TextField
+      data-testid={`grassCureInput-${props.rowId}`}
       type="number"
       className={props.classNameMap.grassCure}
       size="small"
```
andrea-williams, line 59: the `TextField` element in `WindSpeedCell` has an additional attribute `inputMode="numeric"`, which means that if users are using a mobile device to access the app, when they click into the cell, the number keyboard gets automatically rendered instead of the alphabetical one. Probably not very important because I don't think many (or maybe not any) of our users are accessing FireBat on mobile devices, but might be a nice-to-have just in case?

web/src/features/fbaCalculator/components/grassCureCell.test.tsx
```diff
@@ -0,0 +1,74 @@
...
+    })
+    it('should return field in error state o1b without percentage set', () => {
+      const zero1BRow = buildInputRow(FuelTypes.lookup('o1b'))
+      const props = buildProps(zero1BRow, 1)
+      render(<GrassCureCell {...props} />)
+      const renderedGrassCureInputField = screen.getByTestId(`grassCureInput-0`)
+        .firstChild?.firstChild
+      expect(renderedGrassCureInputField).toHaveValue(1)
+      expect(renderedGrassCureInputField).not.toHaveClass('Mui-error')
+    })
```
andrea-williams, line 72: FYI - if you're using the stepper buttons in the grass cure cell, the value maxes out at 100, but if you type in the value on a keyboard, you can still set values > 100 and the HFI, etc. calculations will increase accordingly (e.g., grass cure of 100 has HFI 4500; grass cure of 150 has HFI 5500).

Is this an opportunity for another unit test? A bug to be fixed in this ticket? Or in a separate ticket? Don't know. Just logging this while I still remember.

web/src/features/fbaCalculator/components/windSpeedCell.test.tsx
```diff
@@ -0,0 +1,47 @@
...
+    render(<WindSpeedCell {...props} />)
+    expect(screen.getByTestId('windSpeedInput-0').firstChild?.firstChild).toHaveValue(1)
+  })
+  it('should set calculated value if no input value', () => {
+    const row = buildInputRow(undefined)
+    const props = buildProps(row, 2)
+    render(<WindSpeedCell {...props} />)
+    expect(screen.getByTestId('windSpeedInput-0').firstChild?.firstChild).toHaveValue(2)
+  })
+})
```
andrea-williams, line 47: fun insanity testing: if I use the keyboard to type in a stupidly large wind speed value, everything blows up and there's no error message (e.g., https://wps-prod.apps.silver.devops.gov.bc.ca/fire-behaviour-calculator?s=322&f=o1b&c=100&w=11111111)

Quick bug fix for this PR, or something to add to the backlog for later? I'll leave it up to you to decide.

web/src/features/fbaCalculator/components/grassCureCell.test.tsx
conbrad: I think I can fix that issue in this PR, thanks!

andrea-williams: I left a message in the General channel on Friday because I thought that capping the grass cure at a max of 100 might have been an assumption that we made, rather than knowing for sure, and I was feeling somewhat hesitant because it is theoretically possible for RH to go above 100. EK replied saying he wasn't sure whether grass cure can go above 100. Didn't get a reply from BB or DH - it miiiiight be worth confirming before you go to the trouble of actually doing the work.

conbrad: Ok cool thanks for letting me know, I wouldn't have thought of that. I'll hold off on this change until it's confirmed then.

conbrad: Limited grass cure input to 100 both via stepper and keyboard.

web/src/features/fbaCalculator/components/windSpeedCell.test.tsx
conbrad: Limited windspeed input to 100 both via stepper and keyboard.

web/src/features/fbaCalculator/components/GrassCureCell.tsx
conbrad: Added `inputMode="numeric"` property to `GrassCureCell`.

andrea-williams:
If I type in a value > 100 into grass cure or wind speed, the calculations will still happen, I'll just see a red border around the cell. I'll defer to @tessmgood 's judgment on whether this amount of user feedback is sufficient.

andrea-williams (approved):
I'm happy with the refactors - letting Tess make the final call on error feedback to users re: >100 grass cure & wind speed values.

tessmgood (approved):

PR #1199, opened 2021-08-03
Bug/clearing windspeed/1193

conbrad:
Test: https://wps-pr-1199.apps.silver.devops.gov.bc.ca/fire-behaviour-calculator
- Have update row function dispatch API call instead of TextField components


tessmgood (approved):
LGTM

andrea-williams (approved):

PR #1212, opened 2021-08-09
Border highlighting for adjusted windspeed inputs

conbrad:
Test: https://wps-pr-1212.apps.silver.devops.gov.bc.ca/fire-behaviour-calculator

 - [x] When wind speed is adjusted, a bold/coloured style is applied to the Adjusted status label and the wind speed value that changed (usability heuristics: visibility of system status, recognition rather than recall) 
 - Based on code sandbox quick prototyping of material ui component theme overriding: https://codesandbox.io/s/material-demo-forked-rol4j

### DIFF
--- a/web/src/features/fbaCalculator/components/FBAInputGrid.tsx
+++ b/web/src/features/fbaCalculator/components/FBAInputGrid.tsx
@@ -74,7 +74,8 @@ const useStyles = makeStyles({
     maxWidth: 1900
   },
   adjustedValueCell: {
-    fontWeight: 'bold'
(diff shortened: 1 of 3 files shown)


andrea-williams (approved):
Approving, although it's slightly irritating that the corner radius on the purple box is smaller than the corner radius of the default input box, so there's this little sliver of white at every corner. (Enjoy trying to ignore this now)

Screen Shot 2021-08-09 at 5 31 53 PM

tessmgood (approved):

PR #1214, opened 2021-08-10
Calculate HFI for HFI calculator and display it

conbrad:
Test: https://wps-pr-1214.apps.silver.devops.gov.bc.ca/hfi-calculator

- [x] Given I am logged into the HFI calc, When I look at the daily view table, Then the HFI column is populated for every station displayed.

- HFI depends on `ros`, `cfl` and `sfc` calculated values which are all conditionally computed only if their respective calculations execute without a `CFFDRSException` raised. I'm noticing the `buildUpIndex` field from WFWX API is frequently `None` which causes `sfc` to be None.
- `hfi` depends on `tfc` which does not raise a `CFFDRSException` if `sfc` is `None` so I assume their is a reasonable default set for it.

### DIFF
--- a/api/app/wildfire_one/schema_parsers.py
+++ b/api/app/wildfire_one/schema_parsers.py
@@ -11,6 +11,7 @@ from app.utils.dewpoint import compute_dewpoint
 from app.data.ecodivision_seasons import EcodivisionSeasons
 from app.schemas.observations import WeatherReading
 from app.schemas.hfi_calc import StationDaily
+from app.utils.fba_calculator import calculate_cfb
 from app.utils.hfi_calculator import FUEL_TYPE_LOOKUP
 from app.utils.time import get_julian_date_now
 from app.wildfire_one.util import is_station_valid, get_zone_code_prefix
@@ -130,6 +131,7 @@ def generate_station_daily(raw_daily, station: WFWXWeatherStation, fuel_type: st
     pc = FUEL_TYPE_LOOKUP[fuel_type]["PC"]
     pdf = FUEL_TYPE_LOOKUP[fuel_type]["PDF"]
     cbh = FUEL_TYPE_LOOKUP[fuel_type]["CBH"]
+    cfl = FUEL_TYPE_LOOKUP[fuel_type]["CFL"]
 
     isi = raw_daily.get('initialSpreadIndex', None)
     bui = raw_daily.get('buildUpIndex', None)
@@ -144,6 +146,13 @@ def generate_station_daily(raw_daily, station: WFWXWeatherStation, fuel_type: st
                                 cc=cc,
                                 pdf=pdf,
                                 cbh=cbh)
+
+    cfb = calculate_cfb(fuel_type, fmc, sfc, ros, cbh)
+
+    hfi = cffdrs.head_fire_intensity(fuel_type=fuel_type,
+                                     percentage_conifer=pc,
+                                     percentage_dead_balsam_fir=pdf,
+                                     ros=ros, cfb=cfb, cfl=cfl, sfc=sfc)
     return StationDaily(
         code=station.code,
         status="Observed" if raw_daily.get('recordType', '').get('id') == 'ACTUAL' else "Forecasted",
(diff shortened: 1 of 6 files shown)


api/app/wildfire_one/schema_parsers.py
```diff
@@ -130,20 +131,50 @@ def generate_station_daily(raw_daily, station: WFWXWeatherStation, fuel_type: st
...
+            hfi = cffdrs.head_fire_intensity(fuel_type=fuel_type,
+                                             percentage_conifer=pc,
+                                             percentage_dead_balsam_fir=pdf,
+                                             ros=ros, cfb=cfb, cfl=cfl, sfc=sfc)
+    except cffdrs.CFFDRSException as exception:
+        logger.error(exception, exc_info=True)
+
     return StationDaily(
         code=station.code,
         status="Observed" if raw_daily.get('recordType', '').get('id') == 'ACTUAL' else "Forecasted",
```
andrea-williams, line 180: for discussion - FireBat just displays "ACTUAL", "FORECAST", or whatever else WFWX happens to give to us, because that's what SS requested. However, modifying "ACTUAL" -> "Observed" and "FORECAST" -> "Forecasted" matches current MoreCast behaviour. I'd imagine that there's some UX advice on this subject regarding consistent behaviour across apps.

andrea-williams: if a weather station experiences a malfunction and reports out really wacky weather values (or no values), staff can go into WFWX and manually overwrite the dodgy values. In these situations, I thiiiink the status actually shows as "MANUAL" in WFWX, which would get displayed as "Forecasted" with our current logic

tessmgood: This is a good one, it's been on my mind. Ideally we are consistent across BCWS tools:
- Forecasters see the same terminology in WildfireOne interface and MoreCast
- Fire Behaviour Analysts see the same terminology across PSU Tools (FireBAT, HFI Calc., etc.)
- - presently, FBANs would use 'forecast' in their HFI Workbook, as they copy the data from WildfireOne 

Follow-up: Not sure why we adjusted this in MoreCast - I'll ask forecasters if actual vs. observed is significant to them/why. 

Recommendation:
I'm inclined to say we stick to "ACTUAL" and "FORECAST" for nowww. 
We should pull that "MANUAL" status across if we can - nice catch.

tessmgood (approved):
Nice! *We have some HFI gaps due to WFWX issues, writing ZenHub tickets for UI error messages _and_ will raise forecasted grass cure with DH/SMEs asap for grass fuel types.

api/app/wildfire_one/schema_parsers.py
andrea-williams: what I propose is that we simply use whatever status we get from WFWX, unmodified

api/app/wildfire_one/schema_parsers.py
```diff
@@ -130,20 +131,50 @@ def generate_station_daily(raw_daily, station: WFWXWeatherStation, fuel_type: st
...
     if cc is None:
         cc = FUEL_TYPE_LOOKUP[fuel_type]["CC"]
-    ros = cffdrs.rate_of_spread(fuel_type, isi, bui, fmc, sfc, pc=pc,
-                                cc=cc,
-                                pdf=pdf,
-                                cbh=cbh)
+
+    ros = None
+    try:
+        if sfc is not None:
```
andrea-williams, line 153: why are we checking that `sfc` is not None but we don't do checks for `None` on `isi`, `bui`, `ffmc`, `fmc`? Seems like the error handling is inconsistent across variables

andrea-williams (changes requested):
Please remove the overwrite of status in StationDaily (in schema_parsers) - I think we should just display whatever status WFWX gives us.

I also have some questions about error handling when calculating things with missing variables.

api/app/wildfire_one/schema_parsers.py
conbrad: It's because this is the only variable that breaks the call to the R library. I'm basically trusting the `cffdrs` wrapper error handling verbatim.

conbrad: I'll double check the `hfi` block though.

andrea-williams (approved):

PR #1232, opened 2021-08-13
getting gdal and jnius to work

Sybrand:
- update README.md instructions (re. GDAL and Java)
- change .toml to require python 3.8.6 (matching our prod environment exactly)
- run `poetry update` to make lock file match .toml


api/README.md
```diff
@@ -41,106 +41,58 @@ For local development, you can copy .env.example to .env.
...
+##### Gdal
 
 ```bash
 brew install gdal
-
 ```
 
 ##### Poetry
 
+Try to match the latest version of python in out production environment (as of writing, API is on 3.8.6
```
conbrad, line 67: Is `poetry` choosing the minor number differently when running API vs C-Haines?

conbrad (approved):

api/README.md
Sybrand: Poetry is grabbing whatever the available python version is on the OS. API (debian) has a different version from C-Haines (ubuntu);

Sybrand: Ideally we should really get rid of the Ubuntu version! Once Debian has a new version of GDAL we can do that easily.

andrea-williams (approved):

PR #1240, opened 2021-08-16
Refactor critical hours response object

andrea-williams:
https://wps-pr-1240.apps.silver.devops.gov.bc.ca/fire-behaviour-calculator?s=239&f=c7&c=NaN&w=NaN,s=266&f=c5&c=NaN&w=NaN,s=305&f=c7&c=NaN&w=NaN,s=322&f=o1b&c=90&w=NaN,s=1082&f=c3&c=NaN&w=NaN,s=1108&f=c5&c=NaN&w=NaN,s=280&f=c7&c=NaN&w=NaN,s=306&f=c7&c=NaN&w=NaN,s=309&f=c7&c=NaN&w=NaN,s=1029&f=c7&c=NaN&w=NaN,s=1055&f=c7&c=NaN&w=NaN,s=836&f=c7&c=NaN&w=NaN,s=1399&f=c7&c=NaN&w=NaN,s=328&f=c7&c=NaN&w=58,s=334&f=c7&c=NaN&w=38,s=286&f=c7&c=NaN&w=NaN,s=298&f=c7&c=NaN&w=NaN,s=344&f=c5&c=NaN&w=NaN,s=346&f=c7&c=NaN&w=NaN,s=388&f=c7&c=NaN&w=NaN


tessmgood (approved):
rad

api/app/utils/fba_calculator.py
```diff
@@ -457,7 +457,7 @@ def get_critical_hours(  # pylint: disable=too-many-arguments
     if critical_ffmc == 0.0 and resulting_hfi >= target_hfi:
         logger.info('All hours critical for HFI %s. FFMC %s has HFI %s',
                     target_hfi, critical_ffmc, resulting_hfi)
-        return '13:00 - 7:00'
+        return (13.0, 7.00)
```
conbrad, line 460: Can we just create the `CriticalHoursHFI` object here to remove the extra tuple step later on?

api/app/utils/fba_calculator.py
```diff
@@ -468,13 +468,7 @@ def get_critical_hours(  # pylint: disable=too-many-arguments
...
         critical_ffmc, daily_ffmc, critical_hours_start)
 
-    # format result as string
-    critical_hours_start = str(critical_hours_start).replace('.', ':')
-    critical_hours_end = str(critical_hours_end).replace('.', ':')
-    critical_hours_start = critical_hours_start.replace(':0', ':00').replace(':5', ':30')
-    critical_hours_end = critical_hours_end.replace(':0', ':00').replace(':5', ':30')
-    response_string = critical_hours_start + ' - ' + critical_hours_end
-    return response_string
+    return (critical_hours_start, critical_hours_end)
```
conbrad, line 471: Similar to above

web/src/features/fbaCalculator/RowManager.ts
```diff
@@ -121,6 +124,12 @@ export class RowManager {
       case SortByColumn.HFI: {
         return _.orderBy(tableRows, 'head_fire_intensity', order)
       }
+      case SortByColumn.CriticalHours4000: {
+        return _.orderBy(tableRows, 'critical_hours_hfi_4000.start', reverseOrder)
```
conbrad, line 128: Curious why we need to reverse the order for critical hours?

web/src/features/fbaCalculator/components/FBAInputGrid.tsx
```diff
@@ -417,6 +440,12 @@ const FBAInputGrid = (props: FBAInputGridProps) => {
             </TableHead>
             <TableBody data-testid="fba-table-body">
               {sortedRows.map((row, ri) => {
+                const criticalHoursString4000 = formatCriticalHoursAsString(
```
conbrad, line 443: Might be easier to read to just call the formatting function where critical hours is displayed

web/src/features/fbaCalculator/components/FBAInputGrid.tsx
```diff
@@ -487,8 +516,8 @@ const FBAInputGrid = (props: FBAInputGridProps) => {
                     <TableCell>
                       {row.head_fire_intensity?.toFixed(DECIMAL_PLACES)}
                     </TableCell>
-                    <TableCell>{row.critical_hours_hfi_4000}</TableCell>
-                    <TableCell>{row.critical_hours_hfi_10000}</TableCell>
+                    <TableCell>{criticalHoursString4000}</TableCell>
```
conbrad, line 519: E.g. : `<TableCell>formatCriticalHoursAsString(row.critical_hours_hfi_{4000,10000})</TableCell>`

conbrad (changes requested):
Looks good, just some minor stuff

web/src/features/fbaCalculator/RowManager.ts
andrea-williams: good point - this is something that I should probably confirm with Tess. But basically, when sorting crit hours in descending order, I'd expect that the "worst" scenarios (i.e., the rows with the largest time range of critical hours) would appear first. If I don't reverse the order, then the blank critical hours get displayed first, because of how lodash compares null to a number

andrea-williams: is my logic correct @tessmgood - sorting critical hours in descending order means "put worst-case scenarios at the top of the list"?

conbrad (approved):
Assuming the sorting logic is what's desired, LGTM.

web/src/features/fbaCalculator/RowManager.ts
conbrad: So my understanding of sort is just by the start value which is how it's sorted currently. This sort doesn't order by largest / smallest range since you could have a large range regardless of when the start time is.

PR #1248, opened 2021-08-20
FireBAT table styling

andrea-williams:
- changed maxHeight of table container from 1080px to 600px
- set padding to 8px all round
- removed unnecessary info icon from Wind Speed column
- make table cells shorter (fits same number of rows into considerably smaller table container)

https://wps-pr-1248.apps.silver.devops.gov.bc.ca/fire-behaviour-calculator


tessmgood (approved):
yay

PR #1255, opened 2021-08-23
Hotfix: Gracefully handle both conider and conifer

conbrad:
Tests: 
- `m2_50conifer`: https://wps-pr-1255.apps.silver.devops.gov.bc.ca/fire-behaviour-calculator?s=322&f=m2_50conifer&c=NaN&w=NaN
- `m2_50conider`: https://wps-pr-1255.apps.silver.devops.gov.bc.ca/fire-behaviour-calculator?s=322&f=m2_50conider&c=NaN&w=NaN

- `FuelTypes.lookup` for `m2_50conider` and `m2_50conifer` return the same object keyed by `m2_50conifer`
- Removes `m2_50conider` key from fuel type map so it is not a drop down option anymore

### DIFF
--- a/web/src/features/fbaCalculator/fuelTypes.ts
+++ b/web/src/features/fbaCalculator/fuelTypes.ts
@@ -1,4 +1,4 @@
-import { isUndefined } from 'lodash'
+import { isEqual, isUndefined } from 'lodash'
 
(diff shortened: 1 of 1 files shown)


tessmgood (approved):
Looks great on the front end, ty!

PR #1268, opened 2021-08-27
Unit and cypress test coverage

conbrad:
- Adds some test coverage for `RowManager` sort fields.
- Adds build step to merge coverage files for both cypress and jest tests. Example based on `RowManager.ts` coverage differences between Cypress and Jest unit tests.

Cypress (previously only reported to CodeCov):
FECypressCoverageRowManager

Jest (previously not reported to CodeCov):
FEUnitCoverageRowManager

Merged (currently with this PR change):
merged


andrea-williams (approved):

Sybrand (approved):
Approved, but will say that web/mergeCoverage.js is kinda gross. Would be nice if we could just include an dev dependancy or something that does that. Just seems like something super obscure that's going to break again.

PR #1272, opened 2021-08-30
Row sorting: Force empty rows to the bottom

conbrad:
Test: https://wps-pr-1272.apps.silver.devops.gov.bc.ca/fire-behaviour-calculator

- Users do not want to see empty rows at the top when sorting columns, regardless of whether sort order is ascending or descending
- Lodash `orderBy` is designed to order `NaN`, nullish and symbol values at the top when ascending, bottom when descending: https://github.com/lodash/lodash/issues/4169
- Solution: Partition nullish values based on sort key into their own array, sort non-nullish values in it's own partition, then concatenate nullish partition at the end of the sorted result.

- [x] Sorting always maintains rows with empty column values at the bottom of the table

### DIFF
--- a/web/src/features/fbaCalculator/RowManager.ts
+++ b/web/src/features/fbaCalculator/RowManager.ts
@@ -50,84 +50,91 @@ export class RowManager {
     tableRows: FBATableRow[]
   ): FBATableRow[] => {
     const reverseOrder = order === 'asc' ? 'desc' : 'asc'
+    const orderByWithEmptyAtBottom = (field: string, otherOrder?: Order) => {
+      const partition = _.partition(tableRows, x => !!_.get(x, field, null))
+      return _.concat(
+        _.orderBy(partition[0], field, otherOrder ? otherOrder : order),
+        partition[1]
+      )
+    }
     switch (sortByColumn) {
       case SortByColumn.Zone: {
-        return _.orderBy(tableRows, 'zone_code', order)
+        return orderByWithEmptyAtBottom('zone_code')
       }
       case SortByColumn.Station: {
-        return _.orderBy(tableRows, 'station_name', order)
+        return orderByWithEmptyAtBottom('station_name')
       }
       case SortByColumn.Elevation: {
-        return _.orderBy(tableRows, 'elevation', order)
+        return orderByWithEmptyAtBottom('elevation')
       }
       case SortByColumn.FuelType: {
-        return _.orderBy(tableRows, 'fuel_type', order)
+        return orderByWithEmptyAtBottom('fuel_type')
       }
       case SortByColumn.GrassCure: {
-        return _.orderBy(tableRows, 'grass_cure', order)
+        return orderByWithEmptyAtBottom('grass_cure')
       }
       case SortByColumn.Status: {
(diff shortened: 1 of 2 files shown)


web/src/features/fbaCalculator/RowManager.ts
```diff
@@ -50,84 +50,91 @@ export class RowManager {
     tableRows: FBATableRow[]
   ): FBATableRow[] => {
     const reverseOrder = order === 'asc' ? 'desc' : 'asc'
+    const orderByWithEmptyAtBottom = (field: string, otherOrder?: Order) => {
+      const partition = _.partition(tableRows, x => !!_.get(x, field, null))
+      return _.concat(
+        _.orderBy(partition[0], field, otherOrder ? otherOrder : order),
```
Sybrand, line 56: You could hear my neurons stretching and snapping while I looked at this. Could you add some documentation on this? What is partition, why is concat happening? What what?

web/src/features/fbaCalculator/RowManager.ts
```diff
@@ -50,84 +50,99 @@ export class RowManager {
...
+      const partition = _.partition(tableRows, row => !!_.get(row, field, null))
+      return _.concat(
+        _.orderBy(partition[0], field, otherOrder ? otherOrder : order),
+        partition[1]
+      )
+    }
     switch (sortByColumn) {
       case SortByColumn.Zone: {
-        return _.orderBy(tableRows, 'zone_code', order)
+        return orderByWithEmptyAtBottom('zone_code')
```
andrea-williams, line 70: hmmm I recommend double-checking with @tessmgood on this - I know BB didn't like the behaviour that new rows automatically go to the bottom as soon as the "Add Row" button is clicked, because when you have a long list of stations, it means scrolling up to the top to click the add row button, then scrolling to the bottom of the page to add the station info, etc. And repeat.

I think the reason the ticket was created was because it was illogical that when sorting by HFI descending, rows with no HFI data were being listed before the rows with high HFIs.

tessmgood (approved):

Sybrand (approved):

web/src/features/fbaCalculator/RowManager.ts
tessmgood: Good point - I'm looking at BB's feedback as a separate ticket to isolate the interaction of adding a new row (it's different to sorting behaviour IMO) - I have thoughts & ideas about this that I will put in a ticket.

web/src/features/fbaCalculator/RowManager.ts
andrea-williams: My brain is still pretty broken by this - why is `!!` necessary?

conbrad: `!!` is a (unfortunate?) javascript idiom to coerce a value into a boolean based on it's truthiness: https://dev.to/sanchithasr/what-is-the-double-bang-operator-in-javascript-4i3h

andrea-williams: damn you Javascript

andrea-williams (approved):

PR #1285, opened 2021-09-02
Story/temp accuracy map/911

andrea-williams:
https://wps-pr-1285.apps.silver.devops.gov.bc.ca/morecast

Note: this doesn't work in Test env bc WFWX API isn't quick enough in responding to request for weather data - it times out.


tessmgood (approved):
V rad! Conditional approval: 
Small, weird thing — when I toggle between RH and temp using the select field, an active state persists after the switch which makes the 'view accuracy map for' label hard to read (blue on blue) unless I click away, then the label reverts to white. Your call: we can fix it later, I want to tighten up that whole footer area UI wise. Someone is very likely to click a station or the map which immediately fixes it.

andrea-williams:
> V rad! Conditional approval:
> Small, weird thing — when I toggle between RH and temp using the select field, an active state persists after the switch which makes the 'view accuracy map for' label hard to read (blue on blue) unless I click away, then the label reverts to white. Your call: we can fix it later, I want to tighten up that whole footer area UI wise. Someone is very likely to click a station or the map which immediately fixes it.

Should the label just always be white?

tessmgood:
> > V rad! Conditional approval:
> > Small, weird thing — when I toggle between RH and temp using the select field, an active state persists after the switch which makes the 'view accuracy map for' label hard to read (blue on blue) unless I click away, then the label reverts to white. Your call: we can fix it later, I want to tighten up that whole footer area UI wise. Someone is very likely to click a station or the map which immediately fixes it.
> 
> Should the label just always be white?

Yes :)

api/Dockerfile
```diff
@@ -1,7 +1,7 @@
 # NOTE:
 # This Dockerfile is for local development only!
 
-FROM python:3.8-buster
+FROM python:3.8.12-buster
```
conbrad, line 4: Why are we pinning a minor version?

api/app/wildfire_one/wfwx_api.py
```diff
@@ -128,7 +128,7 @@ async def get_detailed_stations(time_of_interest: datetime):
                     temperature=daily.get('temperature'),
                     relative_humidity=daily.get('relativeHumidity'))
                 record_type = daily.get('recordType').get('id')
-                if record_type == 'ACTUAL':
+                if record_type in ['ACTUAL', 'MANUAL']:
```
conbrad, line 131: What is manual?

web/package.json
```diff
@@ -60,7 +60,7 @@
     "start": "CI=true react-scripts start",
     "start:cypress": "export $(cat .env.cypress | xargs) && react-scripts -r @cypress/instrument-cra start",
     "build": "react-scripts build",
-    "test": "react-scripts test",
+    "test": "react-scripts test --transformIgnorePatterns \"node_modules/(?!ol)/\"",
```
conbrad, line 63: Why is this necessary?

web/src/features/fireWeather/components/AccuracyColorLegend.tsx
```diff
@@ -10,31 +20,30 @@ const useStyles = makeStyles({
...
-    marginLeft: '40px'
+    textAlign: 'center'
   },
   leftLabel: {
     height: '15px',
     width: '80px',
     fontSize: '10px',
     color: 'white',
-    marginLeft: '4px'
+    justifyContent: 'flex-start'
```
conbrad, line 32: Nice 😎

web/src/features/fireWeather/components/AccuracyColorLegend.tsx
```diff
@@ -44,52 +53,107 @@ const useStyles = makeStyles({
...
+    backgroundColor: darkOrangeColor,
+    borderTop: '2px solid #FFFFFF',
+    borderRight: '2px solid #FFFFFF',
+    borderBottom: '2px solid #FFFFFF',
+    borderRadius: '0px 2px 2px 0px',
+    height: 27,
+    width: 32
+  },
+  darkestRed: {
+    backgroundColor: '#720505',
```
conbrad, line 119: I like how the rest of the classes referenced named colors, can we do the same for the remaining ones?

web/src/features/fireWeather/components/AccuracyVariablePicker.tsx
```diff
@@ -0,0 +1,57 @@
...
+    color: 'white'
+  },
+  '& .MuiFormLabel-root.Mui-focused': {
+    color: 'white'
+  }
+})
+
+interface Props {
+  selectedWxVariable: AccuracyWeatherVariableEnum
+  changeHandler: any
```
conbrad, line 26: Lint is complaining about the any, I think this can be a function that accepts an event and returns void.

web/src/features/fireWeather/components/maps/WeatherMap.tsx
```diff
@@ -44,24 +40,51 @@ const source = new olSource.XYZ({
 
 const zoom = 6
 
+const rhPointStyleFunction = (feature: any) => {
```
conbrad, line 43: This any we don't know the type for, I think an eslint exception comment to suppress the warning would help.

web/src/features/fireWeather/components/maps/WeatherMap.tsx
```diff
@@ -44,24 +40,51 @@ const source = new olSource.XYZ({
...
+  return new Style({
+    image: new CircleStyle({
+      radius: computeRHAccuracySize(feature.values_),
+      fill: new Fill({ color: rhPointColor }),
+      stroke: computeStroke(rhPointColor)
+    })
+  })
+}
+
+const tempPointStyleFunction = (feature: any) => {
```
conbrad, line 54: Similar to above.

web/src/features/fireWeather/components/maps/WeatherMap.tsx
```diff
@@ -93,7 +116,7 @@ const WeatherMap = ({
     },
     [dispatch]
   )
-
+  console.log(stations)
```
conbrad: console.log snuck in

web/src/features/fireWeather/components/maps/stationAccuracy.test.ts
```diff
@@ -1,155 +1,509 @@
...
+
+    it('should return the largest scale index when observed RH is much lower than forecasted RH', () => {
+      expect(computeRHScaleIndex(100, 10)).toEqual(rhColorScale.length - 1)
+    })
+
+    it('should return a scale index of 0 when observed RH is much greater than forecasted RH', () => {
+      expect(computeRHScaleIndex(10, 100)).toEqual(0)
+    })
+
+    it('should return a scale index of neutralIndex+1 when observed RH is 4 percentage points lower than forecasted RH', () => {
```
conbrad, line 66: Minor: I thought it was 3 percentage points? Maybe this has changed or I'm remembering wrong.

conbrad (changes requested):

api/app/wildfire_one/wfwx_api.py
andrea-williams: Manual is where someone goes into WFWX API and manually overrides whatever values were automatically reported by the sensors at a weather station (I think this only ever happens when there's something wrong with a sensor). There were a lot of them recently in WFWX API and it was breaking things

api/Dockerfile
andrea-williams: Because poetry.lock specifies that the Python version must be `>=3.8.6`, but `3.8-buster` has a lower minor version, so the local Docker doesn't work unless we specify the only available minor version >= .6

web/package.json
andrea-williams: In WeatherMap we import styling components from OpenLayers, and Jest breaks unless we override `ol` node_modules because of problems with how/when the package is compiled

api/Dockerfile
conbrad: Ah this is for local docker. We'd lose higher minor updates by pinning this but for local dev it's low risk.

web/src/features/fireWeather/components/maps/stationAccuracy.test.ts
andrea-williams: Each step is 3 percentage points. So that means if observed RH is 3 percentage points lower than forecasted RH, the forecast is considered "correct" and the neutral colour will be shown. This test is checking for what happens when the forecast is considered "slightly wrong", as in the map marker should not be a neutral colour.

web/src/features/fireWeather/components/maps/WeatherMap.tsx
andrea-williams: Deleted :)

JulianForeman (approved):

conbrad (approved):

Sybrand (approved):

PR #1291, opened 2021-09-08
FireBAT: Sticky Columns

conbrad:
Test: https://wps-pr-1291.apps.silver.devops.gov.bc.ca/fire-behaviour-calculator?s=322&f=c1&c=undefined&w=undefined

- Adds `StickyCell` component to make chid components sticky
- Encapsulates checkbox into `SelectionCell`
- `SelectionCell`, `WeatherStationCell` and `FuelTypeCell` headers and table cells are children of `StickyCell`s so they remain on screen while scrolling


andrea-williams (approved):
LGTM

PR #1292, opened 2021-09-09
FireBAT: Loading Indicator

conbrad:
Test: https://wps-pr-1292.apps.silver.devops.gov.bc.ca/fire-behaviour-calculator?s=838&f=c1&c=undefined&w=20,s=322&f=c1&c=undefined&w=19

- Adds `LoadingIndicatorCell` that wraps cells and shows a loading indicator if that row is currently loading.
- Uses the `Skeleton` component from MaterialUI: https://material-ui.com/components/skeleton/
- Will attempt fixing progress bar in follow up PR

- [x] Given that one or many rows are loading, due to a new input or change, the loading row(s) have a loading state


andrea-williams (approved):
Nice

PR #1301, opened 2021-09-10
createMuiTheme -> createTheme

conbrad:
`createMuiTheme` is deprecated. Import and invoke `createTheme` instead.


Sybrand (approved):

andrea-williams (approved):
Sorry for the delay. Missed the email when it came in.

PR #1308, opened 2021-09-13
Added instructions for Firebat

JulianForeman:
https://wps-pr-1308.apps.silver.devops.gov.bc.ca/fire-behaviour-calculator

Instructions added for Firebat while the table is empty, and removes it once a row has been added.


web/src/features/fbaCalculator/components/FBATable.tsx
```diff
@@ -380,7 +401,10 @@ const FBATable = (props: FBAInputGridProps) => {
             variant="contained"
             color="primary"
             spinnercolor="white"
-            onClick={deleteSelectedStations}
+            onClick={() => {
+              deleteSelectedStations()
+              checkTable()
```
conbrad, line 406: An easier, more reactive way would be to conditionally render the table if rows are empty, or the table instructions otherwise: https://reactjs.org/docs/conditional-rendering.html#inline-if-else-with-conditional-operator

I can show you what I mean if you like.

JulianForeman: Ah yes, Sy mentioned there might be a better way to do this. I will read up and attempt it myself, if I hit a block I will reach out. Thanks Connor!

tessmgood (changes requested):
This is taking shape, nice!
When an [empty FireBAT page](https://psu.nrs.gov.bc.ca/fire-behaviour-calculator) loads now, the table header is displayed.
Screen Shot 2021-09-13 at 4 40 16 PM
 
I'd love to retain this, it's great context, and display the instructions in an 'empty table' container below that header. See [Figma](https://www.figma.com/file/AUhBnTB50dnY7ysoxBSqPN/FireBAT-2021?node-id=371%3A65) concept for a general direction - I'd recommend 48 top and bottom padding between copy and container.

web/cypress/integration/fire-behaviour-advisory-calculator-page.spec.ts
```diff
@@ -188,7 +183,8 @@ describe('FireBAT Calculator Page', () => {
 
       cy.getByTestId('remove-rows').click()
 
-      cy.rowCountShouldBe(0)
+      cy.getByTestId('fba-instructions')
```
conbrad, line 186:
```
cy.getByTestId('fba-instructions').should('be.visible')
```

web/cypress/integration/fire-behaviour-advisory-calculator-page.spec.ts
```diff
@@ -208,7 +204,8 @@ describe('FireBAT Calculator Page', () => {
...
```
conbrad, line 207:
```
cy.getByTestId('fba-instructions').should('be.visible')
```

web/src/features/fbaCalculator/components/FBATable.tsx
```diff
@@ -411,271 +412,282 @@ const FBATable = (props: FBAInputGridProps) => {
...
-                              }
-                              rowId={row.id}
-                            />
-                          </StickyCell>
-                          <LoadingIndicatorCell
-                            loading={loading}
-                            rowUpdating={rowIdsToUpdate.has(row.id)}
-                            initialLoad={initialLoad}
-                          >
+                {rows.length === 0 ? null : (
```
conbrad, line 415: Instead of returning `null` here you can return your new component. Then you won't need any more code below.

web/src/features/fbaCalculator/components/FBATable.tsx
```diff
@@ -411,271 +412,282 @@ const FBATable = (props: FBAInputGridProps) => {
...
+                        )
                       )
-                    )
-                  })}
-                </TableBody>
+                    })}
+                  </TableBody>
+                )}
               </Table>
+              {rows.length === 0 ? <FBATableInstructions /> : null}
```
conbrad, line 690: This can be captured in the existing ternary conditional above.

conbrad (changes requested):
Looking good, almost there.

conbrad (approved):

Sybrand (approved):
Works for me if it works for @tessmgood

tessmgood (changes requested):
Awesome! 
Last small thing and it's my fault: please left-align the copy. 

We get some responsive weirdness on screen sizes / browser widths that I didn't fully consider - some of the copy might get cut off.

tessmgood (approved):
RAD!

PR #1328, opened 2021-09-14
Deploy to production using GitHub Actions

Sybrand:
Using a workflow triggered by a deployment allows us to get rid of Jenkins, and aligns us better with introducing something like chatops.

Currently, to trigger deploy, you have to use the github api:

`gh api repos/bcgov/wps/deployments -f ref="[branch name]"`

### DIFF
--- a/.github/workflows/production.yml
+++ b/.github/workflows/production.yml
@@ -56,7 +56,7 @@ jobs:
           PROJ_TARGET=e1e498-prod CERTBOT_STAGING=false DRYRUN=false bash openshift/scripts/oc_provision_certbot_cronjob.sh prod apply
   set_deployment_status_success:
     name: Production deployment
-    needs: [deploy-prod]
+    needs: [deploy_to_production]
     runs-on: ubuntu-20.04
     steps:
       - name: Set status to deployed
--- a/.github/workflows/production.yml
+++ b/.github/workflows/production.yml
@@ -9,6 +9,7 @@ on:
 
 jobs:
   deploy_to_production:
+    name: Deploy to production
     runs-on: ubuntu-20.04
     steps:
       - name: Find PR number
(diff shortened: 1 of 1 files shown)


.github/workflows/production.yml
```diff
@@ -0,0 +1,72 @@
...
+        with:
+          # reference used to figure out this part:
+          # https://docs.github.com/en/rest/reference/pulls#list-pull-requests
+          # https://octokit.github.io/rest.js/v18#pulls
+          # https://github.com/actions/github-script
+          script: |
+            // get pull requests relating to this commit hash
+            const pulls = (await github.repos.listPullRequestsAssociatedWithCommit({owner: context.repo.owner, repo: context.repo.repo, commit_sha: context.sha }))
+            console.log(pulls)
+            // find an open PR (potentially problematic if you have multiple PR's open!)
```
conbrad, line 25: You can get this exact PR number from an environment variable, check out: https://github.com/bcgov/wps/blob/lnd/pr-github-action/.github/actions/pr-test-url/index.js

conbrad (changes requested):
Nice

.github/workflows/production.yml
Sybrand: It's not common to have multiple PR's at the same commit point. Thanks for the link - I'll test.

.github/workflows/production.yml
```diff
@@ -0,0 +1,67 @@
+name: Production
+
+# trigger this deployment by running:
+# gh api repos/bcgov/wps/deployments -f ref="[your branch]"
+on:
+  # reference used to figure this part out:
+  # https://docs.github.com/en/actions/reference/events-that-trigger-workflows#deployment
+  deployment:
```
andrea-williams, line 8: is having the `:` after `deployment` correct?

andrea-williams (approved):

conbrad (approved):
🚀

.github/workflows/production.yml
Sybrand: It doesn't break it - but sure - looks weird. I've removed it.

.github/workflows/production.yml
Sybrand: @conbrad so actually that doesn't work - loading that file is the same as referencing context - and it doesn't have the PR in this context.

conbrad: Oops, sorry about that.

PR #1336, opened 2021-09-16
Fixed Headers

JulianForeman:
Altered spacing on headers to align with content across all apps.

https://wps-pr-1336.apps.silver.devops.gov.bc.ca/percentile-calculator


web/src/components/GeneralHeader.tsx
```diff
@@ -0,0 +1,54 @@
...
+  spacing: number
+}
+
+export const GeneralHeader: React.FunctionComponent<Props> = (props: Props) => {
+  const { title, productName, spacing, padding } = props
+  const classes = useStyles(props)
+
+  return (
+    <nav className={classes.root}>
+      <OptionalContainer style={{ paddingLeft: padding }} className={classes.container}>
```
conbrad, line 44: Can we add this `paddingLeft` to the `container` class? It would make CSS harder to reason about with both inline styles and classes.

conbrad (approved):
Nice!!!

tessmgood (changes requested):
Loving this, we're so close!
Our two H2 page titles/product names for HFI Calculator and C-Haines should also be aligned to the left, with the logo and page body
Screen Shot 2021-09-16 at 12 46 59 PM
Screen Shot 2021-09-16 at 12 46 16 PM

Sybrand (approved):
Address @conbrad 's comment and I'm happy.

web/src/components/GeneralHeader.tsx
JulianForeman: Is there a way to do this while keeping it passed in through props?

conbrad: Yep, check out: https://github.com/bcgov/wps/blob/e911111ec6000d49297d85ab8334762d7c2480a8/web/src/features/fireWeather/components/ExpandableContainer.tsx#L18

tessmgood (approved):
Awesome, thanks J!

PR #1338, opened 2021-09-16
Bug Fix: Jumping Morecast Map

conbrad:
Test: https://wps-pr-1338.apps.silver.devops.gov.bc.ca/morecast

- Pushes `VectorLayer` child creation down to the `Map` component so `WeatherMap` (parent) doesn't re-render it with the BC center point, every time station selection changes 
- Pulls out VectorLayer-only logic into `FireIndicesVectorLayer` to simplify the `Map` component


tessmgood (approved):
Big UX win. Thanks!

JulianForeman (approved):

andrea-williams (approved):
Niiiiiiiiiiice

PR #1339, opened 2021-09-16
Bug Fix: Round upward for Mean Intensity Group

conbrad:
`ceil`


web/src/features/hfiCalculator/components/meanIntensity.ts
```diff
@@ -20,10 +20,12 @@ export const calculateMeanIntensityGroup = (
   }
   return stationIntensityGroups.length === 0
     ? undefined
-    : Math.round(
-        (10 * stationIntensityGroups.reduce((a, b) => a + b, 0)) /
-          stationIntensityGroups.length
-      ) / 10
+    : Math.ceil(
+        Math.round(
```
andrea-williams, line 24: I don't think it's necessary to do both Math.round() and Math.ceil() - shouldn't Math.ceil() alone produce the desired output?

conbrad: `Math.round` is around the intermediate operation, it's then divided by 10. `Math.ceil` is around the last operation. It would change the behaviour but seems reasonable.

conbrad: Removed `round`.

andrea-williams (approved):

andrea-williams (changes requested):
Cypress tests failing.

JulianForeman (approved):

andrea-williams (approved):

conbrad:
Closed because we already have the desired behaviour in prod.

PR #1371, opened 2021-09-23
HFI Weekly Table View

andrea-williams:


tessmgood:
looking rad @conbrad 
this may be more than you wanted, but gotta capture it when I see it: 
- make date input field wider, so 'date of interest' label isn't wrapping (I believe we did the same on FireBAT)
- use the BC blue for the toggle please (Figma)
- now that we have the toggle, we can drop the copy/label at the top left of the table "HFI Calculator Daily/Weekly View"
- in daily and weekly view, the planning area row is two different colours - what's going on there? (see Figma)
- can we align all table header labels to the bottom (scannability)
- the width of the columns in weekly view is more condensed than daily view - can't figure out why that's happening, but weekly view is too squished together (scannability) - screen share chat?

PR #1398, opened 2021-10-05
FireBAT - Show/hide columns in table

andrea-williams:


tessmgood:
Looking great!
- Can we reduce space between list items/checkboxes by about half?
- What do you think about making 'apply' and 'save' always visible? Create a persistent component at the bottom of the modal, and only the list of checkboxes is scrollable. The challenge: the actions aren't visible and obvious unless you want to exclude the latter columns of the table (arguably the least likely to be excluded) 
- copy change: from Filter Table by Columns... to Show Columns

andrea-williams:
Future Work (maybe (probably)):

It doesn't make sense to allow users to hide the Weather Station or Fuel Type columns - the table is basically useless without them. But keeping these columns fixed while maintaining their current L->R order in the table is complicated and beyond the scope of this ticket. So for now, it's a "feature" that users have the freedom to hide columns that provide crucial context as to what they're looking at.

Also noteworthy is that because the current implementation doesn't maintain the L->R order that columns appear in the table, if a user deselects (hides) a column and then later re-selects it (makes it visible again), that column gets appended to the far right of the table, thereby messing up the order. This ~~bug~~ feature could be fixed as a consequence of implementing a set order for the columns.

tessmgood:
> Future Work (maybe (probably)):
> 
> It doesn't make sense to allow users to hide the Weather Station or Fuel Type columns - the table is basically useless without them. But keeping these columns fixed while maintaining their current L->R order in the table is complicated and beyond the scope of this ticket. So for now, it's a "feature" that users have the freedom to hide columns that provide crucial context as to what they're looking at.
> 
> Also noteworthy is that because the current implementation doesn't maintain the L->R order that columns appear in the table, if a user deselects (hides) a column and then later re-selects it (makes it visible again), that column gets appended to the far right of the table, thereby messing up the order. This ~bug~ feature could be fixed as a consequence of implementing a set order for the columns.

FireBAT backlog #1406

PR #1399, opened 2021-10-06
Story/weekly table view/1022

conbrad:


web/src/features/hfiCalculator/components/DailyViewTable.tsx
```diff
@@ -178,11 +97,6 @@ export const DailyViewTable = (props: Props): JSX.Element => {
 
   return (
     <FireContainer testId={props.testId}>
-      <div className={classes.controls}>
-        <Typography component="div" variant="subtitle2">
-          {props.title}
-        </Typography>
-      </div>
       <FireTable
         maxHeight={700}
```
Sybrand, line 101: Why are we limiting to 700 pixels? I have a huge screen, and everything is squished to the top. (I guess this is an existing limitation?) I've added a ticket for it.

web/src/features/hfiCalculator/components/DayHeaders.test.tsx
```diff
@@ -0,0 +1,56 @@
+import { Table, TableBody, TableContainer, TableRow } from '@material-ui/core'
+import { render } from '@testing-library/react'
+import React from 'react'
+import DayHeaders from 'features/hfiCalculator/components/DayHeaders'
+describe('DayHeaders', () => {
+  it('should return table row with the headers for Monday - Friday given the ISO Date', () => {
+    const prepCycle = [
+      'Mon., Oct. 04',
+      'Tue., Oct. 05',
+      'Wed., Oct.06',
```
Sybrand, line 10: Is this working? There's no space between "Oct." and "06" - why is this passing?

web/src/features/hfiCalculator/components/DayHeaders.test.tsx
```diff
@@ -0,0 +1,56 @@
...
+        <Table>
+          <TableBody>
+            <TableRow>
+              <DayHeaders isoDate={'2021-10-05'}></DayHeaders>
+            </TableRow>
+          </TableBody>
+        </Table>
+      </TableContainer>
+    )
+    for (let i = 0; i < 5; i++) {
```
Sybrand, line 25: Do a foreach rather than hard coding 5 iterations.

web/src/features/hfiCalculator/components/DayHeaders.test.tsx
```diff
@@ -0,0 +1,56 @@
...
+        <Table>
+          <TableBody>
+            <TableRow>
+              <DayHeaders testId={'monday-test'} isoDate={'2021-10-08'}></DayHeaders>
+            </TableRow>
+          </TableBody>
+        </Table>
+      </TableContainer>
+    )
+    for (let i = 0; i < 5; i++) {
```
Sybrand, line 50: Do a foreach rather than doing 5 iterations.

web/src/features/hfiCalculator/components/DayHeaders.test.tsx
```diff
@@ -0,0 +1,56 @@
...
+      const cell = getByTestId(i)
+      expect(cell.className).toMatch(/makeStyles-dayHeader-/)
+      expect(cell.innerHTML.match(prepCycle[i]))
+    }
+  })
+  it('should return table row with the headers for Thursday - Monday given the ISO Date', () => {
+    const prepCycle = [
+      'Thu.,  Oct. 07',
```
Sybrand, line 33: Why is this passing when there's an extra space between "Thu.," and "Oct."

web/src/features/hfiCalculator/components/DayHeaders.test.tsx
```diff
@@ -0,0 +1,56 @@
+import { Table, TableBody, TableContainer, TableRow } from '@material-ui/core'
+import { render } from '@testing-library/react'
+import React from 'react'
+import DayHeaders from 'features/hfiCalculator/components/DayHeaders'
+describe('DayHeaders', () => {
+  it('should return table row with the headers for Monday - Friday given the ISO Date', () => {
+    const prepCycle = [
+      'Mon., Oct. 04',
```
Sybrand, line 8: Will this fail when run with different locale settings? If your local machine is french, what will this work?

web/src/features/hfiCalculator/components/DayHeaders.tsx
```diff
@@ -0,0 +1,44 @@
...
+    borderLeft: '1px solid #C4C4C4',
+    borderBottom: 'none',
+    textAlign: 'center'
+  },
+  spaceHeader: {
+    border: 'none'
+  }
+})
+const DayHeaders = (props: DayHeadersProps) => {
+  const { start } = getPrepWeeklyDateRange(props.isoDate + 'T00:00:00-07:00')
```
Sybrand, line 26: The 'T00:00:00-07:00' was a hack I added just to move us forward, I don't think this is a good solution. At the very least this should probably live in a function somewhere - but ideally we need to have a better way to set the date.

web/src/features/hfiCalculator/components/CellHeaders.tsx
```diff
@@ -0,0 +1,51 @@
...
+const useStyles = makeStyles({
+  grassCureBorder: {
+    borderLeft: '1px solid #C4C4C4'
+  }
+})
+const CellHeaders = () => {
+  const classes = useStyles()
+  return (
+    <React.Fragment>
+      {range(5).map(i => (
```
Sybrand, line 15: Make use of the NUM_WEEK_DAYS constant here.

web/src/features/hfiCalculator/components/MeanIntensityGroupRollup.tsx
```diff
@@ -47,6 +47,9 @@ const useStyles = makeStyles({
   alignErrorIcon: {
     marginTop: '6px',
     textAlign: 'center'
+  },
+  default: {
+    background: 'rgba(40, 53, 147, 0.05)'
```
Sybrand, line 52: I have an urge to want to start creating a colour palette file somewhere.... just a thought.

web/src/features/hfiCalculator/components/PrepLevelCell.tsx
```diff
@@ -0,0 +1,107 @@
...
+    color: 'white'
+  },
+  default: {
+    backgroundColor: 'rgba(40, 53, 147, 0.05)'
+  }
+})
+
+export type PrepLevel = 1 | 2 | 3 | 4 | 5 | 6 | undefined
+
+export const calculatePrepLevel = (meanIntensityGroup: number | undefined): PrepLevel => {
```
Sybrand, line 60: I've made a ticket to move this kind of logic to the back-end. I feel like this is all stuff that should be part of the json response coming back from the API.

web/src/utils/date.ts
```diff
@@ -48,3 +48,50 @@ export const formatDateInUTC00Suffix = (dtISO: string): string => {
...
   return isoNoon.substring(0, isoNoon.length - 1) + '+00:00'
 }
+
+export const getPrepWeeklyDateRange = (
+  dateOfInterest: string
+): { start: DateTime; end: DateTime } => {
+  const day = DateTime.fromISO(dateOfInterest).weekdayLong
+  let dayOffset = 0
+  switch (day) {
+    case 'Tuesday':
```
Sybrand, line 58: This will fail of your computers locale happens to be set to a different language.
Rather use the javascript .getDay() function available on the Date object. That will give you the day of the week.
e.g. for Wednesday, getDay() will return 3. So for day.getDay() => 2 day.getDay() <=3, a.getDate() - 1 etc. etc.

Sybrand (changes requested):
I've asked for quite a few changes - nothing major I don't think - except maybe for the locale stuff in the unit tests.

Ux testing: For Sep 27, There's a grass cure warning on on the Afton in the daily view, but not on the weekly view. It puts the grass cure warning in the Kamloops rolled up values in the heading, but shows grass cure as 0.0 instead of the warning icon on the Afton line.

web/src/features/hfiCalculator/components/PrepLevelCell.tsx
Sybrand: It wouldn't be another network call - I'd rather envision everything being returned at once.

tessmgood:
> I've asked for quite a few changes - nothing major I don't think - except maybe for the locale stuff in the unit tests.
> 
> Ux testing: For Sep 27, There's a grass cure warning on on the Afton in the daily view, but not on the weekly view. It puts the grass cure warning in the Kamloops rolled up values in the heading, but shows grass cure as 0.0 instead of the warning icon on the Afton line.

Communicating errors in weekly view problem is broader, CB and I chatted through it a bit yesterday: #1405 in Sprint N

Sybrand:
> > I've asked for quite a few changes - nothing major I don't think - except maybe for the locale stuff in the unit tests.
> > Ux testing: For Sep 27, There's a grass cure warning on on the Afton in the daily view, but not on the weekly view. It puts the grass cure warning in the Kamloops rolled up values in the heading, but shows grass cure as 0.0 instead of the warning icon on the Afton line.
> 
> Communicating errors in weekly view problem is broader, CB and I chatted through it a bit yesterday: #1405 in Sprint N

@tessmgood but we still shouldn't be showing 0.0 for the grass cure in the weekly view if it's not defined right? If we're not going to show an error thing there - then at least we should show nothing, instead of making up 0.0?

web/src/features/hfiCalculator/components/PrepLevelCell.test.tsx
```diff
@@ -0,0 +1,86 @@
...
+            </TableRow>
+          </TableBody>
+        </Table>
+      </TableContainer>
+    )
+    const cell = getByTestId('weekly-prep-level-afton')
+    expect(cell.className).toMatch(/makeStyles-prepLevel4-/)
+    expect(getByText('4')).toBeDefined
+  })
+})
```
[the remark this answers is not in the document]
andrea-williams: (it's also complaining about not testing for prep levels 5 and 6, which I don't think is as important, but I do think a test for the default would be valuable

web/src/features/hfiCalculator/components/DayHeaders.test.tsx
JulianForeman: Fixed!

web/src/features/hfiCalculator/components/PrepLevelCell.test.tsx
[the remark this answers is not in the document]
JulianForeman: When I looked at the meanIntensity calculation file there was no logic for prep levels 5 and 6

web/src/features/hfiCalculator/components/DailyViewTable.tsx
conbrad: It's because the headers are not sticky if it takes up the full screen.

tessmgood:
> There's a few mostly nitpicky comments I've added,
> 
> * I'd recommend adding some whitespace at the very bottom of the page. It's esp noticeable in Daily View that the table ends right at the bottom of the page - in our other apps (such as Firebat) we've added a small amount of whitespace.
> * I'm not a big fan of how the weekly table looks when only some days have data, but I'll leave that up to TG's discretion whether it gets addressed now or later.

valid, not nitpicky. 
- agreed. there is whitespace at some resolutions/screen sizes, but the table isn't properly responsive. This was captured in #1402 but we aren't in agreement on optimal behaviour / solution. I'll dedicate our next Human Experience meeting to this topic unless we get to it sooner. 
- agreed again. #1405 in progress now to figure this out

PR #1442, opened 2021-10-25
HFI - copy to clipboard - 1425

andrea-williams:
- Fixed a bug (that's been sitting in prod for a while) - daily table was displaying FFMC in FWI column
- Weekly and Daily View Tables on webpage now sorted by zone code instead of zone name (based on user feedback in testing sessions)
- Added threshold of 2% for Codecov patch (Sybrand told me to do it because codecov on the patch was 85.5% before adding any unit tests, and then went down to 84.x% AFTER adding unit tests. So that's just terrific)

-------
# Test Links:
[HFI Calculator](https://wps-pr-1442.apps.silver.devops.gov.bc.ca/hfi-calculator)


tessmgood (changes requested):
Nice. I think we need an info (i) icon here. 
Right now the intent and step following "copy to clipboard" isn't obvious. They'll figure it out, but given there is a step required in excel as well (Data > Text to columns > Comma separated...) there's some friction we can ease.  
- (i) tooltip: You can paste all table data in Excel. To format: go to the Data tab, using Text to Columns > Delimited > Comma.
- rough first pass, I'll test it, feel free to build on it
- place (i) to the right of the Copy button
- there's no way to automagically do that Excel formatting, right?

andrea-williams:
> Nice. I think we need an info (i) icon here. Right now the intent and step following "copy to clipboard" isn't obvious. They'll figure it out, but given there is a step required in excel as well (Data > Text to columns > Comma separated...) there's some friction we can ease.
> 
> * (i) tooltip: You can paste all table data in Excel. To format: go to the Data tab, using Text to Columns > Delimited > Comma.
> * rough first pass, I'll test it, feel free to build on it
> * place (i) to the right of the Copy button
> * there's no way to automagically do that Excel formatting, right?

It is already automagically doing the Excel formatting. After clicking copy to clipboard, you just have to paste into Excel and ta-daaaaaa it's already formatted

tessmgood:
> > Nice. I think we need an info (i) icon here. Right now the intent and step following "copy to clipboard" isn't obvious. They'll figure it out, but given there is a step required in excel as well (Data > Text to columns > Comma separated...) there's some friction we can ease.
> > 
> > * (i) tooltip: You can paste all table data in Excel. To format: go to the Data tab, using Text to Columns > Delimited > Comma.
> > * rough first pass, I'll test it, feel free to build on it
> > * place (i) to the right of the Copy button
> > * there's no way to automagically do that Excel formatting, right?
> 
> It is already automagically doing the Excel formatting. After clicking copy to clipboard, you just have to paste into Excel and ta-daaaaaa it's already formatted

that didn't happen for me :/ do I have to paste a certain way?

web/package.json
```diff
@@ -29,8 +29,8 @@
     "@types/react-redux": "^7.1.16",
     "@types/react-router-dom": "^5.3.0",
     "@types/webpack-env": "^1.15.1",
-    "@typescript-eslint/parser": "^5.0.0",
     "@typescript-eslint/eslint-plugin": "^5.0.0",
+    "@typescript-eslint/parser": "^5.0.0",
```
conbrad, line 33: Why this change?

web/src/features/hfiCalculator/pages/HfiCalculatorPage.tsx
```diff
@@ -57,6 +64,11 @@ const HfiCalculatorPage: React.FunctionComponent = () => {
   const [previouslySelectedDateOfInterest, setPreviouslySelectedDateOfInterest] =
     useState(DateTime.now().setZone('UTC-7').toISO())
 
+  const weeklyViewAsString = RowManager.exportWeeklyRowsAsStrings(fireCentres, dailies)
```
conbrad, line 67: Should we only generate these strings when the copy button is clicked?

web/src/features/hfiCalculator/RowManager.ts
```diff
@@ -0,0 +1,249 @@
...
+              )
+              rowArray.push(
+                !isUndefined(day) && !grassCureError
+                  ? day.intensity_group.toString()
+                  : 'ND'
+              )
+              rowArray.push(Array(2).join(','))
+            })
+
+            rowArray.push(Array(2).join(','))
```
conbrad, line 240: Similiar to above, 2?

web/src/features/hfiCalculator/RowManager.ts
```diff
@@ -0,0 +1,249 @@
...
+    const rowsAsStrings: string[] = []
+    const dateSet = new Set()
+
+    dailies.flatMap(dailyRecord => {
+      dateSet.add(
+        `${dailyRecord.date.weekdayShort} ${dailyRecord.date.monthShort} ${dailyRecord.date.day}`
+      )
+    })
+    // build header row of dates
+    rowsAsStrings.push(Array(5).join(',').concat(Array.from(dateSet).join(',,,,,')))
```
conbrad, line 192: Can we pull out this constant, I'm assuming it's the number of header rows?

web/src/features/hfiCalculator/RowManager.ts
```diff
@@ -0,0 +1,249 @@
...
+  calculateMaxMeanIntensityGroup,
+  calculateMeanIntensity,
+  calculateMeanIntensityGroupLevel
+} from 'features/hfiCalculator/components/meanIntensity'
+import { calculatePrepLevel } from 'features/hfiCalculator/components/prepLevel'
+import { isValidGrassCure } from 'features/hfiCalculator/validation'
+import { NUM_WEEK_DAYS } from 'features/hfiCalculator/constants'
+
+// the number of decimal places to round to
+const DECIMAL_PLACES = 1
```
conbrad, line 18: Already exists in https://github.com/bcgov/wps/blob/5ac0ad8e9ca1e8d777acdd5fe8f98ae2468c29a8/web/src/features/hfiCalculator/constants.ts#L1

conbrad (changes requested):

web/src/features/hfiCalculator/RowManager.ts
```diff
@@ -0,0 +1,249 @@
...
+const printGrassCurePercentage = (daily: StationDaily): string => {
+  if (!isUndefined(daily) && !isNull(daily.grass_cure_percentage)) {
+    return daily.grass_cure_percentage.toString()
+  } else {
+    return 'ND'
+  }
+}
+
+export class RowManager {
+  public static exportDailyRowsAsStrings = (
```
Sybrand, line 29: I'm concerned about the disconnect; I was thinking you'd have something along these lines:
```
rows = makeTheRows()
// so that:
gimmeCsv(rows)
// and so that
<MyTable rows=rows/>
```
Why is there such a big different between the csv, and what's being displayed in the html?

Sybrand (commented):
@conbrad did a good review, I'm only interested in discussing the html vs. csv differences.

web/src/features/hfiCalculator/pages/HfiCalculatorPage.tsx
andrea-williams: Reworked - string is only generated when the copy button is clicked

web/src/features/hfiCalculator/RowManager.ts
```diff
@@ -0,0 +1,249 @@
...
+                !isUndefined(day) && !grassCureError
+                  ? day.hfi.toFixed(DECIMAL_PLACES)
+                  : 'ND'
+              )
+              rowArray.push(
+                !isUndefined(day) && !grassCureError
+                  ? day.intensity_group.toString()
+                  : 'ND'
+              )
+              rowArray.push(Array(2).join(','))
```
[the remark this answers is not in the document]
andrea-williams: it's just creating an array of 2 commas. It's needed to create empty cells. I'll add a comment in the code explaining this

web/src/features/hfiCalculator/RowManager.ts
andrea-williams: removed this declaration, importing from constants.ts instead

web/package.json
andrea-williams: I think at some point I ran a yarn update and this happened. I don't remember what I was doing at the time. I'll revert this.

andrea-williams: oh it might have been a dependency that was installed when I initially installed the `copy-to-clipboard` package

web/src/features/hfiCalculator/components/WeeklyViewTable.tsx
```diff
@@ -109,7 +119,7 @@ export const WeeklyViewTable = (props: Props): JSX.Element => {
                 </TableCell>
               </TableRow>
               {Object.entries(centre.planning_areas)
-                .sort((a, b) => (a[1].name < b[1].name ? -1 : 1))
+                .sort((a, b) => (a[1].name.slice(-3) < b[1].name.slice(-3) ? -1 : 1)) // sort by zone code
```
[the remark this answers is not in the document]
andrea-williams: it's possible, yes, but I don't think it wouldn't be a trivial refactor because the `PlanningArea` object is used in a bunch of places in our codebase

web/src/features/hfiCalculator/RowManager.ts
andrea-williams: Remember near the start of this ticket (a very long time ago, I know) when I had a dev chat with you & Conor about this - initially I tried to just export the HTML table to a string but it wasn't working with Excel at all, so we all agreed that the most efficient approach would be to just manually construct the CSV string

tessmgood (changes requested):
One change, my fault: bad designer!
Can we change the 'about this data' icon to a question mark (?) please. Material-UI Help-Outline icon. I don't want us to get too loosey goosey with the (i). It should always indicate there is a tooltip with more info.

web/src/features/hfiCalculator/RowManager.ts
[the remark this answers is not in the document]
conbrad: Can we make it a named constant like `DECIMAL_PLACES`?

web/src/features/hfiCalculator/components/WeeklyViewTable.tsx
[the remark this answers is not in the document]
conbrad: Maybe for now just make a small helper function called `getZoneFromAreaName` that does this so it's self documenting and reusable.

web/package.json
```diff
@@ -29,7 +29,6 @@
     "@types/react-redux": "^7.1.16",
     "@types/react-router-dom": "^5.3.0",
     "@types/webpack-env": "^1.15.1",
-    "@typescript-eslint/parser": "^5.0.0",
```
conbrad, line 32: Why this change?

codecov.yml
```diff
@@ -3,6 +3,9 @@
 coverage:
   # https://docs.codecov.com/docs/commit-status
   status:
+    patch:  
+      default:      
+        threshold: 2%
```
conbrad, line 8: Why this change?

conbrad (changes requested):
Can we pull out the `FormControl` block that contains the copy button into it's own component along with the `isCopied` state and `copyTable` method?

It would reduce the scope of the HFICalculatorPage and make the copy functionality more easily testable and reusable.

codecov.yml
Sybrand: Chasing a 0.01% patch change seemed like a waste of time. I'm not sure what we want as a threshold, but essentially this was my idea.

web/src/features/hfiCalculator/FormatTableAsCSV.ts
```diff
@@ -0,0 +1,252 @@
...
+                : 'ND'
+            )
+            rowArray.push(!isUndefined(daily) && !grassCureError ? daily.fire_type : 'ND')
+            rowArray.push(
+              !isUndefined(daily) && !grassCureError
+                ? daily.intensity_group.toString()
+                : 'ND'
+            )
+
+            rowsAsStrings.push(rowArray.toString())
```
Sybrand, line 132: Uh oh - this isn't going to work - you can't just turn it into a string, for csv you have to escape things correctly.
e.g. - look at the output of:
`
[1, 2, 3, '4,5'].toString()
`
it returns the string: '1,2,3,4,5' - which is not what you want. The correct csv would be something like:
1,2,3,'4,5'

web/package.json
andrea-williams: because https://github.com/bcgov/wps/pull/1442#discussion_r742220103

web/src/features/hfiCalculator/formatTableAsCSV.test.ts
```diff
@@ -0,0 +1,691 @@
+import { FormatTableAsCSV } from 'features/hfiCalculator/FormatTableAsCSV'
+import { DateTime } from 'luxon'
+import { FireCentre } from 'api/hfiCalcAPI'
+import { StationDaily } from 'api/hfiCalculatorAPI'
+describe('RowManager', () => {
+  const fireCentres: Record<string, FireCentre> = {
+    0: {
+      id: 12,
+      name: 'Kamloops Fire Centre',
```
Sybrand, line 9: Please add a name that contains some csv characters that need to be escaped. For example, 'Kamloops, Fire Centre' or 'Kamloops \' fire " Centre' to make sure escaping is working.

Sybrand (changes requested):
CSV - it's not as easy as just turning an array into a string.

web/package.json
conbrad: Please fix with `git checkout origin/main -- path/to/package.json`

web/src/features/hfiCalculator/FormatTableAsCSV.ts
andrea-williams: It's not perfect but I don't see how this would ever actually break in production - HFI calc doesn't accept any user-provided values, and the values coming from the API are strictly numbers, so there won't be a `'` or anything characters like that in the values ever.

andrea-williams:
> Can we pull out the `FormControl` block that contains the copy button into it's own component along with the `isCopied` state and `copyTable` method?
> 
> It would reduce the scope of the HFICalculatorPage and make the copy functionality more easily testable and reusable.

Are you sure this is really valuable? AFAIK HFI Calc is the only app that needs copy to clipboard functionality, in which case I wouldn't want to spend the effort make something reusable for nothing else.

conbrad:
> > Can we pull out the `FormControl` block that contains the copy button into it's own component along with the `isCopied` state and `copyTable` method?
> > It would reduce the scope of the HFICalculatorPage and make the copy functionality more easily testable and reusable.
> 
> Are you sure this is really valuable? AFAIK HFI Calc is the only app that needs copy to clipboard functionality, in which case I wouldn't want to spend the effort make something reusable for nothing else.

Ok let's defer.

web/src/features/hfiCalculator/FormatTableAsCSV.ts
conbrad: I agree with @andrea-williams here. For our simple use case this works. If we start accepting user input / more complex CSV use cases then we should pull in a well-tested library that handles all the edge cases.

web/src/features/hfiCalculator/RowManager.ts
Sybrand: Oh yeah - I'm not saying HTML -> csv ; I'm saying 0) logic -> data 1) data -> html 2) data -> csv ; right now the pattern seems to be 0) logic -> blah 1) blah -> data -> html 2) blah -> other data -> csv) ;
It's a concern - I'm ok not addressing it right now, just feel it's going to come back an bite us.

web/src/features/hfiCalculator/FormatTableAsCSV.ts
Sybrand: I'm going to be obstinate here - I've seen this fail so many time. We have no control over what's going into the name field - it's very much outside of our control, and IS a user provided value. Someone is going around giving stations names.

It's really trivial if we use something like this: https://www.npmjs.com/package/csv-string

conbrad: True, I didn't think about station names.

web/src/features/hfiCalculator/FormatTableAsCSV.ts
```diff
@@ -0,0 +1,253 @@
...
+    dailies: StationDaily[]
+  ): string => {
+    const rowsAsStrings: string[] = []
+
+    rowsAsStrings.push(dailyTableColumnLabels.toString())
+
+    Object.entries(fireCentres).forEach(([, centre]) => {
+      rowsAsStrings.push(centre.name)
+      Object.entries(centre.planning_areas)
+        .sort((a, b) => (a[1].name.slice(-3) < b[1].name.slice(-3) ? -1 : 1)) // sort by zone code
```
conbrad, line 43: Can we replace the comment with a function similar to `getZoneFromAreaName`

web/src/features/hfiCalculator/FormatTableAsCSV.ts
```diff
@@ -0,0 +1,253 @@
...
+    // build header row of dates
+    rowsAsStrings.push(
+      Array(NUM_STATION_DATA_COLS).join(',').concat(Array.from(dateSet).join(',,,,,'))
+    )
+    rowsAsStrings.push(weeklyTableColumnLabels.toString())
+
+    Object.entries(fireCentres).forEach(([, centre]) => {
+      rowsAsStrings.push(centre.name)
+      Object.entries(centre.planning_areas)
+        .sort((a, b) => (a[1].name.slice(-3) < b[1].name.slice(-3) ? -1 : 1)) // sort by zone code
```
conbrad, line 202: Similar to above

web/src/features/hfiCalculator/components/meanIntensity.ts
```diff
@@ -40,8 +40,10 @@ export const calculateMaxMeanIntensityGroup = (
...
 ): number => Math.max(...dailyMeanIntensities.filter(isDefined))
 
-export const calculateMeanPrepLevel = (
+export const calculateMeanIntensityGroupLevel = (
   dailyMeanIntensities: (number | undefined)[]
-): number =>
-  dailyMeanIntensities.filter(isDefined).reduce((prev, curr) => prev + curr) /
-  dailyMeanIntensities.length
+): number | undefined =>
+  dailyMeanIntensities.filter(isDefined).length > 0
```
conbrad, line 46: Nice can we add a small regression test for this bug? It can be added to `web/src/features/hfiCalculator/components/meanIntensity.test.ts`

web/src/features/hfiCalculator/FormatTableAsCSV.ts
```diff
@@ -0,0 +1,253 @@
...
+            })
+
+            rowArray.push(Array(NUM_WEEKLY_SUMMARY_CELLS).join(','))
+
+            rowsAsStrings.push(CSV.stringify(rowArray))
+          })
+        })
+    })
+
+    return rowsAsStrings.join('\r')
```
conbrad, line 251: Just curious, what does this character do and is it platform agnostic?

web/src/features/hfiCalculator/FormatTableAsCSV.ts
```diff
@@ -0,0 +1,253 @@
...
+              !isUndefined(daily) && !grassCureError
+                ? daily.intensity_group.toString()
+                : 'ND'
+            )
+
+            rowsAsStrings.push(CSV.stringify(rowArray))
+          })
+        })
+    })
+    return rowsAsStrings.join('\r\n')
```
conbrad, line 137: Does this behave different on windows vs mac? Should it be this?
https://nodejs.org/api/os.html#oseol

conbrad (approved):

Sybrand (changes requested):
Pls. get rid of that silly duplicate line break

JulianForeman (approved):

web/src/features/hfiCalculator/FormatTableAsCSV.ts
andrea-williams: `\r` is carriage return. Turns out that it's no longer necessary with the csv-string Typescript module we're using

web/src/features/hfiCalculator/FormatTableAsCSV.ts
andrea-williams: Deleted `\r\n` because it's no longer needed with the string-to-CSV module we're using now

web/src/features/hfiCalculator/formatTableAsCSV.test.ts
```diff
@@ -0,0 +1,793 @@
...
+    
+    FRENCH BAR (306),1320,C7,ND,ND,ND,ND,ND,ND,ND,ND,ND,ND,ND,ND,ND,ND,ND,ND,ND,ND,ND
+    
+    GWYNETH LAKE (309),1205,C7,ND,ND,ND,ND,ND,ND,ND,ND,ND,ND,ND,ND,ND,ND,ND,ND,ND,ND,ND
+    
+    LILLOOET (280),408,C7,ND,ND,ND,ND,ND,ND,ND,ND,ND,ND,ND,ND,ND,ND,ND,ND,ND,ND,ND
+    
+    SPLINTLUM (1055),424,C7,ACTUAL,4.8,57.0,223,4.8,0.0,NaN,47.664,5.62,460.989,0.154,10.908,0.098,NaN,0.0,0.0,0.0,SUR,1`
+
+    expect(dailyTableCSVString.replace(/\s+/g, '')).toEqual(
```
Sybrand, line 789: Why do we have the white-space removal?

web/src/features/hfiCalculator/formatTableAsCSV.test.ts
```diff
@@ -0,0 +1,793 @@
...
+
+    const dailyTableCSVString = FormatTableAsCSV.exportDailyRowsAsStrings(
+      missingElevationFireCentres,
+      dailies
+    )
+    const expectedDailyNDString = `Location,Elev. (m),FBP Fuel Type,Status,Temp (°C),RH (%),Wind Dir (°),Wind Speed (km/h),Precip (mm),Grass Cure (%),FFMC,DMC,DC,ISI,BUI,FWI,DGR CL,ROS (m/min),HFI,60 min fire size (ha),Fire Type,M/FIG,Fire Starts,Prep Level
+    Kamloops Fire Centre
+    Kamloops (K2), ,,,,,,,,,,,,,,,,,,,, 1, 0-1, 1
+    AFTON (322),ND,O1B,ND,ND,ND,ND,ND,ND,ERROR,ND,ND,ND,ND,ND,ND,ND,ND,ND,ND,ND,ND
+
```
Sybrand, line 746: Why do we have the additional line break here? Isn't that now redundant? Doesn't it also make the white-space removal below redundant?

web/src/features/hfiCalculator/formatTableAsCSV.test.ts
```diff
@@ -0,0 +1,793 @@
...
+    escapeCharFireCentres[0].planning_areas[2].stations[0].station_props.name = 'FINTRY,'
+    const dailyTableCSVString = FormatTableAsCSV.exportDailyRowsAsStrings(
+      escapeCharFireCentres,
+      dailies
+    )
+    const expectedDailyEscapeString = `Location,Elev. (m),FBP Fuel Type,Status,Temp (°C),RH (%),Wind Dir (°),Wind Speed (km/h),Precip (mm),Grass Cure (%),FFMC,DMC,DC,ISI,BUI,FWI,DGR CL,ROS (m/min),HFI,60 min fire size (ha),Fire Type,M/FIG,Fire Starts,Prep Level
+    Kamloops Fire Centre
+    Kamloops (K2), ,,,,,,,,,,,,,,,,,,,, 1, 0-1, 1
+    AF\`TON (322),780,O1B,ND,ND,ND,ND,ND,ND,ERROR,ND,ND,ND,ND,ND,ND,ND,ND,ND,ND,ND,ND
+
```
Sybrand, line 683: Redundant carriage return?

Sybrand (changes requested):
I see the double line break thing happening in the tests?

web/src/features/hfiCalculator/formatTableAsCSV.test.ts
andrea-williams: It's necessary for tests to pass. Otherwise it fails because the test suite appears to be inserting tab characters when a record spills over into the next line

Sybrand (approved):
Flames to dust
Lovers to friends
Why do all good things come to an end?
Flames to dust
Lovers to friends
Why do all good things come to an end?
Come to an end, come to an...
Why do all good things come to an end?
Come to an end, come to an...
Why do all good things come to an end?

PR #1515, opened 2021-11-08
Fire Zone Outlines Displayed on Map

JulianForeman:


andrea-williams (commented):
Is there any way to make the zone boundary lines thicker? The way that the AC for this ticket is written says that fire zones should be visible when the map is zoomed in on a fire centre, which isn't really happening now - users have to zoom in further than that before the lines are visible.

Sybrand:
> Is there any way to make the zone boundary lines thicker? The way that the AC for this ticket is written says that fire zones should be visible when the map is zoomed in on a fire centre, which isn't really happening now - users have to zoom in further than that before the lines are visible.

Using the WMS tileset available to us we're very limited. We basically get to use the style they gives us - and it sucks. To do anything better, we'd need to switch provides, which @conbrad is busy cracking for us.

Sybrand (approved):
The WMS layers totally suck. I think we merge this in, but circle back with a better solution.

andrea-williams (approved):

conbrad (approved):

PR #1516, opened 2021-11-08
Story/zooming fba map/1482

conbrad:
- Switches out fire zones WMS tile layer for vector layer from arcrest server
- Moves fire behaviour advisory page / components to own features folder
- Adds fire center vector layer from arcrest server, initial page load will be slow
  - And caches the large vectors for faster rendering on subsequent page loads because they're huge
  - Uses `idb` npm package for simplifying IndexedDB setup for caching vectors
- Reverse engineered fire center extents from map once vectors were inflated at runtime -- getting the extents from arcrest server / openlayers map was a painful avenue.
- Zooms to fire center extent on map when selected
- Zooms to full province view when no fire center is selected

Screen Shot 2021-11-10 at 11 11 31 AM

Assumptions:
- Outline colors are placeholders

# Test Links:
[Percentile Calculator](https://wps-pr-1516.apps.silver.devops.gov.bc.ca/)
[MoreCast](https://wps-pr-1516.apps.silver.devops.gov.bc.ca/morecast)
[C-Haines](https://wps-pr-1516.apps.silver.devops.gov.bc.ca/c-haines)
[FireBat](https://wps-pr-1516.apps.silver.devops.gov.bc.ca/fire-behaviour-calculator)
[FireBat bookmark](https://wps-pr-1516.apps.silver.devops.gov.bc.ca/fire-behaviour-calculator?s=266&f=c5&c=NaN&w=20,s=286&f=c7&c=NaN&w=16,s=1055&f=c7&c=NaN&w=NaN,s=305&f=c7&c=NaN&w=NaN,s=344&f=c5&c=NaN&w=NaN,s=346&f=c7&c=NaN&w=NaN,s=328&f=c7&c=NaN&w=NaN,s=1399&f=c7&c=NaN&w=NaN,s=334&f=c7&c=NaN&w=NaN,s=1082&f=c3&c=NaN&w=NaN,s=388&f=c7&c=NaN&w=NaN,s=309&f=c7&c=NaN&w=16,s=306&f=c7&c=NaN&w=NaN,s=1029&f=c7&c=NaN&w=NaN,s=298&f=c7&c=NaN&w=NaN,s=1108&f=c5&c=NaN&w=NaN,s=836&f=c7&c=NaN&w=NaN#state=2ec784ca-c46a-49d0-b2b3-1cf32a9015a2&session_state=7d9447c8-db66-4661-b4cb-03d2ac0d1d8f&code=32292df4-2bdf-4f90-a4a8-c8dbcda682a9.7d9447c8-db66-4661-b4cb-03d2ac0d1d8f.2b63f390-f3dc-43ae-89f2-016453863476)
[Fire Behaviour Advisory](https://wps-pr-1516.apps.silver.devops.gov.bc.ca/fire-behaviour-advisory)
[HFI Calculator](https://wps-pr-1516.apps.silver.devops.gov.bc.ca/hfi-calculator)


web/src/api/external/VectorDataManager.ts
```diff
@@ -0,0 +1,137 @@
...
+          extent,
+          this.buildKey(layer, extent),
+          data
+        )
+        const tx = this.db.transaction(VectorDataManager.dataStoreName, 'readwrite')
+        await Promise.all(records.map(record => tx.store.put(record)))
+      }
+      return data
+    } catch (error) {
+      console.log(error)
```
andrea-williams, line 61: is this all we want to do for catching errors? It doesn't really help users (I would expect most aren't going to know to inspect the console log). Is there something more helpful we can give as an error message to users?

web/src/features/fba/fireCenterExtents.ts
```diff
@@ -0,0 +1,50 @@
+export const extentsMap = new Map([
```
andrea-williams, line 1: I'd like to see a bit more documentation here. e.g., I'm guessing the coordinates are in BC Albers? What did you do to calculate the extents?

andrea-williams (changes requested):
Observation or possibly a question for @tessmgood - when I have my browser window to max size on my laptop screen (15"), once I've selected a fire centre from the dropdown, the map gets pushed down below the table. Shouldn't the map stay in line with the table? There looks to be a decent amount of whitespace between the table and the map - could this just be tightened up a little bit?

tessmgood:
> Observation or possibly a question for @tessmgood - when I have my browser window to max size on my laptop screen (15"), once I've selected a fire centre from the dropdown, the map gets pushed down below the table. Shouldn't the map stay in line with the table? There looks to be a decent amount of whitespace between the table and the map - could this just be tightened up a little bit?

@conbrad and I have spoken about this topic generally, my understanding in this small-tickets approach is that the layout of the page incl. responsiveness may need to be a separate ticket. My vision is for the map to fill 50% of the screen, and be responsive to screen size. Re: whitespace, keep in mind that the table is about to get much larger - anticipating approx. 12 columns

conbrad:
> > Observation or possibly a question for @tessmgood - when I have my browser window to max size on my laptop screen (15"), once I've selected a fire centre from the dropdown, the map gets pushed down below the table. Shouldn't the map stay in line with the table? There looks to be a decent amount of whitespace between the table and the map - could this just be tightened up a little bit?
> 
> @conbrad and I have spoken about this topic generally, my understanding in this small-tickets approach is that the layout of the page incl. responsiveness may need to be a separate ticket. My vision is for the map to fill 50% of the screen, and be responsive to screen size. Re: whitespace, keep in mind that the table is about to get much larger - anticipating approx. 12 columns

Thanks @tessmgood. Yes I'd like to avoid getting distracted and bogged down with layout/styling before getting the critical map features in. The acceptance criteria states:

- If there is no fire centre selected then we show entire province on the map
    - Given I land on the new advisory page, the page remembers my previous fire centre selection, and the selection is zoomed in to the extent that the edges (boundaries) of the fire centre fill the map container
- When I select a fire centre, the map is zoomed in to the extent that the edges (boundaries) of the fire centre fill the map container.

andrea-williams (approved):
Approving after discussion - layout to be addressed later. 

Just some minor documentation requests/questions

tessmgood (approved):

Sybrand (approved):

web/src/api/external/VectorDataManager.ts
conbrad: Yeah there's nothing we really can do if there are errors; users are not going to know what it means, and we have no control over fixing them.

Ideally we'd be sending stack traces to something like Sentry so we can track, reproduce, and fix errors we have control over, but we're not there yet: https://try.sentry-demo.com/organizations/correct-bird/issues/8322/?project=3013#exception

Also I'm hoping this code goes away soon.

PR #1539, opened 2021-11-17
Unnecessary Date Picker Reloads 

JulianForeman:
# Test Links:
[Percentile Calculator](https://wps-pr-1539.apps.silver.devops.gov.bc.ca/)
[MoreCast](https://wps-pr-1539.apps.silver.devops.gov.bc.ca/morecast)
[C-Haines](https://wps-pr-1539.apps.silver.devops.gov.bc.ca/c-haines)
[FireBat](https://wps-pr-1539.apps.silver.devops.gov.bc.ca/fire-behaviour-calculator)
[FireBat bookmark](https://wps-pr-1539.apps.silver.devops.gov.bc.ca/fire-behaviour-calculator?s=266&f=c5&c=NaN&w=20,s=286&f=c7&c=NaN&w=16,s=1055&f=c7&c=NaN&w=NaN,s=305&f=c7&c=NaN&w=NaN,s=344&f=c5&c=NaN&w=NaN,s=346&f=c7&c=NaN&w=NaN,s=328&f=c7&c=NaN&w=NaN,s=1399&f=c7&c=NaN&w=NaN,s=334&f=c7&c=NaN&w=NaN,s=1082&f=c3&c=NaN&w=NaN,s=388&f=c7&c=NaN&w=NaN,s=309&f=c7&c=NaN&w=16,s=306&f=c7&c=NaN&w=NaN,s=1029&f=c7&c=NaN&w=NaN,s=298&f=c7&c=NaN&w=NaN,s=1108&f=c5&c=NaN&w=NaN,s=836&f=c7&c=NaN&w=NaN#state=2ec784ca-c46a-49d0-b2b3-1cf32a9015a2&session_state=7d9447c8-db66-4661-b4cb-03d2ac0d1d8f&code=32292df4-2bdf-4f90-a4a8-c8dbcda682a9.7d9447c8-db66-4661-b4cb-03d2ac0d1d8f.2b63f390-f3dc-43ae-89f2-016453863476)
[Fire Behaviour Advisory](https://wps-pr-1539.apps.silver.devops.gov.bc.ca/fire-behaviour-advisory)
[HFI Calculator](https://wps-pr-1539.apps.silver.devops.gov.bc.ca/hfi-calculator)


web/cypress/integration/fire-behaviour-advisory-calculator-page.spec.ts
```diff
@@ -2,6 +2,30 @@ import { FIRE_BEHAVIOR_CALC_ROUTE } from '../../src/utils/constants'
...
+    expect(req.body).to.deep.include({
+      date: yesterday
+    })
+  }).as('calculateResults')
+
+  visitAndAddRow()
+
+  cy.wait('@getStations')
+
+  commandToUse(yesterday)
```
Sybrand, line 20:
```suggestion
  setDate(yesterday)
```

Just think it would read easier this way?

web/cypress/integration/fire-behaviour-advisory-calculator-page.spec.ts
```diff
@@ -2,6 +2,30 @@ import { FIRE_BEHAVIOR_CALC_ROUTE } from '../../src/utils/constants'
 import { FuelTypes } from '../../src/features/fbaCalculator/fuelTypes'
 import { DateTime } from 'luxon'
 
+const datePickerTest = (commandToUse: Function, visitAndAddRow: Function) => {
```
Sybrand, line 5:
```suggestion
const datePickerTest = (setDate: Function, visitAndAddRow: Function) => {
```

web/cypress/support/fba-commands.ts
```diff
@@ -86,8 +92,22 @@ Cypress.Commands.add('setFBAWindSpeed', (windSpeed: string, rowId: number) => {
...
+Cypress.Commands.add('setDateTypeMethod', (date: string) => {
+  return cy
+    .getByTestId('date-of-interest-picker')
+    .find('input')
+    .clear()
+    .type(date)
+    .type('{enter}')
+})
+
+Cypress.Commands.add('setDateBlurMethod', (date: string) => {
```
Sybrand, line 104: This isn't blur is it? This is typing the date, and then click there? A blur would be clicking somewhere else.

web/src/features/fba/pages/FireBehaviourAdvisoryPage.tsx
```diff
@@ -64,13 +64,11 @@ export const FireBehaviourAdvisoryPage: React.FunctionComponent = () => {
...
-    useState(DateTime.now().setZone('UTC-7').toISO())
-
-  const updateDate = () => {
-    if (previouslySelectedDateOfInterest !== dateOfInterest) {
-      setPreviouslySelectedDateOfInterest(dateOfInterest)
-    }
+  const updateDate = (date: string) => {
+    /*This needs to be done in order for timezone to function correctly and not have the 
+    date picker display the incorrect date*/
+    date = `${date}T00:00:00-08:00`
```
Sybrand, line 70: Please use constant instead of hard -08:00 in string. There's a -8 constant somwhere.

web/src/features/fba/pages/FireBehaviourAdvisoryPage.tsx
```diff
@@ -64,13 +64,11 @@ export const FireBehaviourAdvisoryPage: React.FunctionComponent = () => {
     DateTime.now().setZone('UTC-7').toISO()
```
Sybrand, line 64: I know this is from another PR - but where shouldn't have -7 in some places, and 8 in others - it should all be -8 I think?

Sybrand (changes requested):
Please see comments re. variable naming and constants.

web/cypress/support/fba-commands.ts
JulianForeman: This clicks off somewhere else because the mouse isn't focused on the input element.

web/src/features/hfiCalculator/pages/HfiCalculatorPage.tsx
```diff
@@ -64,23 +66,33 @@ const HfiCalculatorPage: React.FunctionComponent = () => {
 
   // the DatePicker component requires dateOfInterest to be in string format
   const [dateOfInterest, setDateOfInterest] = useState(
-    DateTime.now().setZone('UTC-7').toISO()
```
conbrad, line 67: Why do we need to switch our timezone for everything?

web/src/features/fba/pages/FireBehaviourAdvisoryPage.tsx
conbrad: Still using a string here

web/cypress/integration/fire-behaviour-advisory-calculator-page.spec.ts
```diff
@@ -133,28 +157,11 @@ describe('FireBAT Calculator Page', () => {
...
-      cy.wait('@getStations')
-
-      cy.setDate(yesterday)
-
-      cy.selectFBAStationInDropdown(322, 1)
-
-      cy.selectFBAFuelTypeInDropdown(FuelTypes.get()['c1'].friendlyName, 1)
-
-      cy.wait('@calculateResults')
+    it.only('Sets the date correctly when typing and pressing enter', () => {
```
conbrad, line 160: Need to remove the `only` call so all tests run

web/cypress/integration/fire-behaviour-advisory-calculator-page.spec.ts
```diff
@@ -2,6 +2,30 @@ import { FIRE_BEHAVIOR_CALC_ROUTE } from '../../src/utils/constants'
 import { FuelTypes } from '../../src/features/fbaCalculator/fuelTypes'
 import { DateTime } from 'luxon'
 
+const datePickerTest = (setDate: Function, visitAndAddRow: Function) => {
```
conbrad, line 5: It looks like these tests do stuff but aren't asserting anything, do we need to expect certain results?

web/src/components/DatePicker.tsx
```diff
@@ -1,37 +1,59 @@
...
+              //Gets the input component that holds the date value
+              .getElementsByTagName('input')[0]
+              .value.toString()
+              //Replaces the '/' in the date with '-' otherwise the formatting is incompatible
+              .replaceAll('/', '-')
+            event.preventDefault()
+            props.updateDate(newDate)
+          }
+        }}
+        onBlur={event => {
```
conbrad, line 43: If I click on the date text in the DatePicker, then click on the calendar button in it without changing anything the page reloads. It also looks like it removes a day from the existing date. Might need to check for the same date being set before updating it.

web/src/components/DatePicker.tsx
```diff
@@ -1,37 +1,59 @@
...
+            console.log('accept', newDate)
+          }
+        }}
+        onKeyDown={event => {
+          if (event.key === 'Enter') {
+            const newDate = event.currentTarget
+              //Gets the input component that holds the date value
+              .getElementsByTagName('input')[0]
+              .value.toString()
+              //Replaces the '/' in the date with '-' otherwise the formatting is incompatible
```
conbrad, line 37: Instead of this, can we do something like below for all the `on` methods:
```
const newDateString = event.currentTarget
              //Gets the input component that holds the date value
              .getElementsByTagName('input')[0]
              .value.toString()
              
const newDate = DateTime.from(newDateString).toISO()
```

This way we can leverage `luxon` to do the date formatting for us.

conbrad (changes requested):

web/cypress/integration/fire-behaviour-advisory-calculator-page.spec.ts
JulianForeman: `cy.intercept('POST', 'api/fba-calc/stations', req => {
    expect(req.body).to.deep.include({
      date: yesterday
    })` It happens here from what I could tell, I just repurposed a test that was already here.

web/src/features/fbaCalculator/components/FBATable.tsx
```diff
@@ -310,8 +312,11 @@ const FBATable = (props: FBATableProps) => {
     })
   }
 
-  const updateDate = () => {
-    dispatch(fetchFireBehaviourStations(dateOfInterest, rows))
+  const updateDate = (newDate: string) => {
+    if (newDate !== dateOfInterest) {
+      dispatch(fetchFireBehaviourStations(newDate, rows))
```
conbrad, line 317: We don't need this here, it gets dispatched in an effect hook

web/src/features/hfiCalculator/pages/HfiCalculatorPage.tsx
```diff
@@ -64,23 +64,25 @@ const HfiCalculatorPage: React.FunctionComponent = () => {
...
-      setPreviouslySelectedDateOfInterest(dateOfInterest)
+  const refreshView = () => {
+    const { start, end } = getDateRange(isWeeklyView, dateOfInterest)
+    callDispatch(start, end)
+  }
+
+  const updateDate = (newDate: string) => {
+    if (newDate !== dateOfInterest) {
+      setDateOfInterest(newDate)
+      const { start, end } = getDateRange(isWeeklyView, newDate)
```
conbrad, line 84: We're duplicating code that's already in `refreshView` here.

web/src/components/DatePicker.tsx
```diff
@@ -1,37 +1,63 @@
...
-          props.updateDate()
-        }
-      }}
-    />
+          props.updateDate(newDate)
+        }}
+        onChange={e => {
+          /*This is a required attribute we don't use because 
+          it makes editting the date with a keyboard impossible*/
+          console.log(e)
```
conbrad, line 57: Stray `console.log`

web/cypress/integration/fire-behaviour-advisory-calculator-page.spec.ts
conbrad: We need logical asserts for these tests to mean anything.

conbrad (changes requested):

web/src/features/hfiCalculator/pages/HfiCalculatorPage.tsx
conbrad: Discussed offline, let's keep this since refreshView doesn't have the latest date, even though it's set (asynchronous javascript problems).

web/cypress/integration/fire-behaviour-advisory-calculator-page.spec.ts
conbrad: My bad, I didn't see the expect assertions in the request body intercept

web/src/components/DatePicker.tsx
```diff
@@ -1,37 +1,63 @@
-import { TextField } from '@material-ui/core'
 import React from 'react'
+import LuxonUtils from '@date-io/luxon'
+import { KeyboardDatePicker, MuiPickersUtilsProvider } from '@material-ui/pickers'
+import { DateTime } from 'luxon'
 
 interface DatePickerProps {
   testId?: string
   date: string
   onChange: (d: string) => void
```
conbrad, line 9: We can remove this from props now since it's not used anymore

conbrad (approved):
Just one last thing and I'm happy

Sybrand (changes requested):
Firebat is now broken (it's picking the wrong date)
Can you please
- add a test to catch this edge case in firebat
- fix it

conbrad:
> Firebat is now broken (it's picking the wrong date) Can you please
> 
> * add a test to catch this edge case in firebat
> * fix it

My bad Julian, we need zones to be in UTC-7, not UTC-8. We can change all `.setZone('UTC-8')` to `.setZone('UTC-7').
But the regression test is also a good idea.

Sybrand:
> 

Please use the constants - that's why we have them.

tessmgood (changes requested):
Thanks for changing the component - UI pattern is much better! Two qs: 

Is there anything we can do to reduce some padding / the height of the component? MoreCast is a good example, where the select fields and button are closer in height. Can we implement that styling? (hoping its quick CSS) 
Screen Shot 2021-11-29 at 1 37 27 PM

Still noticing an unnecessary page load, if I click into the date select field on HFI or Fire Behaviour and click away without making changes, the page/table reloads, can we avoid this? I believe we achieved this, this summer on FireBAT.

tessmgood (approved):
NICE

_Capturing for transparency that the UI request was not straightforward given the Material UI component, interface changes coming in backlog tickets can address this feedback with an alternate approach._

PR #1568, opened 2021-11-22
Story/fire areas backend/1436

conbrad:
# Test Links:
[Percentile Calculator](https://wps-pr-1568.apps.silver.devops.gov.bc.ca/)
[MoreCast](https://wps-pr-1568.apps.silver.devops.gov.bc.ca/morecast)
[C-Haines](https://wps-pr-1568.apps.silver.devops.gov.bc.ca/c-haines)
[FireBat](https://wps-pr-1568.apps.silver.devops.gov.bc.ca/fire-behaviour-calculator)
[FireBat bookmark](https://wps-pr-1568.apps.silver.devops.gov.bc.ca/fire-behaviour-calculator?s=266&f=c5&c=NaN&w=20,s=286&f=c7&c=NaN&w=16,s=1055&f=c7&c=NaN&w=NaN,s=305&f=c7&c=NaN&w=NaN,s=344&f=c5&c=NaN&w=NaN,s=346&f=c7&c=NaN&w=NaN,s=328&f=c7&c=NaN&w=NaN,s=1399&f=c7&c=NaN&w=NaN,s=334&f=c7&c=NaN&w=NaN,s=1082&f=c3&c=NaN&w=NaN,s=388&f=c7&c=NaN&w=NaN,s=309&f=c7&c=NaN&w=16,s=306&f=c7&c=NaN&w=NaN,s=1029&f=c7&c=NaN&w=NaN,s=298&f=c7&c=NaN&w=NaN,s=1108&f=c5&c=NaN&w=NaN,s=836&f=c7&c=NaN&w=NaN#state=2ec784ca-c46a-49d0-b2b3-1cf32a9015a2&session_state=7d9447c8-db66-4661-b4cb-03d2ac0d1d8f&code=32292df4-2bdf-4f90-a4a8-c8dbcda682a9.7d9447c8-db66-4661-b4cb-03d2ac0d1d8f.2b63f390-f3dc-43ae-89f2-016453863476)
[Fire Behaviour Advisory](https://wps-pr-1568.apps.silver.devops.gov.bc.ca/fire-behaviour-advisory)
[HFI Calculator](https://wps-pr-1568.apps.silver.devops.gov.bc.ca/hfi-calculator)


api/alembic/versions/304daefe5f80_import_vernon_thessian_polygons.py
```diff
@@ -0,0 +1,28 @@
...
+
+# revision identifiers, used by Alembic.
+revision = '304daefe5f80'
+down_revision = '11b821827824'
+branch_labels = None
+depends_on = None
+
+
+file_path = os.getcwd() + '/alembic/vernon_area_prototype/VernonFZBATPolys.shp'
+command = f'ogr2ogr -f "PostgreSQL" PG:"dbname=wps user=postgres" "{file_path}" -nln fire_area_thessian_polygons -append'
```
Sybrand, line 20: This is probably going to fail - I had to do a `sudo apt install gdal-bin` 1st.

api/app/db/models/thessian_polygon_areas.py
```diff
@@ -0,0 +1,19 @@
...
+from app.db.database import Base
+
+
+class ThessianPolygonArea(Base):
+    """ Thessian polygon for a given area """
+    __tablename__ = 'fire_area_thessian_polygons'
+
+    id = Column(Integer, Sequence('thessian_area_id_seq'),
+                primary_key=True, nullable=False, index=True)
+    polygon = Column(Geometry(geometry_type='POLYGON', from_text='ST_GeomFromGeoJSON'))
```
Sybrand, line 15: You need to be explicit about the creation of an index on the geometry.  You'll need something like
`
Index('idx_fire_area_thessian_polygons_geom',
      ThessianPolygonArea.geom, postgresql_using='gist')`

in the global scope. See api/app/db/models/weather_models.py

If your not explicit, then subsequent runs of alembic autogenerate will create a migration to remove the index (unless that's been fixed since!)

Sybrand:
```suggestion
    geom = Column(Geometry(geometry_type='POLYGON', from_text='ST_GeomFromGeoJSON'))
```
Convention is to call the geometry column "geom"

Sybrand: I think having the from_text='ST_GeomFromGeoJSON' implies that the data is stored as json. We don't want to store it as json because then we lose a lot of the veratility that postgis gives us, we want this to be binary. I'm not sure about this though - can you check?

api/app/db/models/thessian_polygon_areas.py
```diff
@@ -0,0 +1,19 @@
...
+from sqlalchemy import (Column, Integer,
+                        Sequence)
+from app.db.database import Base
+
+
+class ThessianPolygonArea(Base):
+    """ Thessian polygon for a given area """
+    __tablename__ = 'fire_area_thessian_polygons'
+
+    id = Column(Integer, Sequence('thessian_area_id_seq'),
```
Sybrand, line 13: Please make id match table name

api/app/db/models/thessian_polygon_areas.py
```diff
@@ -0,0 +1,19 @@
+""" Class models that reflect resources and map to database tables for HFI Calculator.
+"""
+from geoalchemy2.types import Geometry
+from sqlalchemy import (Column, Integer,
+                        Sequence)
+from app.db.database import Base
+
+
+class ThessianPolygonArea(Base):
```
Sybrand, line 9: Pedantic, but classname differs from tablename.

api/app/db/models/thessian_polygon_areas.py
```diff
@@ -0,0 +1,19 @@
...
+    """ Thessian polygon for a given area """
+    __tablename__ = 'fire_area_thessian_polygons'
+
+    id = Column(Integer, Sequence('thessian_area_id_seq'),
+                primary_key=True, nullable=False, index=True)
+    polygon = Column(Geometry(geometry_type='POLYGON', from_text='ST_GeomFromGeoJSON'))
+
+    def __str__(self):
+        return (f'id:{self.id}, '
+                f'fire_center:{self.polygon}')
```
Sybrand, line 19: I would expect the geom to be binary gobbledygook - but not sure.

Sybrand (commented):
Why do we have the pdf checked in?

api/app/db/models/thessian_polygon_areas.py
Sybrand: The source feature has some properties we should probably bring in. Most pertinent/only one I think we need is the station code.

api/app/db/models/thessian_polygon_areas.py
```diff
@@ -0,0 +1,19 @@
+""" Class models that reflect resources and map to database tables for HFI Calculator.
+"""
+from geoalchemy2.types import Geometry
+from sqlalchemy import (Column, Integer,
+                        Sequence)
+from app.db.database import Base
+
+
+class ThessianPolygonArea(Base):
+    """ Thessian polygon for a given area """
```
Sybrand, line 10: We need to comply to standards, so we need to add in a "create_date" and "update_date"
e.g.:

`create_date = Column(TZTimeStamp, nullable=False,
                         default=time_utils.get_utc_now())
    # Date this record was updated.
    update_date = Column(TZTimeStamp, nullable=False,
                         default=time_utils.get_utc_now())`

api/alembic/versions/304daefe5f80_import_vernon_thessian_polygons.py
Sybrand: You can't have user=postgres - that's not going to work across all platforms. you'll have to read in the username and password from config - also - you won't have that database local, you'll have to read the write db from config ; this may not even be a valid approach at all, because it means we'd need to have gdal-bin installed on all the pods purely for the case of bootstrapping. (so that means re-building the python base image)

I think ogr2ogr actually just neatly wraps a bunch of python that's already imported in our project with osgeo/gdal - so we should be able to do all of this from python script. Just takes some looking into.

api/app/db/models/thessian_polygon_areas.py
conbrad: Hmm, we don't have these on all tables, and the tables that do have them have different field names so I didn't think there was a standard.

api/app/db/models/thessian_polygon_areas.py
conbrad: Cannot seem to find docs on how postgis stores geometries.

https://postgis.net/docs/ST_GeomFromGeoJSON.html
Above says "Constructs a PostGIS geometry object from the GeoJSON representation." It reads to me as if PostGIS stores it one way but specifying a different ST geom data type function will marshall inputs/outputs using the function specified.

By default alembic seems to use `from_text='ST_GeomFromEWKT'`

Sybrand: Ok - that seems to make sense. I may just tinker with that when I do the shapefile thing - that's actually pretty cool if that's how it works - align with your input and output.

conbrad:
Closing in favor of: https://github.com/bcgov/wps/pull/1573

Issue #1650, opened 2022-01-17
HFI Calculator: Restructure page header


tessmgood:
**As a** *Z/RWCO* **I want** *a clear interface* **So That** *I can easily understand the info and actions as displayed and make clear decisions*
  
**Additional Context**
![Screen Shot 2022-01-17 at 9.37.49 AM.png](https://images.zenhubusercontent.com/60b527e413189cbf16c31d48/3307ed96-d737-477b-a2d9-e2bc9f66a6bb)

**Acceptance Criteria**
- [x] remove 'Predictive Services Unit' as primary header - change to 'HFI Calculator' 
- [x] change 'Contact' link to include this context: 'Contact Predictive Services Unit'
- [ ] move date selection and fire centre dropdown to secondary menu bar (similar to MoreCast header) 
- [ ] move 'About this data' modal link to the secondary header (right side)

**Definition of Done**
- [ ] Ready to Demo in Sprint Review
- [ ] Does what I have made have appropriate test coverage?
- [ ] Documentation and/or scientific documentation exists and can be found
- [ ] Peer Reviewed by 2 people on the team
- [ ] Manual testing of all PRs in Dev and Prod
- [ ] Merged

conbrad:
Easy to move the date selector up to the page title, tricky bit will be styling it with colors, and padding to fit in the hardcoded 60px height of the page header.

andrea-williams:
Since my current ticket (#1626) includes adding a dropdown to select the fire centre, I could just include these AC in my current ticket...

Sybrand:
@andrea-williams I'd advise against scope creep in #1626...

nmathava:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Wildfire-Predictive-Services-5e321393e038fba5bbe203b8/issues/bcgov/wps/1650?planning-poker) with ZenHub @andrea-williams @conbrad @JulianForeman @Sybrand

PR #1652, opened 2022-01-17
Story/hfi add fcs/1626

andrea-williams:
# Test Links:
[Percentile Calculator](https://wps-pr-1652.apps.silver.devops.gov.bc.ca/)
[MoreCast](https://wps-pr-1652.apps.silver.devops.gov.bc.ca/morecast)
[C-Haines](https://wps-pr-1652.apps.silver.devops.gov.bc.ca/c-haines)
[FireBat](https://wps-pr-1652.apps.silver.devops.gov.bc.ca/fire-behaviour-calculator)
[FireBat bookmark](https://wps-pr-1652.apps.silver.devops.gov.bc.ca/fire-behaviour-calculator?s=266&f=c5&c=NaN&w=20,s=286&f=c7&c=NaN&w=16,s=1055&f=c7&c=NaN&w=NaN,s=305&f=c7&c=NaN&w=NaN,s=344&f=c5&c=NaN&w=NaN,s=346&f=c7&c=NaN&w=NaN,s=328&f=c7&c=NaN&w=NaN,s=1399&f=c7&c=NaN&w=NaN,s=334&f=c7&c=NaN&w=NaN,s=1082&f=c3&c=NaN&w=NaN,s=388&f=c7&c=NaN&w=NaN,s=309&f=c7&c=NaN&w=16,s=306&f=c7&c=NaN&w=NaN,s=1029&f=c7&c=NaN&w=NaN,s=298&f=c7&c=NaN&w=NaN,s=1108&f=c5&c=NaN&w=NaN,s=836&f=c7&c=NaN&w=NaN#state=2ec784ca-c46a-49d0-b2b3-1cf32a9015a2&session_state=7d9447c8-db66-4661-b4cb-03d2ac0d1d8f&code=32292df4-2bdf-4f90-a4a8-c8dbcda682a9.7d9447c8-db66-4661-b4cb-03d2ac0d1d8f.2b63f390-f3dc-43ae-89f2-016453863476)
[Fire Behaviour Advisory](https://wps-pr-1652.apps.silver.devops.gov.bc.ca/fire-behaviour-advisory)
[HFI Calculator](https://wps-pr-1652.apps.silver.devops.gov.bc.ca/hfi-calculator)


tessmgood (commented):
Nicee! Thank you for making the empty state. 
- I will write a ticket for the sticky columns issue in the header, in the empty state. 

Can we make the fire centre dropdown the same height as all the other components on that line - buttons and select fields? Either now or in #1650 when we restructure the header.

andrea-williams:
> Nicee! Thank you for making the empty state.
> 
> * I will write a ticket for the sticky columns issue in the header, in the empty state.
> 
> Can we make the fire centre dropdown the same height as all the other components on that line - buttons and select fields? Either now or in #1650 when we restructure the header.

Ask and it shall be given! ... Yes I've made a change so that the fire centre dropdown matches the height of the date selector. I don't know how I didn't notice that myself.

web/src/api/hfiCalculatorAPI.ts
```diff
@@ -44,12 +46,20 @@ const url = '/hfi-calc/daily'
...
   const { data } = await axios.get<StationDailyResponse>(url, {
     params: {
       start_time_stamp: startTime,
-      end_time_stamp: endTime
+      end_time_stamp: endTime,
+      station_codes: stationCodes
+    },
+    // have to add a paramsSerializer to axios get request and stringify (using querystring library) the params in order
+    // for the stationCodes array to be formatted correctly, per
+    // https://stackoverflow.com/a/51444749
```
conbrad, line 60: We can do this as a one-liner without having to add another dependency: https://stackoverflow.com/a/63402277

web/src/features/hfiCalculator/pages/HfiCalculatorPage.tsx
```diff
@@ -141,6 +139,21 @@ const HfiCalculatorPage: React.FunctionComponent = () => {
     setModalOpen(true)
   }
 
+  const getAllPlanningWeatherStationCodesFromFireCentre = (
+    centre: FireCentre | undefined
+  ): number[] => {
+    const stationCodes: number[] = []
```
conbrad, line 145: This function can be simplified to:
```
Object.values(centre.planning_areas).flatMap((area) => area.stations.map(station => station.code))
```

local_deploy.sh
```diff
@@ -0,0 +1,28 @@
+#!/bin/bash
+ 
+# Usually "pr-<PR number>" - e.g., "pr-1600" 
+SUFFIX="pr-1666"
```
conbrad, line 4: I think if we want to commit this script we should:
- parameterize and require the pr suffix parameter, so we prevent accidental prod deploys
- have code shared between this script and the actual production.yml job so things don't go out of sync if/when we make changes

I think it might be better to write a dev task ticket for this and add this in it's own PR

api/alembic/versions/b557469a7727_modify_unique_constraint_on_planning_.py
```diff
@@ -0,0 +1,30 @@
...
+revision = 'b557469a7727'
+down_revision = '1caf3488a340'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.drop_constraint('planning_weather_stations_station_code_key', 'planning_weather_stations', type_='unique')
+    op.create_unique_constraint(None, 'planning_weather_stations', ['station_code', 'planning_area_id'])
```
conbrad, line 22: Can we replace `None` with an explicit name from the constraint? I think the library auto-creates a name but it broke when I was writing a migration before and it's not that descriptive. The name can be a variable used in both up/down functions.

api/alembic/versions/481905f1dfe8_merge_alembic_heads.py
```diff
@@ -0,0 +1,26 @@
...
+
+# revision identifiers, used by Alembic.
+revision = '481905f1dfe8'
+down_revision = ('8efe0e7b9712', '39806f02cdec')
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # No action necessary, just merging 2 Alembic version HEADs into one
```
conbrad, line 20: Migrations should be serial, no? What's the case for diverging migrations?

conbrad (changes requested):

api/alembic/versions/481905f1dfe8_merge_alembic_heads.py
andrea-williams: It came about from that nightmare Alembic scenario I had, that I chatted with you & Sy about (where I foolishly tried to use the same .json source data for 2 different migrations).

local_deploy.sh
andrea-williams: sounds good. I'll remove the file from this PR and write a dev task ticket for it instead

web/src/api/hfiCalculatorAPI.ts
andrea-williams: tbh I prefer it the way it is - the code is cleaner

local_deploy.sh
andrea-williams: Ticket created here: https://github.com/bcgov/wps/issues/1670

web/src/api/hfiCalculatorAPI.ts
conbrad: I'm less concerned about how it looks vs. adding another dependency. We're currently running into updating `react-scripts` because we introduced `csv-string`, which depends on Node streams. The more packages we add the more pain we experience with maintenance. Let's keep this for now and dev chat as a team.

web/src/features/hfiCalculator/pages/HfiCalculatorPage.tsx
```diff
@@ -151,24 +151,27 @@ const HfiCalculatorPage: React.FunctionComponent = () => {
   }
 
   const copyTable = () => {
-    if (isWeeklyView) {
+    if (isUndefined(selectedFireCentre)) {
```
conbrad, line 154: To simplify, how about:

```
if(!isUndefined(selectedFireCentre)) {
// do work
}
```

Removes the else and unnecessary state change to the copy state

web/src/features/hfiCalculator/pages/HfiCalculatorPage.tsx
```diff
@@ -213,7 +216,7 @@ const HfiCalculatorPage: React.FunctionComponent = () => {
     ) {
       localStorage.setItem('hfiCalcPreferredFireCentre', selectedFireCentre?.name)
     }
-    const { start, end } = getDateRange(isWeeklyView, dateOfInterest)
+    const { start, end } = getDateRange(true, dateOfInterest)
```
conbrad, line 219: What does this change do? Is it fixing the data loading bug when a user is in the daily view?

conbrad (changes requested):

conbrad:
I'm getting those `toFixed` errors in console when I try to copy to CSV in the daily view:
[toFixedError.log](https://github.com/bcgov/wps/files/7952147/toFixedError.log)

web/src/features/hfiCalculator/pages/HfiCalculatorPage.tsx
andrea-williams: Yes. Basically, if a user changes the fire centre, we want to make 1 call to retrieve all data for the prep period

api/alembic/versions/b557469a7727_modify_unique_constraint_on_planning_.py
```diff
@@ -0,0 +1,25 @@
...
+
+# revision identifiers, used by Alembic.
+revision = 'b557469a7727'
+down_revision = 'd99fcdc4800d'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.create_unique_constraint('unique_station_code_for_planning_area',
```
conbrad, line 20: There's an existing unique constraint on this table just for station code, removing it was missed in this migration and causes some issues locally, hence the `api/alembic/versions/af108f346073_remove_redundant_constraint.py` to get prod into a good state.

Sybrand (changes requested):
Please run against copy of production to ensure that this actually works. This fails on my local - which was a restore from prod from some time ago.

api/alembic/versions/6d2f4d058ebc_add_column_for_order_of_appearance_in_.py
```diff
@@ -0,0 +1,54 @@
...
+                                'order_of_appearance_in_list', 'fire_centre_id'])
+    op.create_table_comment(
+        'planning_areas',
+        'Only one planning area can be assigned a position in the list for a fire centre',
+        existing_comment=None,
+        schema=None
+    )
+
+    # Load json file with planning areas order data
+    with open('alembic/versions/6d2f4d058ebc_add_column_for_order_of_appearance_in_.json') as json_file:
```
Sybrand, line 32: Better to split the loading of data from the updating of the schema.

PR #1705, opened 2022-02-08
Story/hfi calc backend/1685

conbrad:
- Implements HFI calculation logic on the backend, including test coverage
- Isolated change that is dormant, follow up PR will hook this up to the frontend and remove the logic there

# Test Links:
[Percentile Calculator](https://wps-pr-1705.apps.silver.devops.gov.bc.ca/)
[MoreCast](https://wps-pr-1705.apps.silver.devops.gov.bc.ca/morecast)
[C-Haines](https://wps-pr-1705.apps.silver.devops.gov.bc.ca/c-haines)
[FireBat](https://wps-pr-1705.apps.silver.devops.gov.bc.ca/fire-behaviour-calculator)
[FireBat bookmark](https://wps-pr-1705.apps.silver.devops.gov.bc.ca/fire-behaviour-calculator?s=266&f=c5&c=NaN&w=20,s=286&f=c7&c=NaN&w=16,s=1055&f=c7&c=NaN&w=NaN,s=305&f=c7&c=NaN&w=NaN,s=344&f=c5&c=NaN&w=NaN,s=346&f=c7&c=NaN&w=NaN,s=328&f=c7&c=NaN&w=NaN,s=1399&f=c7&c=NaN&w=NaN,s=334&f=c7&c=NaN&w=NaN,s=1082&f=c3&c=NaN&w=NaN,s=388&f=c7&c=NaN&w=NaN,s=309&f=c7&c=NaN&w=16,s=306&f=c7&c=NaN&w=NaN,s=1029&f=c7&c=NaN&w=NaN,s=298&f=c7&c=NaN&w=NaN,s=1108&f=c5&c=NaN&w=NaN,s=836&f=c7&c=NaN&w=NaN#state=2ec784ca-c46a-49d0-b2b3-1cf32a9015a2&session_state=7d9447c8-db66-4661-b4cb-03d2ac0d1d8f&code=32292df4-2bdf-4f90-a4a8-c8dbcda682a9.7d9447c8-db66-4661-b4cb-03d2ac0d1d8f.2b63f390-f3dc-43ae-89f2-016453863476)
[Fire Behaviour Advisory](https://wps-pr-1705.apps.silver.devops.gov.bc.ca/fire-behaviour-advisory)
[HFI Calculator](https://wps-pr-1705.apps.silver.devops.gov.bc.ca/hfi-calculator)


api/app/schemas/hfi_calc.py
```diff
@@ -41,6 +42,101 @@ class StationDailyResponse(BaseModel):
...
+    selected_station_codes: List[int]
+    selected_fire_center: Optional[str]
+    planning_area_hfi_results: Mapping[str, PlanningAreaResult]
+    planning_area_fire_starts: Mapping[str, List[FireStartRange]]
+
+
+class HFIResultRequest(BaseModel):
+    num_prep_days: int
+    selected_prep_date: datetime
+    start_time_stamp: Optional[int]
```
Sybrand, line 133: I'm assuming this is the prep period start, and "end_time_stamp" is the prep period end? Can we makes these iso date strings instead of timestamps please? Everything past the day of the month is redundant and just gives us time zone nightmares. We only need to know the start date: "2022-02-09" and end date "2022-02-12".

api/app/schemas/hfi_calc.py
```diff
@@ -41,6 +42,101 @@ class StationDailyResponse(BaseModel):
...
+    end_time_stamp: Optional[int]
+    selected_station_codes: List[int]
+    selected_fire_center: Optional[str]
+    planning_area_hfi_results: Mapping[str, PlanningAreaResult]
+    planning_area_fire_starts: Mapping[str, List[FireStartRange]]
+
+
+class HFIResultRequest(BaseModel):
+    num_prep_days: int
+    selected_prep_date: datetime
```
Sybrand, line 132: For the selected prep date, we don't need datetime, date will do. (See below)

api/app/schemas/hfi_calc.py
```diff
@@ -41,6 +42,101 @@ class StationDailyResponse(BaseModel):
...
+    start_time_stamp: Optional[int]
+    end_time_stamp: Optional[int]
+    selected_station_codes: List[int]
+    selected_fire_center: Optional[str]
+    planning_area_hfi_results: Mapping[str, PlanningAreaResult]
+    planning_area_fire_starts: Mapping[str, List[FireStartRange]]
+
+
+class HFIResultRequest(BaseModel):
+    num_prep_days: int
```
Sybrand, line 131: If we have pre start date and prep end date, we don't need period, and vice versa.

api/app/schemas/hfi_calc.py
```diff
@@ -41,6 +42,101 @@ class StationDailyResponse(BaseModel):
...
+    selected_prep_date: datetime
+    start_time_stamp: Optional[int]
+    end_time_stamp: Optional[int]
+    selected_station_codes: List[int]
+    selected_fire_center: Optional[str]
```
Sybrand, line 136: Are we passing around the fire centre description, or the fire centre id? I feel like we should be passing an id around. (One could get super technical and say we shouldn't pass out internal id around, but I think it's fine)

api/app/schemas/hfi_calc.py
```diff
@@ -41,6 +42,101 @@ class StationDailyResponse(BaseModel):
...
+    planning_area_fire_starts: Mapping[str, List[FireStartRange]]
```
Sybrand, line 137: Similar comment on planning area key - I'm guessing that key should be an id - in which case it's an integer.

api/app/schemas/hfi_calc.py
Sybrand: Also - if it's an id it should probably be an integer.

api/app/schemas/hfi_calc.py
```diff
@@ -41,6 +42,101 @@ class StationDailyResponse(BaseModel):
...
+    selected_prep_date: datetime
+    start_time_stamp: Optional[int]
+    end_time_stamp: Optional[int]
+    selected_station_codes: List[int]
+    selected_fire_center: Optional[str]
+    planning_area_hfi_results: Mapping[str, PlanningAreaResult]
+    planning_area_fire_starts: Mapping[str, List[FireStartRange]]
+
+
+class HFIResultRequest(BaseModel):
```
Sybrand, line 130: When you're building the request - keep in mind that the back-end is going to come back with a prep period (either default or stored - most likely stored) - so the front end should start off with null values for prep period to allow back-end to provide it with correct period. Only once the user changes it, should the request contain the result. (Same principle for all other fields - keep in mind whatever the front end sends down is going to be stored in back-end, so initial load request should have all values unpopulated)

api/app/schemas/hfi_calc.py
conbrad: Currently, the frontend doesn't know about the planning area id's it just knows it's name: https://github.com/bcgov/wps/blob/d3b180a9246a34404a4ad8eb6148be9d27e9f1b1/api/app/schemas/hfi_calc.py#L58

api/app/schemas/hfi_calc.py
conbrad: Agreed, but same as `PlanningArea`, the `FireCentre` schema only has a string name, not an id:
https://github.com/bcgov/wps/blob/d3b180a9246a34404a4ad8eb6148be9d27e9f1b1/api/app/schemas/hfi_calc.py#L66

Sybrand: Oi. Well - either of us can fix that - if your stuff works without that fix, I'll just have to fix it in my PR.

api/app/schemas/hfi_calc.py
Sybrand: It would save me time if you did, but you could argue it's out of scope for your PR. You didn't have that re-factor as part of your scope, and it isn't required for your story. Up to you if you want to do it now.

api/app/schemas/hfi_calc.py
conbrad: Yeah, I'd like to leave it for this PR because changing it now will break the frontend I think, and I don't want to touch that yet. In the PR where I adjust the frontend I can add id's to these schemas.

api/app/routers/hfi_calc.py
```diff
@@ -18,12 +19,50 @@
 
 logger = logging.getLogger(__name__)
 
+ageless_cache = "max-age=0"  # don't let the browser cache this
```
Sybrand, line 22: Ageless cache is just slightly weird for me - could sound like the cache never expires. (the exact opposite)
I'm bad with names, but what about "no_cache" or something like that? As opposed to a vampire cache?

api/app/schemas/hfi_calc.py
conbrad: Agreed, I'd like to address this in the next PR though, as it will affect how the frontend works.

api/app/schemas/hfi_calc.py
conbrad: Added to: https://github.com/bcgov/wps/pull/1705#issuecomment-1035265778

api/app/schemas/hfi_calc.py
conbrad: Added to: https://github.com/bcgov/wps/pull/1705#issuecomment-1035265778

api/app/schemas/hfi_calc.py
conbrad: Added to: https://github.com/bcgov/wps/pull/1705#issuecomment-1035265778

api/app/schemas/hfi_calc.py
conbrad: Added to: https://github.com/bcgov/wps/pull/1705#issuecomment-1035265778

api/app/schemas/hfi_calc.py
```diff
@@ -36,11 +36,95 @@ class StationDaily(BaseModel):
...
+    """
+    Request that contains inputs necessary for calculating HFI
+    """
+    num_prep_days: int
+    selected_prep_date: datetime
+    start_time_stamp: Optional[int]
+    end_time_stamp: Optional[int]
+    selected_station_codes: List[int]
```
Sybrand, line 123: Same as above, station is unique per planning area, not centre, so this needs to be grouped.

api/app/hfi/hfi.py
```diff
@@ -0,0 +1,118 @@
...
+    valid_prep_levels = list(filter(None, prep_levels))
+    return None if len(valid_prep_levels) == 0 else mean(valid_prep_levels)
+
+
+def calculate_mean_intensity(dailies: List[StationDaily]):
+    """ Returns the mean intensity group from a list of values """
+    intensity_groups = list(map(lambda daily: (daily.intensity_group), dailies))
+    valid_intensity_groups = list(filter(None, intensity_groups))
+    return None if len(valid_intensity_groups) == 0 else \
+        round((10 * sum(valid_intensity_groups) / len(valid_intensity_groups))) / 10
```
andrea-williams, line 97: couldn't you just use `mean(valid_intensity_groups)` like you've done for mean prep level?

api/app/routers/hfi_calc.py
```diff
@@ -42,7 +81,7 @@ async def get_daily_view(response: Response,
     """ Returns daily metrics for each station code. """
     try:
         logger.info('/hfi-calc/daily')
-        response.headers["Cache-Control"] = "max-age=0"  # don't let the browser cache this
+        response.headers["Cache-Control"] = no_cache
         valid_start_time, valid_end_time = validate_time_range(start_time_stamp, end_time_stamp)
```
andrea-williams, line 85: I know you haven't made any changes to this line of code, but I just noticed it, and huh??? How is this not infinitely recursive - the method is calling itself ??????

api/app/hfi/hfi.py
conbrad: I've been wondering this too and think it makes sense. This code is verbatim what's on the frontend and I cannot remember why it is this way.

api/app/routers/hfi_calc.py
conbrad: Hmm I don't see where any method is calling itself here. `validate_time_range` is not recursive.

api/app/schemas/hfi_calc.py
```diff
@@ -36,11 +36,95 @@ class StationDaily(BaseModel):
...
+    selected_fire_center: Optional[str]
+    planning_area_fire_starts: Mapping[str, List[FireStartRange]]
```
Sybrand, line 125: Another issue here - I'm guessing the first key is the planning area, but you also need a key for the day. Each day in the prep period.

andrea-williams (approved):

api/app/hfi/hfi.py
conbrad: Made the change and the tests are happy, just pushed it.

api/app/schemas/hfi_calc.py
conbrad: Added to the list: https://github.com/bcgov/wps/pull/1705#issuecomment-1035265778

PR #1711, opened 2022-02-12
Feature/hfi persist changes rethink/1686

Sybrand:
- persist requests
- test hfi endpoint


conbrad (approved):
LGTM

api/app/db/crud/hfi_calc.py
```diff
@@ -21,9 +24,50 @@ def get_all_stations(session: Session) -> CursorResult:
...
+        .filter(PlanningArea.fire_centre_id == fire_centre_id)\
+        .order_by(PlanningArea.order_of_appearance_in_list)
+
+
+def get_fire_centre_planning_area_stations(session, fire_centre_id: int) -> CursorResult:
+    """ Get all the stations for a planning area """
+    return session.query(PlanningWeatherStation)\
+        .join(PlanningArea, PlanningArea.id == PlanningWeatherStation.planning_area_id)\
+        .filter(PlanningArea.fire_centre_id == fire_centre_id)\
+        .order_by(PlanningArea.id)
```
andrea-williams, line 39: I'm confused by this function - is it returning all stations for a specific planning area or a specific fire centre? The documentation/comment says PA, the method signature suggests FC?

Sybrand: Maybe you can think of a better name for it? The relationship is Fire Centre -> Planning Area -> Station - so saying "get fire centre station" is really the same thing as "get fire centre planning area stations" - but less confusing? Should I just go with "get_fire_centre_stations" ?
Obviously have to update the documentation - yes - for sure.

Sybrand: I've pushed a change - have a look if that's better?

andrea-williams: Gotcha. I think the name of the function is fine, I was just confusing by planning area in the documentation, `fire_centre_id` in the parameters

andrea-williams: Next question - I haven't found yet where this method gets called, but does it make sense to order the results by `PlanningArea.id`? Shouldn't they be getting sorted according to the planning area's order within the fire centre, and each station's order within the planning area? (Obviously this could overlap with my ticket - I just want to get a sense of what extra changes I might need to make in my PR related to this PR)

Sybrand: Oh - thanks for spotting that! It was for iterating through with grouped planning areas, but I fixed that - I'll remove the sorting.

api/app/routers/hfi_calc.py
```diff
@@ -27,37 +30,94 @@
...
+def save_request(request: HFIResultRequest, username: str):
+    """ Save the request to the database (if there's a valid prep period) """
+    if request.start_date is not None and request.end_date is not None:
+        with app.db.database.get_write_session_scope() as session:
+            store_hfi_request(session, request, username)
+    return request
+
+
+def validate_date_range(start_date: date, end_date: date):
+    """ Sets the start_date to today if it is None.
```
andrea-williams, line 52: is it necessary to also validate that the date range is a minimum of 2 days? I believe feedback we got from the last sprint review was that prep periods can be anywhere from 2-7 days inclusive.

andrea-williams (approved):
One more question.
And I didn't check for the spelling mistakes I'm sure you've snuck in somewhere. I'm not taking responsibility for that.
Otherwise I'm happy.

api/app/routers/hfi_calc.py
Sybrand: Oh dear - I missed that - I'll update.

Sybrand: I'll default to 5, limit to 7 max and min 2.

PR #1780, opened 2022-03-14
Bug/empty fire centre spinner/1759

conbrad:
- Handle the case where fire centre isn't selected yet
- Reuse component across daily and weekly views

### DIFF
--- a/web/src/features/hfiCalculator/pages/HfiCalculatorPage.tsx
+++ b/web/src/features/hfiCalculator/pages/HfiCalculatorPage.tsx
@@ -16,7 +16,13 @@ import {
   selectHFIStationsLoading,
   selectHFICalculatorState
 } from 'app/rootReducer'
-import { CircularProgress, FormControl, makeStyles } from '@material-ui/core'
+import {
+  CircularProgress,
+  FormControl,
+  makeStyles,
+  Table,
+  TableBody
+} from '@material-ui/core'
 import { getDateRange, pstFormatter } from 'utils/date'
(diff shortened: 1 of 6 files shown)


andrea-williams:
Can't test atm... Not Found :(

web/src/features/hfiCalculator/pages/HfiCalculatorPage.tsx
```diff
@@ -255,6 +262,23 @@ const HfiCalculatorPage: React.FunctionComponent = () => {
     }
   }
 
+  const buildNonReadyState = () => {
+    if (isUndefined(result) && isUndefined(selectedFireCentre)) {
```
Sybrand, line 266: Already doing an isUndefined check down below.
What about a flow where it's:
```
if (isUndefined(selectedFireCentre)) {
   return gimmeTheEmptyFireCentreRow()
}
else if (loading || stationDataLoading || isUndefined(result)) {
   return gimmeTheLoadingCircle()
}
else {
  return gimmeTheTableWithAllTheStuff()
}
```

Sybrand (commented):
I propose a different flow.... what you think?

web/src/features/hfiCalculator/pages/HfiCalculatorPage.tsx
conbrad: Sonarcloud will complain about complexity in the component JSX :(

Sybrand (approved):
lgtm. (assuming it passes manual & automatic testing)

andrea-williams (approved):
Beauty.

PR #1789, opened 2022-03-16
Explicit fire starts

Sybrand:
- [x] Change from using requests, to using path parameters - API makes more sense that way.
- [x] Stick the ranges in the db
- [x] Implement explicit fire range endpoint.
- [x] Define all the other endpoints
- [x] Pass range id's around - instead of entire objects!
- [x] Back end tests
- [x] Front end tests
# Test Links:
[Percentile Calculator](https://wps-pr-1789.apps.silver.devops.gov.bc.ca/)
[MoreCast](https://wps-pr-1789.apps.silver.devops.gov.bc.ca/morecast)
[C-Haines](https://wps-pr-1789.apps.silver.devops.gov.bc.ca/c-haines)
[FireBat](https://wps-pr-1789.apps.silver.devops.gov.bc.ca/fire-behaviour-calculator)
[FireBat bookmark](https://wps-pr-1789.apps.silver.devops.gov.bc.ca/fire-behaviour-calculator?s=266&f=c5&c=NaN&w=20,s=286&f=c7&c=NaN&w=16,s=1055&f=c7&c=NaN&w=NaN,s=305&f=c7&c=NaN&w=NaN,s=344&f=c5&c=NaN&w=NaN,s=346&f=c7&c=NaN&w=NaN,s=328&f=c7&c=NaN&w=NaN,s=1399&f=c7&c=NaN&w=NaN,s=334&f=c7&c=NaN&w=NaN,s=1082&f=c3&c=NaN&w=NaN,s=388&f=c7&c=NaN&w=NaN,s=309&f=c7&c=NaN&w=16,s=306&f=c7&c=NaN&w=NaN,s=1029&f=c7&c=NaN&w=NaN,s=298&f=c7&c=NaN&w=NaN,s=1108&f=c5&c=NaN&w=NaN,s=836&f=c7&c=NaN&w=NaN#state=2ec784ca-c46a-49d0-b2b3-1cf32a9015a2&session_state=7d9447c8-db66-4661-b4cb-03d2ac0d1d8f&code=32292df4-2bdf-4f90-a4a8-c8dbcda682a9.7d9447c8-db66-4661-b4cb-03d2ac0d1d8f.2b63f390-f3dc-43ae-89f2-016453863476)
[Fire Behaviour Advisory](https://wps-pr-1789.apps.silver.devops.gov.bc.ca/fire-behaviour-advisory)
[HFI Calculator](https://wps-pr-1789.apps.silver.devops.gov.bc.ca/hfi-calculator)
[FWI Calculator](https://wps-pr-1789.apps.silver.devops.gov.bc.ca/fwi-calculator)


api/app/hfi/hfi_calc.py
```diff
@@ -328,15 +390,18 @@ def calculate_mean_intensity(dailies: List[StationDaily]):
     return math.ceil(mean_intensity_group)
```
andrea-williams, line 390: my spidey senses have been alerted by this, although I could be wrong - we used to have logic on the frontend for rounding MIG values down to x when MIG was <= x.7, and rounding up to x+1 otherwise. I think we've now removed that frontend logic to do all the calculations on the backend, so `math.ceil()` might not be good enough here??

andrea-williams (commented):
Review still in progress, just submitting comments I have to save my progress...

api/app/routers/hfi_calc.py
```diff
@@ -31,8 +34,71 @@
...
 
+def get_prepared_request(
+        session: Session,
+        fire_centre_id: int,
+        start_date: Optional[date],
+        fire_centre_fire_start_ranges: List[FireStartRange]) -> Tuple[HFIResultRequest, bool]:
+    """ Attempt to load the most recent request from the database, failing that creates a new request all
+    set up with default values.
+
+    TODO: request_loaded should become redudant once all requests have been changed.
```
andrea-williams, line 45:
```suggestion
    TODO: request_loaded should become redundant once all requests have been changed.
```

andrea-williams: Just to prove that I'm paying attention...

api/app/routers/hfi_calc.py
```diff
@@ -31,8 +34,71 @@
...
+            logger.error(validation_error)
+    if not request_loaded:
+        # No stored request, so we need to create one.
+        # TODO: selected_station_code_ids make it impossible to have a station selected in one area,
+        # and de-selected in another area. This has to be fixed!
+        selected_station_code_ids = set()
+        planning_area_station_info: Mapping[int, List[StationInfo]] = {}
+        planning_area_fire_starts: Mapping[int, FireStartRange] = {}
+        start_date, end_date = validate_date_range(start_date, None)
+        num_prep_days = (end_date - start_date).days
```
andrea-williams, line 67: won't this lead to an off-by-1 error again? e.g., if a user is requesting a prep period of a single day, end_date = start_date, and `num_prep_days` will equal 0. On the front end we've been adding `+ 1` to the date delta because it's really requesting for the start of the day of `start_date` to the end of the day of `end_date`

api/app/routers/hfi_calc.py
```diff
@@ -69,65 +135,162 @@ def extract_selected_stations(request: HFIResultRequest) -> List[int]:
...
+    planning_area_id: int, station_code: int,
+    enable: bool,
+    response: Response,  # pylint: disable=unused-argument
+    token=Depends(authentication_required)  # pylint: disable=unused-argument
+):
+    """ Enable / disable a station withing a planning area """
+    # TODO: stub - implement!
+    logger.info('/fire_centre/%s/%s/planning_area/%s/station/%s/enable/%s',
+                fire_centre_id, start_date, planning_area_id, station_code, enable)
+    raise NotImplementedError('This function is not yet implemented.')
```
andrea-williams, line 151: Nice. You may have angered the gods of Agile by prematurely partially implementing this function, but I like that it's here to remind us of remaining work and that the endpoint exists already.

api/app/routers/hfi_calc.py
```diff
@@ -140,6 +303,7 @@ async def get_hfi_results(request: HFIResultRequest,
                           response: Response,
                           token=Depends(authentication_required)):
     """ Returns calculated HFI results for the supplied details in the POST body """
+    # THIS FUNCTION IS DEPRECATED.
```
andrea-williams, line 306: any reason for not deleting this function right now?

api/app/routers/hfi_calc.py
```diff
@@ -208,24 +376,41 @@ def get_wfwx_station(wfwx_stations_data: List[WeatherStation], station_code: int
...
+    raise NotImplementedError('Not implemented')
+
+
 @router.post('/download-pdf')
 async def download_result_pdf(request: HFIResultRequest,
                               token=Depends(authentication_required)):
     """ Assembles and returns PDF byte representation of HFI result. """
-
+    # TODO: Replace this method with @router.get('/fire_centre/{fire_centre_id}/{start_date}/pdf')
+    # THIS FUNCTION IS DEPRECATED.
```
andrea-williams, line 396: as above - delete now?

andrea-williams (commented):
Again, submitting to save my progress

web/src/features/hfiCalculator/components/FireStartsDropdown.tsx
```diff
@@ -1,37 +1,47 @@
...
-  setFireStarts: (areaId: number, dayOffset: number, newFireStarts: FireStarts) => void
+  setFireStarts: (
+    areaId: number,
+    dayOffset: number,
+    newFireStarts: FireStartRange
+  ) => void
 }
 
+const useStyles = makeStyles({
+  thing: {
```
andrea-williams, line 20: could we use a more descriptive name than "thing" please?

web/src/features/hfiCalculator/pages/HfiCalculatorPage.tsx
```diff
@@ -128,20 +147,21 @@ const HfiCalculatorPage: React.FunctionComponent = () => {
   const setNewFireStarts = (
     areaId: number,
     dayOffset: number,
-    newFireStarts: FireStarts
+    newFireStarts: FireStartRange
   ) => {
     if (!isUndefined(result)) {
-      const newPlanningAreaFireStarts = cloneDeep(result.planning_area_fire_starts)
-      newPlanningAreaFireStarts[areaId][dayOffset] = { ...newFireStarts }
+      // TODO: Move the smart date higher up
```
andrea-williams, line 153: did you plan to move smart date in a later PR?

web/src/features/hfiCalculator/slices/hfiCalculatorSlice.ts
```diff
@@ -57,21 +58,22 @@ export interface HFICalculatorState {
...
-  planningAreaFireStarts: { [key: string]: FireStarts[] }
+  planningAreaFireStarts: { [key: string]: FireStartRange[] }
   planningAreaHFIResults: { [key: string]: PlanningAreaResult }
   selectedFireCentre: FireCentre | undefined
   result: HFIResultResponse | undefined
   saved: boolean
 }
 
 export interface HFIResultResponse {
+  // TODO: Change from string to SmartDate
```
andrea-williams, line 69: I presume this TODO is for a subsequent PR?

andrea-williams (changes requested):
Found a bug. Selected Cariboo FC, date was then automatically selected. You can see in the screenshot that for some planning areas/date combos, if a fire starts range hasn't already been selected, the default is "0-1", although when I click on the dropdown, there is no "0-1" option (as there shouldn't be).

Screen Shot 2022-03-28 at 12 21 50 PM

api/app/routers/hfi_calc.py
conbrad: +1 please write a tests for this

web/cypress/integration/hfi-calculator-page.spec.ts
```diff
@@ -32,26 +42,27 @@ describe('HFI Calculator Page', () => {
...
     })
     it('save button should be disable', () => {
       // cypress/fixtures/hfi-calc/dailies-saved.json has "request_persist_success": true, save button should be looking at that.
       cy.getByTestId('save-button').should('be.disabled')
     })
-    it('fire start dropdown triggers hfi request', () => {
-      // Selecting a new fire start, should result in a new request to the server, that comes back with "request_persist_success": false, or
-      // no request_save, which should cause the save button to become enabled.
+    it('fire start dropdown triggers hfi set state request', () => {
+      // Selecting a new fire start, should result in a new request to the server, that comes back with "request_persist_success": true, or
```
conbrad, line 54: Please make this comment the test name

conbrad: `it('should send a new request to the server, that persists the change and sets the save button to disabled, when selecting a new fire start ')`

web/src/features/hfiCalculator/pages/HfiCalculatorPage.tsx
```diff
@@ -33,11 +35,24 @@ import { PST_UTC_OFFSET, PST_ISO_TIMEZONE } from 'utils/constants'
 import PrepDaysDropdown from 'features/hfiCalculator/components/PrepDaysDropdown'
 import { FireCentre } from 'api/hfiCalcAPI'
 import { HFIPageSubHeader } from 'features/hfiCalculator/components/HFIPageSubHeader'
-import { cloneDeep, isNull, isUndefined } from 'lodash'
+import { isNull, isUndefined } from 'lodash'
 import HFIErrorAlert from 'features/hfiCalculator/components/HFIErrorAlert'
 import DownloadPDFButton from 'features/hfiCalculator/components/DownloadPDFButton'
 import EmptyFireCentreRow from 'features/hfiCalculator/components/EmptyFireCentre'
 
+function constructPlanningAreaFireStarts(
```
conbrad, line 43: Shouldn't this happen at the API boundary too? In `web/src/api/hfiCalculatorAPI.ts` or in the fetch functions

conbrad (changes requested):

api/alembic/versions/fe33ab8c6c01_fire_start_range.py
```diff
@@ -0,0 +1,61 @@
...
+
+# revision identifiers, used by Alembic.
+revision = 'fe33ab8c6c01'
+down_revision = '9a5bb047ae19'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
```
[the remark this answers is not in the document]
Sybrand: Thanks - I'll keep the comments, just dropping the "please adjust".

api/app/routers/hfi_calc.py
Sybrand: It's technically not deprecated I guess - it's still in use, doesn't have a replacement but will soon.

Sybrand: It will get deleted in a subsequent PR as part of the ticket I'm working on.

api/app/routers/hfi_calc.py
Sybrand: As above - not yet - close.

web/src/features/hfiCalculator/pages/HfiCalculatorPage.tsx
Sybrand: yes/no - depend on what feedback I get on the smartdate. Yes if we keep it, not if we don't.

web/src/features/hfiCalculator/slices/hfiCalculatorSlice.ts
Sybrand: Yep - again depending on the feedback I get on smartdate.

JulianForeman (approved):

web/src/features/hfiCalculator/pages/HfiCalculatorPage.tsx
conbrad: Please make smart date a separate PR unless it's not possible.

api/app/hfi/hfi_calc.py
```diff
@@ -225,15 +228,85 @@ async def calculate_latest_hfi_results(request: HFIResultRequest):
...
+    daily_results: List[DailyResult] = []
+    for index in range(num_prep_days):
+        dailies_date = start_date + timedelta(days=index)
+        prep_day_dailies = get_prep_day_dailies(dailies_date, area_dailies)
+        daily_fire_starts: FireStartRange = planning_area_fire_starts[area_id][index]
+        mean_intensity_group = calculate_mean_intensity(prep_day_dailies)
+        prep_level = calculate_prep_level(mean_intensity_group, daily_fire_starts, fire_start_lookup)
+        validated_dailies: List[ValidatedStationDaily] = list(
+            map(validate_station_daily, prep_day_dailies))
+        all_dailies_valid = all(map(lambda validated_daily: (
```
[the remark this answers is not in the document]
Sybrand: I'm not a big fan either - I'll attempt. I find it really hard to read high order functions that are strung up like that - but it's deliciously concise. (b.t.w. i didn't write that! that's @conbrad signature - he's the master of high order functions)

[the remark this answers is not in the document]
conbrad: I highly recommend getting comfortable with functional programming, it helps with reasoning about code correctness (purity through referential integrity). I wouldn't write it this way if it didn't help making code correct and it was just a style thing. 

Check out this for more motivation: https://github.com/97-things/97-things-every-programmer-should-know/blob/master/en/thing_02/README.md

api/app/hfi/hfi_calc.py
Sybrand: I think that's exactly what's being done here? `if >= x.8` then `math.ceil(x)` - which `==` math.flor(x)+1`

PS. Not sure why this is being caught now? I didn't touch this code. This is code you wrote.

api/app/hfi/hfi_calc.py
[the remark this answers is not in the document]
Sybrand: Totally. Not an attack on @conbrad - I feel it's one of my weak points - I struggle reading and writing code like that.

api/app/hfi/hfi_calc.py
andrea-williams: well if I wrote it, then it must be perfect! /s 
my concern is that `math.ceil(x.7)` will be rounded up to `x+1`, when it should be left at `x`. I'll double-check the logic in your dev env later

api/app/hfi/hfi_calc.py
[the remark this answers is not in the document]
conbrad: All good, it's not easy to read if you're unfamiliar, just wanted to clear up the reason why I chose to write it this way in the first place.

web/src/features/hfiCalculator/pages/HfiCalculatorPage.tsx
Sybrand: Oh totally - this PR is too monstrous anyway!

api/alembic/versions/c525dbd0c37e_populate_fire_start_range.py
```diff
@@ -0,0 +1,138 @@
...
+                                             column('fire_centre_id', Integer),
+                                             column('order', Integer))
+
+    # Link fire start ranges to each fire centre.
+    fire_centre_ranges = []
+    conn = op.get_bind()
+    res = conn.execute('SELECT id, name FROM fire_centres')
+    results = res.fetchall()
+
+    id_count = 1
```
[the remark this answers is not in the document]
Sybrand: It's not needed for a single insert - will be autogenerated - but I need the id for foreign keys - and assigning it myself is just the easiest way to achieve that.

api/app/routers/hfi_calc.py
Sybrand: Writing a test for this particular edge case is actually a nightmare - there's a function initialize_planning_area_fire_starts, which consumes the number of prep days and ensures the structure is correct. That function is run twice, SOMETIMES - which is just redundant and bad (there's a TODO to that effect).
I hope to re-factor things to address this - right now - reproducing the edge case where this value get's used is too much effort. Instead - I've ripped out the math, and moved it into the schema, so you can do `date_range.num_prep_days()` which ensure the math at least should remain consistent.

conbrad: Nice the date range function should be easy to test then. Which commit was that change made?

Sybrand: Just pushed it up now.

web/src/features/hfiCalculator/pages/HfiCalculatorPage.tsx
Sybrand: This function should be thrown away when fetchHFIResult and fetchPDFDownload get thrown away, so not worried about it right now. Will add a comment to that effect though.

api/app/hfi/hfi_calc.py
```diff
@@ -225,15 +228,85 @@ async def calculate_latest_hfi_results(request: HFIResultRequest):
...
                                         area_station_map,
                                         valid_start_date)
-        return results, start_timestamp, end_timestamp
-
-
-def calculate_hfi_results(planning_area_fire_starts: Mapping[int, FireStartRange],  # pylint: disable=too-many-locals
+        return results, valid_start_date, valid_end_date
+
+
+def build_fire_start_prep_level_lookup(orm_session) -> Dict[int, Dict[int, int]]:
```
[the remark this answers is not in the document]
Sybrand: I've added that as acceptance criteria for story - will do in it's own pr.

api/app/routers/hfi_calc.py
```diff
@@ -31,8 +34,71 @@
 )
 
 
+def get_prepared_request(
```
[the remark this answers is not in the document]
Sybrand: I've added a TODO on that function. It's a mess right now - as there are many ways to get a request. Once I've done the rest - a simple "get_request" will suffice. (It's not necessarily most recent, depends on what's in the database'); Calling it "get_request" right now would just be confusing!

web/cypress/integration/hfi-calculator-page.spec.ts
conbrad: @Sybrand , not addressed?

Sybrand: Weird that it doesn't show as Outdated? I've pushed the change.

api/app/schemas/hfi_calc.py
```diff
@@ -149,12 +139,12 @@ class DateRange(BaseModel):
     start_date: Optional[date]
     end_date: Optional[date]
 
-
-class HFILoadResultRequest(BaseModel):
-    """ Request to load the HFI Calculator. """
-    start_date: Optional[date]
-    end_date: Optional[date]
-    selected_fire_center_id: int
+    def num_days(self) -> Optional[int]:
```
conbrad, line 142: Tests?

conbrad (changes requested):

web/src/utils/date.ts
```diff
@@ -81,3 +81,52 @@ export const getDaysBetween = (startDate: string, endDate: string): DateTime[] =
 
   return dates
 }
+
+export class SmartDate {
```
conbrad, line 85: Is this used anywhere? Why is it in this PR?

conbrad (changes requested):

api/app/schemas/hfi_calc.py
Sybrand: It's covered in higher level tests - you want a unit test specifically for that?

conbrad: I would like unit tests for the +1 on end date yes

web/src/utils/date.ts
Sybrand: Oh sweet - the references were removed by the merge. I'll kill it.

Sybrand: web/src/features/hfiCalculator/pages/HfiCalculatorPage.tsx : 120

I have plans for this class.

conbrad: Can you please write tests for it?

conbrad: Or move it to it's own PR if it's only used in one place, this PR is already huge.

api/app/tests/hfi/test_hfi.py
```diff
@@ -260,3 +260,18 @@ def test_valid_date_range_default_for_end_date_before():
     result = validate_date_range(DateRange(start_date=start_date, end_date=end_date))
     assert result.start_date.isoformat() == '2020-05-21'
     assert result.end_date.isoformat() == '2020-05-21'
+
+
+def test_inclusive_date_math():
```
conbrad, line 265: Can you please also add a test for `startDate` and `endDate` are on the same day (single day prep cycle) and that `days_in_range()` returns 1?

conbrad (changes requested):

web/src/utils/date.ts
Sybrand: Pulled out.

conbrad: @Sybrand it's still here but not used?

Sybrand: I was just about to blame github when I noticed I hadn't pressed save!

Sybrand: Push commit. I blame the bee sting.

api/app/hfi/hfi_calc.py
```diff
@@ -202,18 +206,84 @@ async def calculate_latest_hfi_results(request: HFIResultRequest):
...
+                                        fire_start_lookup,
+                                        dailies, valid_date_range.days_in_range(),
                                         request.selected_station_code_ids,
                                         area_station_map,
                                         valid_date_range.start_date)
         return results, valid_date_range
 
 
-def calculate_hfi_results(planning_area_fire_starts: Mapping[int, FireStartRange],  # pylint: disable=too-many-locals
+def build_fire_start_prep_level_lookup(orm_session) -> Dict[int, Dict[int, int]]:
```
conbrad, line 219: What's the reason behind introducing the `Dict` type here instead of the existing `Mapping` type?

conbrad (approved):
Approving assuming tests pass and dev deployment is good

andrea-williams (approved):
I'm happy (or as happy as I'll ever be). Not reading through all the code again - it was painful enough the first time. Glad to know that the "0-1 when it should be 0-2" issue is only an edge case.

api/app/hfi/hfi_calc.py
Sybrand: I am SOOOOOOO excited you asked. Short answer: because it's returning a dict, not a map. If it were Mapping, it would inherit from collection, but {} or dict() doesn't.

If you peak at the definitions in: /usr/lib/python3.8/typing.py, you'll see
```
Dict = _alias(dict, (KT, VT), inst=False)
Mapping = _alias(collections.abc.Mapping, (KT, VT_co))
```

conbrad: Cool thanks. Should it return a map though? Isn't a `Mapping` an immutable collection? That's what this should be when constructed, yeah?

Sybrand: Oh hell yeah! Super valid comment. Yes! It should totally be that! I'll make that change - and then I'd like to merge this in.

Sybrand: We'd have to go to a bunch of effort to make an immutable object to represent the prep level lookup. It's a fair point that the intention is that the result be immutable - but that's a lot of work.

PR #1792, opened 2022-03-16
Merge daily and prep pdf generation

conbrad:
This PR merges the PDF templates as is. There is still missing data, formatting, etc, that will be addressed as a follow up PR. This one just gets us to a merged PDF state.

# Test Links:
[Percentile Calculator](https://wps-pr-1792.apps.silver.devops.gov.bc.ca/)
[MoreCast](https://wps-pr-1792.apps.silver.devops.gov.bc.ca/morecast)
[C-Haines](https://wps-pr-1792.apps.silver.devops.gov.bc.ca/c-haines)
[FireBat](https://wps-pr-1792.apps.silver.devops.gov.bc.ca/fire-behaviour-calculator)
[FireBat bookmark](https://wps-pr-1792.apps.silver.devops.gov.bc.ca/fire-behaviour-calculator?s=266&f=c5&c=NaN&w=20,s=286&f=c7&c=NaN&w=16,s=1055&f=c7&c=NaN&w=NaN,s=305&f=c7&c=NaN&w=NaN,s=344&f=c5&c=NaN&w=NaN,s=346&f=c7&c=NaN&w=NaN,s=328&f=c7&c=NaN&w=NaN,s=1399&f=c7&c=NaN&w=NaN,s=334&f=c7&c=NaN&w=NaN,s=1082&f=c3&c=NaN&w=NaN,s=388&f=c7&c=NaN&w=NaN,s=309&f=c7&c=NaN&w=16,s=306&f=c7&c=NaN&w=NaN,s=1029&f=c7&c=NaN&w=NaN,s=298&f=c7&c=NaN&w=NaN,s=1108&f=c5&c=NaN&w=NaN,s=836&f=c7&c=NaN&w=NaN#state=2ec784ca-c46a-49d0-b2b3-1cf32a9015a2&session_state=7d9447c8-db66-4661-b4cb-03d2ac0d1d8f&code=32292df4-2bdf-4f90-a4a8-c8dbcda682a9.7d9447c8-db66-4661-b4cb-03d2ac0d1d8f.2b63f390-f3dc-43ae-89f2-016453863476)
[Fire Behaviour Advisory](https://wps-pr-1792.apps.silver.devops.gov.bc.ca/fire-behaviour-advisory)
[HFI Calculator](https://wps-pr-1792.apps.silver.devops.gov.bc.ca/hfi-calculator)
[FWI Calculator](https://wps-pr-1792.apps.silver.devops.gov.bc.ca/fwi-calculator)


andrea-williams (commented):
Is the Download PDF button meant to do anything? Because it doesn't appear to be...

conbrad:
> Is the Download PDF button meant to do anything? Because it doesn't appear to be...

It should, thanks, just pushed a fix in: https://github.com/bcgov/wps/pull/1792/commits/f8a36c8c57bb01ef3fe1ac6d57283b96f383370c

andrea-williams (approved):

api/app/hfi/pdf_generator.py
```diff
@@ -3,16 +3,42 @@
...
 from app.schemas.hfi_calc import FireCentre, HFIResultResponse, PlanningArea, WeatherStation
-from app.hfi.daily_template import str_daily_template, CSS_PATH
-from app.hfi.pdf_data_formatter import response_2_daily_jinja_format
+from app.hfi.pdf_template import PDFTemplateName, get_template, CSS_PATH
+from app.hfi.pdf_data_formatter import response_2_daily_jinja_format, response_2_prep_cycle_jinja_format
 
 # Loads template as string from a function
 # See: https://jinja.palletsprojects.com/en/3.0.x/api/?highlight=functionloader#jinja2.FunctionLoader
-jinja_env = Environment(loader=FunctionLoader(str_daily_template), autoescape=True)
+jinja_env = Environment(loader=FunctionLoader(get_template), autoescape=True)
```
Sybrand, line 11: Does this have to be global scope?

api/app/hfi/pdf_generator.py
```diff
@@ -3,16 +3,42 @@
...
 
 
-def generate_daily_pdf(result: HFIResultResponse, fire_centres: List[FireCentre]) -> bytes:
-    """Generate a daily PDF"""
+def generate_pdf(result: HFIResultResponse, fire_centres: List[FireCentre]) -> bytes:
+    """Generates the full PDF based on the HFIResultResponse"""
+
+    rendered_output: str = ''
```
Sybrand, line 17: Isn't this line redundant? Could you just start with
`rendered_output = generate_prep(result)`

api/app/hfi/pdf_generator.py
```diff
@@ -3,16 +3,42 @@
...
+    rendered_output += generate_prep(result)
```
Sybrand, line 18: Using string concatenation on big strings could start getting slow - but if you're not experiencing that this is fine.

api/app/hfi/pdf_template.py
```diff
@@ -1,8 +1,16 @@
...
 CSS_PATH = os.path.join(os.path.dirname(__file__), "style.css")
 
+
+class PDFTemplateName(Enum):
+    """Enumerates the possible templates client can ask for"""
+    DAILY = 'daily'
+    PREP = 'prep'
+
+
 DAILY_TEMPLATE_HTML = """
```
Sybrand, line 14: I'm not too keen on inline html - can't we load this from a file? Just read it in as a string.

Sybrand (changes requested):
- I'm seeing lots of code that isn't covered with tests, I'm ok with that not being addressed right now, but pls. keep in mind going forward.
- The inline HTML is a bit gross - can we please extract that into an HTML file and read it at runtime?

api/app/hfi/pdf_generator.py
conbrad: Not experiencing it, also just a single concatenation per PDF endpoint call. Does Python have the equivalent of Java's `StringBuilder`?

api/app/hfi/pdf_template.py
conbrad: Was trying to avoid loading HTML files every call, but I think it's a decent trade off in favor of readability to do so

api/app/hfi/pdf_generator.py
conbrad: No, I kept it in module global scope since it's not mutated and only read from. Do you have a specific way you'd like it scoped? 

I can think of 2 ways:
1. Pass in the `jinja_env` from the called, so this would be scoped to the router function
2. Make these module functions a class and have the env instantiated in an instance

Sybrand: I just don't like global scope as it unnecessarily loads on module import. If it loads quickly, I'd rather have it load when needed instead of adding to the memory burden.

api/app/hfi/pdf_template.py
Sybrand: It only happens on pdf rendering, it shouldn't make it too slow. If you're worried, you can always stick it int that singleton class.

conbrad:
> * I'm seeing lots of code that isn't covered with tests, I'm ok with that not being addressed right now, but pls. keep in mind going forward.
> * The inline HTML is a bit gross - can we please extract that into an HTML file and read it at runtime?

Addressed all the individual comments below and will be adding tests in the next PR where I'll be fixing the formatting and data displayed in the prep cycle table.

Sybrand (approved):
Just a comment re. possible redundant # pylint: disable=unused-argument

api/app/hfi/pdf_template.py
```diff
@@ -0,0 +1,25 @@
...
+CSS_PATH = os.path.join(os.path.dirname(__file__), "style.css")
+
+
+class PDFTemplateName(Enum):
+    """Enumerates the possible templates client can ask for"""
+    DAILY = 'daily'
+    PREP = 'prep'
+
+
+def get_template(template_name: str):    # pylint: disable=unused-argument
```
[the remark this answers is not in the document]
conbrad: Ah it's stake, will remove, thanks

PR #1858, opened 2022-04-05
Bump width for prep range picker

conbrad:
'September' date range text does not fit the text field for prep date range picker. Bump width (matches width of fire centre dropdown too)

Before:
Screen Shot 2022-04-05 at 4 10 33 PM

After:
Screen Shot 2022-04-05 at 4 11 06 PM


tessmgood (approved):
Excellent catch! TY

andrea-williams (approved):

PR #1883, opened 2022-04-11
Set fire starts role

conbrad:
- Backend: requires `hfi_set_fire_starts` role for `POST`ing changes fire starts endpoint, returns 401 if role isn't found
- Frontend: Updated redux store to include `shouldEnableFireStarts` when authenticating. This is subscribed to and passed to `FireStartsDropdown` to determine whether to enable or just show the fire starts value.

# Test Links:
[Percentile Calculator](https://wps-pr-1883.apps.silver.devops.gov.bc.ca/)
[MoreCast](https://wps-pr-1883.apps.silver.devops.gov.bc.ca/morecast)
[C-Haines](https://wps-pr-1883.apps.silver.devops.gov.bc.ca/c-haines)
[FireBat](https://wps-pr-1883.apps.silver.devops.gov.bc.ca/fire-behaviour-calculator)
[FireBat bookmark](https://wps-pr-1883.apps.silver.devops.gov.bc.ca/fire-behaviour-calculator?s=266&f=c5&c=NaN&w=20,s=286&f=c7&c=NaN&w=16,s=1055&f=c7&c=NaN&w=NaN,s=305&f=c7&c=NaN&w=NaN,s=344&f=c5&c=NaN&w=NaN,s=346&f=c7&c=NaN&w=NaN,s=328&f=c7&c=NaN&w=NaN,s=1399&f=c7&c=NaN&w=NaN,s=334&f=c7&c=NaN&w=NaN,s=1082&f=c3&c=NaN&w=NaN,s=388&f=c7&c=NaN&w=NaN,s=309&f=c7&c=NaN&w=16,s=306&f=c7&c=NaN&w=NaN,s=1029&f=c7&c=NaN&w=NaN,s=298&f=c7&c=NaN&w=NaN,s=1108&f=c5&c=NaN&w=NaN,s=836&f=c7&c=NaN&w=NaN#state=2ec784ca-c46a-49d0-b2b3-1cf32a9015a2&session_state=7d9447c8-db66-4661-b4cb-03d2ac0d1d8f&code=32292df4-2bdf-4f90-a4a8-c8dbcda682a9.7d9447c8-db66-4661-b4cb-03d2ac0d1d8f.2b63f390-f3dc-43ae-89f2-016453863476)
[Fire Behaviour Advisory](https://wps-pr-1883.apps.silver.devops.gov.bc.ca/fire-behaviour-advisory)
[HFI Calculator](https://wps-pr-1883.apps.silver.devops.gov.bc.ca/hfi-calculator)
[FWI Calculator](https://wps-pr-1883.apps.silver.devops.gov.bc.ca/fwi-calculator)


api/app/auth.py
```diff
@@ -55,3 +55,18 @@ async def authentication_required(token=Depends(authenticate)):
...
         )
     return token
+
+
+async def auth_with_set_fire_starts_permission_required(token=Depends(authentication_required)):
+    """ Only return requests that have set fire starts permission """
+    try:
+        roles = token.get('resource_access', None).get('wps-web', None).get('roles', None)
+        if 'hfi_set_fire_starts' not in roles:
+            raise HTTPException
```
Sybrand, line 65: I think we should cover this edge case in our unit tests somewhere (request to change fire starts without permission).

web/src/features/auth/slices/authenticationSlice.ts
```diff
@@ -86,10 +93,21 @@ export const decodeRoles = (token: string | undefined) => {
...
+  dispatch(authenticateFinished({ isAuthenticated: true, token }))
+}
+
 export const authenticate = (): AppThunk => dispatch => {
   dispatch(authenticateStart())
 
   if (!kcInstance) {
+    console.log('Hey no auth')
```
Sybrand, line 110: console.log ?

web/src/features/auth/slices/authenticationSlice.ts
```diff
@@ -12,7 +12,7 @@ interface State {
   isAuthenticated: boolean
   tokenRefreshed: boolean
   token: string | undefined
-  roles: string[]
+  shouldEnableFireStarts: boolean
```
Sybrand, line 15: I'm worried this is too specific on this level. Aren't we better off having `roles: string[]` ? When we start adding in other roles, for other products, it could get very confusing. e.g. if you're busy working on MoreCast, what relevance does `shouldEnableFireStarts` have on the global authentication slice?

api/app/tests/common.py
```diff
@@ -91,7 +117,7 @@ class DefaultMockAioSession:
     """ Mock aiobotocore.session.AioSession """
     # pylint: disable=unused-argument, no-self-use
 
-    @asynccontextmanager
+    @ asynccontextmanager
```
andrea-williams, line 120: extra space snuck in here - I don't know if it matters for anything tho...

web/src/features/auth/slices/authenticationSlice.ts
andrea-williams, line 110: Do you still want this in the console in prod?

api/app/tests/hfi/test_hfi_endpoint_request.feature
```diff
@@ -10,27 +10,48 @@ Feature: /hfi/
...
+        And request == saved = <request_saved>
 
+        Examples:
+            | url                                                                                              | verb | status_code | response_json                                            | request_saved | stored_request_json                   |
+            | /api/hfi-calc/fire_centre/1/2020-05-21/planning_area/1/fire_starts/2020-05-21/fire_start_range/2 | post | 200         | hfi/test_hfi_endpoint_response_set_fire_start_range.json | True          | None                                  |
+            | /api/hfi-calc/fire_centre/1/2020-05-21/planning_area/1/fire_starts/2020-05-21/fire_start_range/2 | post | 200         | hfi/test_hfi_endpoint_response_set_fire_start_range.json | True          | test_hfi_endpoint_stored_request.json |
 
+    Scenario: HFI - request set fire starts without role
+        Given I have a stored request <stored_request_json>
+        And I spy on store_hfi_request
```
Sybrand, line 46: I think you'd be better off having a `Given request with roles: <roles>` statement. That way you don't need multiple scenarios, and you can specify the roles you want as input.
If my comment doesn't make sense, please reach out.

Sybrand (changes requested):

web/src/features/auth/slices/authenticationSlice.ts
```diff
@@ -86,10 +93,21 @@ export const decodeRoles = (token: string | undefined) => {
   }
 }
 
+export const fireStartsEnabled = (isAuthenticated: boolean, roles: string[]) => {
+  return isAuthenticated && roles.includes('hfi_set_fire_starts')
+}
+
+export const cypressAuthFinished = (): AppThunk => dispatch => {
```
andrea-williams, line 100: I'm assuming this is only ever called when running Cypress tests, in which case I'd argue that this should go somewhere other than this app-wide file, and instead into a Cypress-specific file

conbrad: It can't, it needs access to the authenticateFinished function and I don't want to expose that elsewhere

api/app/tests/hfi/test_hfi_endpoint_request.feature
conbrad: It's difficult to do with the specific JWT mock I need, not sure if there's something easier I'm missing

web/src/features/auth/slices/authenticationSlice.ts
conbrad: I'm worried that's too general! I say we expose only what's needed. This also removes the concern you had in the last PR around if `isAuthenticated` is true or not.

api/app/tests/hfi/test_hfi_endpoint_request.feature
Sybrand: Oh yeah - you can get that  mock in place without doing a decorator on a function - I can explain better with a screen share - reach out.

web/src/features/auth/slices/authenticationSlice.ts
Sybrand: Soooooo.... I think there are some things we can do here. I think it's more about what we're naming this. We're saying "using this hard coded token when we're not doing real authentication".  It's something that one could in theory use beyond just cypress. If I wanted to do local development, without keycloak (say keycloak is down, or something) - then toggling authentication is useful. Let's call it something else.

The other thing to consider is of the token should be in the code, or if we can read it from the environment. I think we could just make it an environment variable.

web/src/features/auth/slices/authenticationSlice.ts
Sybrand: shouldEnableFireStarts relates only to HFI though, and that's not clear looking at this.

conbrad: Good point, I was thinking of throwing these in a `permissions` object within the store. That way it's namespaced.

web/src/features/auth/slices/authenticationSlice.ts
conbrad: I added `TEST_AUTH` in env (true, false or missing) for this. `AuthWrapper` checks for that or `window.Cypress` since env variables in Cypress are a pain.

api/app/tests/hfi/test_hfi_endpoint_request.py
```diff
@@ -44,9 +44,53 @@ def mock_get_fire_weather_stations(_):
...
+
+
+@pytest.mark.usefixtures('mock_jwt_decode')
+@scenario('test_hfi_endpoint_request.feature', 'HFI - GET request')
+def test_fire_behaviour_calculator_get_scenario():
+    """ BDD Scenario. """
+    pass
+
+
 @pytest.mark.usefixtures('mock_jwt_decode')
```
Sybrand, line 91: Can't can drop @pytest.mark.usefixtures('mock_jwt_decode') since we're patching it in the scenario?

api/app/tests/hfi/test_unauthorized.json
```diff
@@ -0,0 +1,3 @@
+{
```
Sybrand, line 1: I think we can drop this file?

api/app/tests/hfi/test_hfi_endpoint_request.feature
```diff
@@ -1,28 +1,47 @@
...
+            # Test set fire start range with correct role
+            | /api/hfi-calc/fire_centre/1/2020-05-21/2020-05-25/planning_area/1/fire_starts/2020-05-21/fire_start_range/2 | hfi_set_fire_starts | 200         | hfi/test_hfi_endpoint_response_set_fire_start_range.json | True          | None                                  |
+            | /api/hfi-calc/fire_centre/1/2020-05-21/2020-05-25/planning_area/1/fire_starts/2020-05-21/fire_start_range/2 | hfi_set_fire_starts | 200         | hfi/test_hfi_endpoint_response_set_fire_start_range.json | True          | test_hfi_endpoint_stored_request.json |
+            # Test set fire start range without roles
+            | /api/hfi-calc/fire_centre/1/2020-05-21/2020-05-25/planning_area/1/fire_starts/2020-05-21/fire_start_range/2 | None                | 401         | hfi/test_unauthorized.json                               | False         | None                                  |
```
Sybrand, line 38: I don't think we need to have this file. Can't we rather set it to None, and just not bother comparing it. If there's a 401 I think it's sufficient. I don't think we need to interrogate the contents of the response beyond that?

web/src/features/auth/slices/authenticationSlice.ts
```diff
@@ -48,6 +48,7 @@ const authSlice = createSlice({
       state.authenticating = false
       state.isAuthenticated = false
       state.error = action.payload
+      state.roles = []
```
Sybrand, line 51: Can we test this file? It doesn't get pulled in as part of cypress, but maybe we can do a normal boring unit test on it? (Either that - or we should just throw it into codecov ignore).
Authentication and role extraction seems pretty import though?

Sybrand (changes requested):
A few minor requests / questions - but I think it's looking good.

api/app/tests/hfi/test_hfi_endpoint_request.py
```diff
@@ -44,9 +44,53 @@ def mock_get_fire_weather_stations(_):
...
 
+def _setup_mock_with_role(monkeypatch: pytest.MonkeyPatch, role: str):
+    """ Prepare jwt decode to be mocked with permission
+    """
+    _setup_mock(monkeypatch)
+
+    class MockJWTDecodeWithRole:
+        """ Mock pyjwt module with role """
+
+        def __init__(self, role):
```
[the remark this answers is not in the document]
conbrad: It's the same variable since, it's just in a closure.

api/app/tests/hfi/test_hfi_endpoint_request.py
conbrad: I still need it here for the station select tests that have no permissions yet

api/app/tests/hfi/test_unauthorized.json
conbrad: Tests fail without it.

web/src/features/auth/slices/authenticationSlice.ts
conbrad: Done in: https://github.com/bcgov/wps/pull/1883/commits/78ae303d7fa1721bc40c078f86835a4f5f2626dd

andrea-williams (commented):
Code LGTM.

andrea-williams (approved):

conbrad:
Closing to reopen as a PR against `roles`

PR #1889, opened 2022-04-13
Story/fire starts role/1869

conbrad:
- Adds backend logic to 401 any request from a token that does not have `hfi_set_fire_starts` role assigned
- Adds frontend logic to disable fire starts dropdowns if token does not have `hfi_set_fire_starts` role assigned
- Simplifies `AuthWrapper` component (removes `shouldAuthenticate` boolean, just checks for `window.Cypress or TEST_AUTH env variable)
- Adds tests around authentication redux store


conbrad:
Closing to reopen against roles with dev pipelines enabled.

PR #1919, opened 2022-04-19
Story/add kamloops stations/1849

conbrad:
- [x] Given I've navigated to Kamloops Fire Centre in HFI Calc, When I look at the Penticton planning area, Then I can see the West Kelowna - 1277 station second in the planning area list, with fuel type C7
- [x] Given I've navigated to Kamloops Fire Centre in HFI Calc, When I look at the Vernon planning area, Then I can see the "Station Bay 2 - 1359" station third in the planning area list, with fuel type C5


andrea-williams (changes requested):
Station Bay 2 is showing up in the Penticton zone, but it should be in the Vernon zone.

conbrad:
Will re-open based on feedback that will be easier to migrate in a new PR

PR #1973, opened 2022-05-02
Story/hfi admin add station backend/1931

conbrad:
- [x] Given I am authenticated as a PSU administrator, Then I can add station
- [x] Given I am authenticated as a PSU administrator, Then the planning areas are filtered based on the selected fire centre
- [x] Given I have specified planning area, station code, default fuel type, When I confirm my selection, then the station is added to our database and reflected in HFI
- [x] Given a failure happens during station submission, we notify the user of the error

# Test Links:
[Percentile Calculator](https://wps-pr-1973.apps.silver.devops.gov.bc.ca/)
[MoreCast](https://wps-pr-1973.apps.silver.devops.gov.bc.ca/morecast)
[C-Haines](https://wps-pr-1973.apps.silver.devops.gov.bc.ca/c-haines)
[FireBat](https://wps-pr-1973.apps.silver.devops.gov.bc.ca/fire-behaviour-calculator)
[FireBat bookmark](https://wps-pr-1973.apps.silver.devops.gov.bc.ca/fire-behaviour-calculator?s=266&f=c5&c=NaN&w=20,s=286&f=c7&c=NaN&w=16,s=1055&f=c7&c=NaN&w=NaN,s=305&f=c7&c=NaN&w=NaN,s=344&f=c5&c=NaN&w=NaN,s=346&f=c7&c=NaN&w=NaN,s=328&f=c7&c=NaN&w=NaN,s=1399&f=c7&c=NaN&w=NaN,s=334&f=c7&c=NaN&w=NaN,s=1082&f=c3&c=NaN&w=NaN,s=388&f=c7&c=NaN&w=NaN,s=309&f=c7&c=NaN&w=16,s=306&f=c7&c=NaN&w=NaN,s=1029&f=c7&c=NaN&w=NaN,s=298&f=c7&c=NaN&w=NaN,s=836&f=c7&c=NaN&w=NaN)
[Fire Behaviour Advisory](https://wps-pr-1973.apps.silver.devops.gov.bc.ca/fire-behaviour-advisory)
[HFI Calculator](https://wps-pr-1973.apps.silver.devops.gov.bc.ca/hfi-calculator)
[FWI Calculator](https://wps-pr-1973.apps.silver.devops.gov.bc.ca/fwi-calculator)


api/app/routers/hfi_calc.py
```diff
@@ -361,6 +368,42 @@ async def get_fire_centres(response: Response):
         raise
 
 
+@router.get('/add-station/{fire_centre_id}', response_model=HFIAddStationOptionsResponse)
+async def get_add_station_options(fire_centre_id: int, response: Response, _=Depends(authentication_required)):
+    """ Returns lists of planning areas, stations and fuel types for adding a station. """
+
+    try:
```
Sybrand, line 375: Don't need to wrap it in try catch anymore - the middleware will log this.

api/app/routers/hfi_calc.py
```diff
@@ -361,6 +368,42 @@ async def get_fire_centres(response: Response):
         raise
 
 
+@router.get('/add-station/{fire_centre_id}', response_model=HFIAddStationOptionsResponse)
```
Sybrand, line 371: I think this endpoint is redundant - since we already have well tested endpoints that provide all this information.

api/app/routers/hfi_calc.py
```diff
@@ -361,6 +368,42 @@ async def get_fire_centres(response: Response):
...
+async def get_add_station_options(fire_centre_id: int, response: Response, _=Depends(authentication_required)):
+    """ Returns lists of planning areas, stations and fuel types for adding a station. """
+
+    try:
+        logger.info('/hfi-calc/add-station')
+        # we can safely cache the options, as they don't change them very often.
+        response.headers["Cache-Control"] = "max-age=86400"
```
Sybrand, line 378: I don't actually think we can cache this - if we need this endpoint, it's because station information is changing. (also - keep in mind when writing the post - it may be a good idea to invalidate the cache - since changing stations could mean that we have a stale station lits)

api/app/routers/hfi_calc.py
```diff
@@ -361,6 +368,42 @@ async def get_fire_centres(response: Response):
...
+            stations: List[BasicWFWXStation] = []
+            for wfwx_station in wfwx_stations:
+                stations.append(BasicWFWXStation(wfwx_station_uuid=wfwx_station.wfwx_id,
+                                code=wfwx_station.code, name=wfwx_station.name))
+
+            return HFIAddStationOptionsResponse(planning_areas=planning_areas,
+                                                stations=stations,
+                                                fuel_types=fuel_types)
+
+    except Exception as exc:
```
Sybrand, line 402: As above - I don't think we need this, our middleware will log the exception.

api/app/wildfire_one/wfwx_api.py
```diff
@@ -291,6 +291,15 @@ async def get_wfwx_stations_from_station_codes(
     return requested_stations
 
 
+async def get_wfwx_all_active_stations(
+        session: ClientSession,
+        header) -> List[WFWXWeatherStation]:
+    """ Return the WFWX station ids and codes that are active from WFWX API. """
```
Sybrand, line 297: Do we have to filter on wildfire stations only? I'm guessing we can rely on the person adding the stations to know what the correct station is. But might be worth checking with @tessmgood ?

api/app/routers/hfi_calc.py
Sybrand, line 371: `/admin/add-station/{fire_centre_id}` ? or `/admin/fire-centre/{fire_centre_id}` ?

api/app/routers/hfi_calc.py
```diff
@@ -361,6 +368,42 @@ async def get_fire_centres(response: Response):
         raise
 
 
+@router.get('/add-station/{fire_centre_id}', response_model=HFIAddStationOptionsResponse)
+async def get_add_station_options(fire_centre_id: int, response: Response, _=Depends(authentication_required)):
```
Sybrand, line 372: Station admin role check? (I guess this is on hold in the other branch?)

Sybrand (commented):
I think it's a better compromise to use our existing less than ideal for this particular use case endpoints.

api/app/wildfire_one/wfwx_api.py
conbrad: What other stations are there besides wildfire stations?

Sybrand: wf1 api propogates weather info for more than just stations managed by bcws. e.g. crd weather stations. Again, I don't know that it matters, since presumable the user knows what they are doing.

api/app/auth.py
```diff
@@ -76,3 +76,8 @@ async def auth_with_set_fire_starts_role_required(token=Depends(authentication_r
 async def auth_with_select_station_role_required(token=Depends(authentication_required)):
     """ Only return requests that have set fire starts permission """
     return await check_token_for_role('hfi_select_station', token)
+
+
+async def auth_with_admin_role_required(token=Depends(authentication_required)):
```
Sybrand, line 81:
```suggestion
async def auth_with_station_admin_role_required(token=Depends(authentication_required)):
```
Maybe more explicit?

api/app/db/models/hfi_calc.py
```diff
@@ -69,8 +69,7 @@ class PlanningWeatherStation(Base):
         {'comment': 'Identifies the unique code used to identify the station'}
     )
 
-    id = Column(Integer, Sequence('planning_weather_station_id_seq'),
-                primary_key=True, nullable=False, index=True)
+    id = Column(Integer, primary_key=True)
```
Sybrand, line 72: I think losing the named sequence is problematic. Have you checked if alembic is going to try to generate a migration based off this change? Why are we losing the index here?

api/app/routers/hfi_calc.py
```diff
@@ -352,7 +357,7 @@ async def get_fire_centres(response: Response):
         logger.info('/hfi-calc/fire-centres')
         # we can safely cache the fire centres, as they don't change them very often.
         # the eco-division logic is very slow, and chomps up 2 seconds!
-        response.headers["Cache-Control"] = "max-age=86400"
+        response.headers["Cache-Control"] = "max-age=0"
```
Sybrand, line 360: We need to have some sort of caching here - that eco-division thing really messes up the load. If we're going to lose browser caching, then I think we need to switch to using redis to cache the response. We can then invalidate the cache entry when we do a post that changes the fire centres.

api/app/wildfire_one/wfwx_api.py
```diff
@@ -291,6 +291,15 @@ async def get_wfwx_stations_from_station_codes(
     return requested_stations
 
 
+async def get_wfwx_all_active_stations(
```
Sybrand, line 294: You've removed the code that uses this - so I think we can drop this function now?

web/src/features/hfiCalculator/components/stationAdmin/AddStationButton.tsx
```diff
@@ -3,6 +3,10 @@ import AddCircleOutlineIcon from '@mui/icons-material/AddCircleOutline'
 import { Button } from '@mui/material'
 import AddStationModal from 'features/hfiCalculator/components/stationAdmin/AddStationModal'
 
+export interface AddStationButtonProps {
+  testId?: string
```
Sybrand, line 7: Is this used?

api/app/routers/hfi_calc.py
```diff
@@ -352,7 +357,7 @@ async def get_fire_centres(response: Response):
...
         fire_centres_list = await hydrate_fire_centres()
         return HFIWeatherStationsResponse(fire_centres=fire_centres_list)
```
Sybrand, line 362: I would add something here - that caches the response in redis if it doesn't exist.

Sybrand (changes requested):

api/app/db/models/hfi_calc.py
conbrad: I have checked, yes. The only reason for this change is because creating rows through the session fails with a sequence error.

api/app/db/crud/hfi_calc.py
```diff
@@ -79,6 +80,22 @@ def store_hfi_request(session: Session, hfi_result_request: HFIResultRequest, us
...
+def store_hfi_station(session: Session, station_code: int, fuel_type_id: int, planning_area_id: int, order: int):
+    """ Store planning weather station """
+    planning_weather_station = PlanningWeatherStation(station_code=station_code,
+                                                      fuel_type_id=fuel_type_id,
+                                                      planning_area_id=planning_area_id,
+                                                      order_of_appearance_in_planning_area_list=order)
+    session.add(planning_weather_station)
+
+
+def get_last_station_in_planning_area(session: Session, planning_area_id: int) -> PlanningWeatherStation:
```
andrea-williams, line 92: Shouldn't sonarcloud be complaining about the missing documentation comment for this function?...

api/app/routers/hfi_calc.py
andrea-williams: This change to the age of the cache contradicts the comment immediately above it. Please update.

api/app/routers/hfi_calc.py
```diff
@@ -361,6 +367,25 @@ async def get_fire_centres(response: Response):
...
 
 
+@router.post('/admin/add-station/{fire_centre_id}', status_code=201)
+async def add_station(fire_centre_id: int,
+                      request: HFIAddStationRequest,
+                      _=Depends(auth_with_station_admin_role_required)):
+    """ Adds a station. """
+    logger.info('/hfi-calc/admin/add-station/')
+    logger.info('request is: %s', request)
+    logger.info('fire centre is: %s', fire_centre_id)
```
andrea-williams, line 377: Do we need these two logger lines?

api/app/tests/hfi/test_hfi_endpoint_request.feature
```diff
@@ -62,3 +62,19 @@ Feature: /hfi/
             | /api/hfi-calc/fire_centre/1/2020-05-21/2020-05-25/planning_area/1/station/230/fuel_type/2                   | None                | 200         | hfi/test_hfi_endpoint_response_set_fuel_type.json        | True          | test_hfi_endpoint_stored_request.json |
             # Invalid fuel type should return 500 error, and not be saved.
             | /api/hfi-calc/fire_centre/1/2020-05-21/2020-05-25/planning_area/1/station/230/fuel_type/-1                  | None                | 500         | None                                                     | False         | None                                  |
+
+    Scenario: HFI - Admin POST add station
+        Given I received a POST request for hfi-calc admin <url> with <role>
```
andrea-williams, line 67: if we're being REALLY pedantic about grammar, I think it should be "Given I _send_ a POST request ..."

api/app/db/models/hfi_calc.py
Sybrand: There is (was?) (i think!) an issue with alembic, if you're not explicit with the sequence, it may create a sequence with a different name, so you end up with multiple sequences.

andrea-williams (changes requested):
If I try to add the same station to a fire centre twice (even if they're in different planning areas), I get a 500 error back.

Also, the spacing between text here is a little ugly:
Screen Shot 2022-05-03 at 4 02 28 PM

api/app/db/models/hfi_calc.py
Sybrand: Also - we are defining explicit sequences all over in hfi_calc and weather_models - we haven't had insert issues before.

Sybrand: Oh - also check if the sequence is only broken on your local machine. Could have been with upgrades + downgrades you did that you've gone out of sync. You'll need to restore from back and run alembic to be sure this change doesn't trigger a migration. I'm pretty sure we need that explicit sequence though.

andrea-williams:
I was playing around in Prince George Fire Centre & found a bug.

1. Add Afton station to G1 planning area.
2. Add Afton station to G3 planning area. On the front end, Afton will be removed from G1 and added to G3 instead.
3. Download the PDF for the fire centre. In the PDF, Afton appears in both G1 and G3 planning areas, which doesn't match the frontend UI. There also seems to be some alignment issues in the PDF
[PrinceGeorgeFireCentre_HFICalculator_2022-05-03_AWILLIAM@IDIR.pdf](https://github.com/bcgov/wps/files/8616818/PrinceGeorgeFireCentre_HFICalculator_2022-05-03_AWILLIAM%40IDIR.pdf)

api/app/tests/hfi/test_hfi_endpoint_request.feature
conbrad: I'd rather not these tests are brittle...

api/app/db/models/hfi_calc.py
Sybrand: Ok - having researched this - I take it all back. I have no idea why we're using explicit sequence naming.

Sybrand (changes requested):
Rather cache the entire response - and invalidate it when stations are changed.

api/app/hfi/fire_centre_cache.py
```diff
@@ -25,38 +25,23 @@ async def get_cached_hydrated_fire_centres(station_codes: List[int]) -> Optional
     if cached_json:
         logger.info('redis cache hit %s', key)
         cache_string = json.loads(json.loads(cached_json))
```
Sybrand, line 27:
```suggestion
        cached_json = json.loads(cached_json)
```
Why the double load? Also - it's not a cache_string once json has loaded it - it's a dict.

api/app/hfi/fire_centre_cache.py
```diff
@@ -0,0 +1,47 @@
...
+import logging
+from typing import Optional
+from app.schemas.hfi_calc import HFIWeatherStationsResponse
+from app.utils.redis import create_redis
+
+logger = logging.getLogger(__name__)
+cache_expiry_seconds = 86400
+key = "fire_centres"
+
+# we can safely cache the fire centres, as they don't change them very often.
```
Sybrand, line 13: Maybe throw this comment up at the top of the file?

api/app/hfi/hfi_calc.py
```diff
@@ -168,7 +168,9 @@ async def hydrate_fire_centres():
         wfwx_stations_data = await get_stations_by_codes(list(station_info_dict.keys()))
         # Iterate through all the stations from wildfire one.
 
+        station_codes = []
```
Sybrand, line 171: This variable isn't being used anymore?

api/app/hfi/hfi_calc.py
```diff
@@ -168,7 +168,9 @@ async def hydrate_fire_centres():
...
         for wfwx_station in wfwx_stations_data:
+            station_codes.append(wfwx_station.code)
```
Sybrand, line 173: No need for this?

api/app/hfi/fire_centre_cache.py
```diff
@@ -0,0 +1,47 @@
...
+async def put_cached_hydrated_fire_centres(response: HFIWeatherStationsResponse):
+    """ Caches fire centres that are assumed to be hydrated by caller. """
+    cache = create_redis()
+    try:
+        cache.set(key, json.dumps(response.json()).encode(), ex=cache_expiry_seconds)
+    except Exception as error:  # pylint: disable=broad-except
+        logger.error(error, exc_info=error)
+
+
+def clear_fire_centre_namespace():
```
Sybrand, line 44:
```suggestion
def clear_cached_hydrated_fire_centres():
```
Consistent naming?

api/app/tests/hfi/test_hfi_endpoint_request.feature
```diff
@@ -62,3 +62,19 @@ Feature: /hfi/
...
+        And it has a <request_body>
+        Then the response status code is <status_code>
+        Examples:
+            | url                               | role               | request_body                        | status_code |
+            # Test add station with correct role
+            | /api/hfi-calc/admin/add-station/1 | hfi_station_admin  | test_admin_add_station_request.json | 201         |
```
Sybrand, line 73: Do we care to test the duplicate record situation? 409 Conflict

Sybrand (changes requested):
Some comment/suggestions - looking good. Will test once build is complete.

api/app/hfi/fire_centre_cache.py
conbrad: It's a string according to the runtime error I get with a single load:
```
    cached_response: HFIWeatherStationsResponse=HFIWeatherStationsResponse(**cache_string)
TypeError: app.schemas.hfi_calc.HFIWeatherStationsResponse() argument after ** must be a mapping, not str
```

Sybrand: cache.get returns a binary array - you need:
```suggestion
        cache_string = json.loads(cached_json.decode())
```

api/app/tests/hfi/test_hfi_endpoint_request.feature
conbrad: Right now it 500's in this case

conbrad: Switched to 409 in https://github.com/bcgov/wps/pull/1973/commits/374daad17c235cd2992ca9e821b9b2ffd16bfc9b

api/app/hfi/fire_centre_cache.py
conbrad: Tested the above solution with a clean redis cache and server, and I still get the same error

conbrad: So the first load decode it as string json, the second load decodes it as a python object. Not sure what's need to skip the first string load.

Sybrand: Let's chat - I tried it and it worked... weird.

tessmgood (commented):
Is it possible to show errors in context, in the modal? 
e.g. 'station already exists' error. 
user hits save, error displays in modal, indicating they need to adjust their choice to proceed.

closing the modal makes me believe that it worked, the banner error at top of page is typically used for global-product errors preventing use of the product

andrea-williams:
This doesn't appear to be working anymore?
No fire centres are being loaded in the dropdown.

Screen Shot 2022-05-09 at 12 03 23 PM

conbrad:
> This doesn't appear to be working anymore? No fire centres are being loaded in the dropdown.
> 
> Screen Shot 2022-05-09 at 12 03 23 PM

Dev cluster overloaded, trying to clean it up

andrea-williams (commented):
There seems to be some flawed logic in when the error message gets displayed in the modal when a duplicated station is added.

Examples: I try to add a station that has the same default fuel type & planning area as what already exists. Error message appears - fine. If I close and reopen the Add Station modal, that error message will appear again, even though at this point I haven't selected anything in the modal. Even if I dismiss the error message, close & reopen the Add Station modal, that error message will appear again.

conbrad:
> There seems to be some flawed logic in when the error message gets displayed in the modal when a duplicated station is added.
> 
> Examples: I try to add a station that has the same default fuel type & planning area as what already exists. Error message appears - fine. If I close and reopen the Add Station modal, that error message will appear again, even though at this point I haven't selected anything in the modal. Even if I dismiss the error message, close & reopen the Add Station modal, that error message will appear again.

Fixed in https://github.com/bcgov/wps/pull/1973/commits/ee62c357ce8c896b93982f14a112bf9793886935

andrea-williams (changes requested):
Bug:

1. Try to add a station to a planning area that's a duplicate of what's already in the planning area.
2. Clicking "Save" button correctly displays error message.
3. In the same Add Station modal, select a different station that isn't in the planning area (for example, "Andrea Station"), and click "Save".
4. The page will refresh, but the newly added station doesn't display in the UI or in the PDF.
5. I'm thinking something must have gone wrong with the add, so I open the Add Station modal again and try to add "Andrea Station" to the planning area. I get an error message saying that the station already exists in the planning area, but it still isn't appearing in the UI.

api/app/db/crud/hfi_calc.py
```diff
@@ -1,23 +1,30 @@
...
 def get_fire_weather_stations(session: Session) -> CursorResult:
     """ Get all PlanningWeatherStation with joined FuelType, PlanningArea and FireCentre
     for the provided list of station_codes. """
-    return session.query(PlanningWeatherStation, FuelType, PlanningArea, FireCentre)\
+    q = session.query(PlanningWeatherStation, FuelType, PlanningArea, FireCentre)\
         .join(FuelType, FuelType.id == PlanningWeatherStation.fuel_type_id)\
         .join(PlanningArea, PlanningArea.id == PlanningWeatherStation.planning_area_id)\
         .join(FireCentre, FireCentre.id == PlanningArea.fire_centre_id)\
         .order_by(FireCentre.name, PlanningArea.name)
+    logger.debug((str(q.statement.compile(dialect=postgresql.dialect()))))
```
Sybrand, line 26: Do we need this going into prod?

conbrad:
> Bug:
> 
> 1. Try to add a station to a planning area that's a duplicate of what's already in the planning area.
> 2. Clicking "Save" button correctly displays error message.
> 3. In the same Add Station modal, select a different station that isn't in the planning area (for example, "Andrea Station"), and click "Save".
> 4. The page will refresh, but the newly added station doesn't display in the UI or in the PDF.
> 5. I'm thinking something must have gone wrong with the add, so I open the Add Station modal again and try to add "Andrea Station" to the planning area. I get an error message saying that the station already exists in the planning area, but it still isn't appearing in the UI.

What dates are you using? If it's for the most recent then it's possible the station added has no data and is not shown.

Sybrand (approved):

Sybrand (approved):

andrea-williams:
> > Bug:
> > 
> > 1. Try to add a station to a planning area that's a duplicate of what's already in the planning area.
> > 2. Clicking "Save" button correctly displays error message.
> > 3. In the same Add Station modal, select a different station that isn't in the planning area (for example, "Andrea Station"), and click "Save".
> > 4. The page will refresh, but the newly added station doesn't display in the UI or in the PDF.
> > 5. I'm thinking something must have gone wrong with the add, so I open the Add Station modal again and try to add "Andrea Station" to the planning area. I get an error message saying that the station already exists in the planning area, but it still isn't appearing in the UI.
> 
> What dates are you using? If it's for the most recent prep range then it's possible the station added has no data and is not shown. If you choose an earlier date with data and see the station that was added then that's why.

Ahhh yes that's it. Selecting a date range from last summer, the missing stations show up. 
However, once I've selected a date range in the past, I can only select other prep periods further in the past, so I can't go back to the present prep period. This doesn't appear to be happening in prod, but I also can't think why it would be happening specifically in your PR. Any idea?

api/app/hfi/hfi_request.py
```diff
@@ -0,0 +1,37 @@
...
+from itertools import groupby
+import operator
+from typing import List
+from app.db.models.hfi_calc import PlanningWeatherStation
+from app.schemas.hfi_calc import HFIResultRequest, StationInfo
+
+
+def update_result_request(result_request: HFIResultRequest,
+                          latest_stations: List[PlanningWeatherStation]):
+    """ Immutable copy of the result request with the latest planning weather stations """
```
Sybrand, line 12: Maybe just lack of sleep, but - it's not immutable is it?
```suggestion
    """Create copy of the result request with the latest planning weather stations """
```

api/app/hfi/hfi_request.py
```diff
@@ -0,0 +1,37 @@
...
+
+        planning_area_stations = stations_by_planning_area.get(planning_area_id)
+        updated_station_info_list: List[StationInfo] = []
+
+        for station in planning_area_stations:
+            existing_station = existing_stations_by_code.get(station.station_code)
+            station_info: StationInfo = StationInfo(
+                station_code=station.station_code,
+                selected=existing_station.selected if existing_station is not None else True,
+                fuel_type_id=station.fuel_type_id)
```
Sybrand, line 34: You need to grab the existing station fuel type if it does exist.

api/app/routers/hfi_calc.py
```diff
@@ -64,6 +67,7 @@ def get_prepared_request(
 
     TODO: give this function a better name.
     """
+    # pylint: disable=too-many-locals
```
Sybrand, line 70: Is it time to just update our linting settings? I can't say I've improved readability in our code by reducing the local count due to this linting warning - and we end up ignoring it a lot because it just doesn't make sense to reduce the count quite often.

api/app/tests/hfi/test_hfi_request.py
```diff
@@ -0,0 +1,45 @@
+""" Unit testing for hfi logic """
```
Sybrand, line 1: Add in a test to ensure the existing fuel type selection isn't overridden.

Sybrand (changes requested):
Some minor changes - but I like it.

api/app/hfi/hfi_request.py
conbrad: Input is immutable, so yeah, not the best comment

api/app/routers/hfi_calc.py
conbrad: No I want to refactor this function, there are too many locals and too much state but I'm not sure now's the time.

tessmgood (changes requested):
error rabbit hole: 
1. I try to add a station/fuel that exists: K2 - Clearwater - C7
2. Click save - I get an error that says station already exists - awesome!
3. I try to correct my error: I clear the fuel type field. Or I add a different station that isn't a duplicate. I'm seeing the error persist, it should go away at this point - it's confusing 
Screen Shot 2022-05-10 at 4 57 39 PM

web/src/features/hfiCalculator/components/stationAdmin/newStationForm.test.tsx
```diff
@@ -1,94 +1,143 @@
...
     )
-    return { getByTestId, queryByText }
   }
   describe('valid states', () => {
     it('should not render error message when new station has not been edited', () => {
-      const { queryByText } = renderNewStationForm({ dirty: false }, false)
+      const { queryByText } = renderNewStationForm({ dirty: false }, false, null)
 
       const errorMessage = queryByText('Please complete empty fields to continue')
       expect(errorMessage).not.toBeInTheDocument()
```
andrea-williams, line 39: Seems a bit risky to be testing again a search text message - this test will always pass if we happen to later change the error message text on the frontend.
But also, I can't think of a better solution right now...

conbrad: Not true, if you change the text this test will fail.

api/app/routers/hfi_calc.py
conbrad: But if we're open to changing lint settings, can we get rid of the mandatory comments?

tessmgood (approved):

Sybrand (changes requested):
bugs as discussed - 1) not loading new stations if change to fuel type is made before adding stations 2) reverting back to default period after adding station.

api/app/hfi/fire_centre_cache.py
```diff
@@ -0,0 +1,47 @@
...
+"""
+
+import json
+import logging
+from typing import Optional
+from app.schemas.hfi_calc import HFIWeatherStationsResponse
+from app.utils.redis import create_redis
+
+logger = logging.getLogger(__name__)
+cache_expiry_seconds = 86400
```
andrea-williams, line 13: Nice to have: a comment quantifying 86400 seconds in a more human-meaningful value (e.g., x days)

api/app/hfi/hfi_request.py
```diff
@@ -0,0 +1,38 @@
...
+from itertools import groupby
+import operator
+from typing import List
+from app.db.models.hfi_calc import PlanningWeatherStation
+from app.schemas.hfi_calc import HFIResultRequest, StationInfo
+
+
+def update_result_request(result_request: HFIResultRequest,
+                          latest_stations: List[PlanningWeatherStation]):
+    """ Returns copy of the result request with the latest planning weather stations """
```
andrea-williams, line 12: sorry, what do you mean by "latest" stations?

api/app/hfi/hfi_request.py
```diff
@@ -0,0 +1,38 @@
...
+    if len(latest_stations) == 0:
+        return result_request
+
+    updated_result_request = result_request.copy(deep=True)
+    get_attr = operator.attrgetter('planning_area_id')
+    stations_by_planning_area = dict((k, list(map(lambda x: x, values)))
+                                     for k, values in groupby(latest_stations, get_attr))
+    for planning_area_id in updated_result_request.planning_area_station_info:
+        existing_stations = result_request.planning_area_station_info[planning_area_id]
```
andrea-williams, line 21: also, what are "existing stations"?

andrea-williams (commented):
Review still in progress - just some questions

api/app/hfi/hfi_request.py
conbrad: `latest_persisted_stations`?

api/app/hfi/hfi_request.py
conbrad: Stations that were persisted in the last stored request

conbrad: `request_stations`?

api/app/hfi/hfi_request.py
conbrad: `currently_persisted_stations`?

andrea-williams: I like `currently_persisted_stations` best

andrea-williams (approved):
Code LGTM!

Sybrand (approved):
I'd prefer the endpoint to look a bit different - but I won't be pushy....

conbrad:
> HFI Calculator

Too afraid to change anything at this point with how fragile HFI backend code is. I'd like to defer this request until after this is deployed.

conbrad:
> I'd prefer the endpoint to look a bit different - but I won't be pushy....

Too afraid to change anything at this point with how fragile HFI backend code is. I'd like to defer this request until after this is deployed.

PR #2030, opened 2022-05-13
Task/firebat pdf spike/1574

andrea-williams:


andrea-williams:
Closing - might not be needed anymore. UX & PO to do deeper research first to explore user needs. Will keep branch in case PR needs to be re-opened later

PR #2121, opened 2022-06-21
Web served up by NGINX

Sybrand:
- Introduced nginx to serve up static content
- Ripped out web build in api - no longer needed
- Ripped out front-end code from api
- Ripped out matomo (less configuration to worry about)

*Doing a bit too much in this PR - but I kept working on other things while waiting for images to build!*

# Test Links:
[Percentile Calculator](https://wps-pr-2121.apps.silver.devops.gov.bc.ca/)
[MoreCast](https://wps-pr-2121.apps.silver.devops.gov.bc.ca/morecast)
[C-Haines](https://wps-pr-2121.apps.silver.devops.gov.bc.ca/c-haines)
[FireBat](https://wps-pr-2121.apps.silver.devops.gov.bc.ca/fire-behaviour-calculator)
[FireBat bookmark](https://wps-pr-2121.apps.silver.devops.gov.bc.ca/fire-behaviour-calculator?s=266&f=c5&c=NaN&w=20,s=286&f=c7&c=NaN&w=16,s=1055&f=c7&c=NaN&w=NaN,s=305&f=c7&c=NaN&w=NaN,s=344&f=c5&c=NaN&w=NaN,s=346&f=c7&c=NaN&w=NaN,s=328&f=c7&c=NaN&w=NaN,s=1399&f=c7&c=NaN&w=NaN,s=334&f=c7&c=NaN&w=NaN,s=1082&f=c3&c=NaN&w=NaN,s=388&f=c7&c=NaN&w=NaN,s=309&f=c7&c=NaN&w=16,s=306&f=c7&c=NaN&w=NaN,s=1029&f=c7&c=NaN&w=NaN,s=298&f=c7&c=NaN&w=NaN,s=836&f=c7&c=NaN&w=NaN,s=9999&f=c7&c=NaN&w=NaN)
[Fire Behaviour Advisory](https://wps-pr-2121.apps.silver.devops.gov.bc.ca/fire-behaviour-advisory)
[HFI Calculator](https://wps-pr-2121.apps.silver.devops.gov.bc.ca/hfi-calculator)
[FWI Calculator](https://wps-pr-2121.apps.silver.devops.gov.bc.ca/fwi-calculator)


Dockerfile.web
```diff
@@ -16,28 +16,23 @@ FROM registry.access.redhat.com/ubi8/nodejs-16 as static
...
-# # Copy the static content:
-# COPY --from=static /opt/app-root/src/build /app/static
+# Copy the static content:
+COPY --from=static /opt/app-root/src/build .
 
-# The fastapi docker image defaults to port 80, but openshift doesn't allow non-root users port 80.
-# EXPOSE 8080
\ No newline at end of file
+EXPOSE 3000
+# CMD ./prepare_config.sh && nginx -g "daemon off;"
```
andrea-williams, line 37: why is this commented out (and duplicated below)?

andrea-williams (approved):

openshift/nginx.conf
```diff
@@ -0,0 +1,38 @@
...
+    default_type  application/octet-stream;
+    access_log /dev/stdout;
+
+    server {
+        # listen on port 3000
+        listen 3000;
+        root /opt/app-root/src;
+
+        gzip on;
+        gzip_types application/javascript application/json text/css;
```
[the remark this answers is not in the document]
Sybrand: Sure - why not.

Dockerfile.web
Sybrand: Because mistake. Fixed.

openshift/nginx.conf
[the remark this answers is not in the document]
Sybrand: "By default, NGINX compresses responses only with MIME type text/html. To compress responses with other MIME types, include the [gzip_types](https://nginx.org/en/docs/http/ngx_http_gzip_module.html#gzip_types) directive and list the additional types."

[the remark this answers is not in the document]
Sybrand: I've added a comment to that effect though.

.github/workflows/deployment.yml
```diff
@@ -44,20 +44,11 @@ jobs:
...
-        shell: bash
-        run: |
-          oc login "${{ secrets.OPENSHIFT_CLUSTER }}" --token="${{ secrets.OC4_DEV_TOKEN }}"
-          EPHEMERAL_STORAGE=True bash openshift/scripts/oc_provision_backup_mariadb_cronjob.sh ${SUFFIX} apply
-
-  build-api-image:
-    name: Build API Image
+  build-web-image:
+    # Declared ahead of build-api-image ; it runs slightly slower than the api build, and putting
+    # it here increases the odds that it get's started 1st, so api and web are slightly more likely
```
conbrad, line 49: I'm not sure that this assumption of the order of job execution is true, but the only documentation I can find on job ordering is if you put them in the same concurrency group: https://docs.github.com/en/actions/using-jobs/using-concurrency. I think the safest assumption is any job can be picked up at any time if it has no dependencies.

openshift/nginx.conf
```diff
@@ -0,0 +1,72 @@
...
+        # https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Frame-Options
+        add_header X-Frame-Options DENY;
+
+        location / {
+            # First attempt to serve request as file, then
+            # as directory, then fall back to redirecting to index.html
+            try_files $uri $uri/ $uri.html /index.html;
+        }
+
+        location /health {
```
conbrad, line 48: Do these `/health` and `/ready` endpoints return 200 if nginx is up but the API isn't?

.github/workflows/deployment.yml
Sybrand: @conbrad I think you're right re. order of job execution. I don't think there's any kind of guarantee. I have however noticed that it seems to go from top to bottom. If it is top to bottom, then this will be slightly faster, if not, all is well.

openshift/nginx.conf
Sybrand: Yes. But the web and api run independently, so the deployment config as a whole only works when both API and WEB are healthy+ready.

conbrad (approved):

PR #2135, opened 2022-06-27
PDF colour coding

Sybrand:
- HTML that was being consumed by PDF was technically incorrect. Multiple <!DOCTYPE html>, html,  head and body tags. (two different HTML files are being concatenated). This seemed very wrong. So fixed. But that means that there are two "html" template files that look wrong - and vscode kept wanting to add the missing tags.
- Added colour coding to summary

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -37,6 +37,9 @@
             100
         ]
     },
+    "[html]": {
+        "editor.formatOnSave": false
+    },
     "[yaml]": {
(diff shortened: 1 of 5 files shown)


tessmgood (approved):
Nice! Curious if the borders are thick enough on the daily tables - will seek feedback in Review on Weds.
Going to create a separate ticket for HFI colour coding, users have requested this in UXR and by email to PSU.

conbrad (approved):

andrea-williams (approved):

PR #2146, opened 2022-07-06
Story/hfi admin remove station/1953

conbrad:


api/app/db/crud/hfi_calc.py
```diff
@@ -50,7 +50,9 @@ def get_planning_weather_stations(session, fire_centre_id: int) -> List[Planning
     """ Get all the stations for a fire centre. """
     return session.query(PlanningWeatherStation)\
         .join(PlanningArea, PlanningArea.id == PlanningWeatherStation.planning_area_id)\
-        .filter(PlanningArea.fire_centre_id == fire_centre_id).all()
+        .filter(PlanningArea.fire_centre_id == fire_centre_id)\
+        .filter(PlanningWeatherStation.is_deleted == False)\
```
Sybrand, line 54: Don't we need the is_deleted on every other query that related to PlanningWeatherStations?

api/app/db/models/hfi_calc.py
```diff
@@ -78,12 +78,20 @@ class PlanningWeatherStation(Base):
     fuel_type_id = Column(Integer, ForeignKey('fuel_types.id'), nullable=False, index=True)
     planning_area_id = Column(Integer, ForeignKey('planning_areas.id'), nullable=False, index=True)
     order_of_appearance_in_planning_area_list = Column(Integer, nullable=True)
+    # Track which user created the record for auditing purposes.
+    create_user = Column(String, nullable=False)
+    create_timestamp = Column(TZTimeStamp, nullable=False, index=True)
```
Sybrand, line 83: Do we need an index here? I don't think we sort / select on create_timestamp or update_timestamp anywhere do we?

api/alembic/versions/b99dce009b9f_change_default_fuel_types_for_coastal_.py
```diff
@@ -12,7 +12,7 @@
 
 # revision identifiers, used by Alembic.
 revision = 'b99dce009b9f'
-down_revision = '52c8a568364e'
+down_revision = '35acf3b96d8a'
```
Sybrand, line 15: Isn't changing the down revision on something that's already been pushed to production going to cause alembic to fail when we go to production?

Sybrand (changes requested):
Looking good - just a few comments. (I'd still have put the complexity on the front end and kept the API simpler. pOtato potAto.)

api/app/db/crud/hfi_calc.py
conbrad: Good catch!!!

api/app/db/models/hfi_calc.py
conbrad: True, and we probably do want an index on `is_deleted` since that will be filtered on in most cases.

api/app/db/crud/hfi_calc.py
```diff
@@ -50,7 +50,9 @@ def get_planning_weather_stations(session, fire_centre_id: int) -> List[Planning
     """ Get all the stations for a fire centre. """
     return session.query(PlanningWeatherStation)\
         .join(PlanningArea, PlanningArea.id == PlanningWeatherStation.planning_area_id)\
-        .filter(PlanningArea.fire_centre_id == fire_centre_id).all()
+        .filter(PlanningArea.fire_centre_id == fire_centre_id)\
```
andrea-williams, line 53: I think technically filtering on fire_centre_id isn't necessary because all the planning areas have unique ids. But also, I don't see any harm in keeping this in here :)

andrea-williams (commented):
Gonna take a look at this again tomorrow and try to stress/insanity test it. Only have a few minor comments re: code review.

andrea-williams (approved):

PR #2153, opened 2022-07-13
Story/hfi admin remove station fe/1953

conbrad:
- Adds new station management form for adding/removing stations to respective planning areas. Add added stations are automatically added in the last order of a planning area list. Editing of existing stations is not supported.
- Removes unique constraint on `(station_code, planning_area_id)` so that deleted stations can be added to the same planning area again.


web/src/features/hfiCalculator/components/stationAdmin/saveStationUpdatesButton.test.tsx
```diff
@@ -0,0 +1,119 @@
+import { render, waitFor } from '@testing-library/react'
+import userEvent from '@testing-library/user-event'
+import SaveStationUpdatesButton from 'features/hfiCalculator/components/stationAdmin/SaveStationUpdatesButton'
+import React from 'react'
+/**
```
Sybrand, line 5: Remove comments?

Sybrand (changes requested):
If I remove a station, after marking a fire zone in prep period as "ready", it remains "ready". I would expect the "ready" state to drop, since I've potentially made a change to the prep level.

conbrad:
> If I remove a station, after marking a fire zone in prep period as "ready", it remains "ready". I would expect the "ready" state to drop, since I've potentially made a change to the prep level.

This is difficult to achieve since station admin is completely decoupled from hfi request/ready states and has no knowledge of the `(fire_centre_id, date_range, create_timestamp)` tuple that uniquely identifies a `hfi request` which in turn identifies the ready states. How would we determine how far back in time we should consider `hfi_request` as not ready? Should all hfi requests that contain planning areas with added/removed stations be marked as not ready?

conbrad:
Closing because database is broken

PR #2157, opened 2022-07-13
Story/hfi admin remove station fe/1953

conbrad:
- Adds new station management form for adding/removing stations to respective planning areas. Added stations are automatically added in the last order of a planning area list. Editing of existing stations is not supported.
- Removes unique constraint on `(station_code, planning_area_id)` so that deleted stations can be added to the same planning area again.
- Adding/removing a station will set ready status to false for all HFI results in the affected planning area
# Test Links:
[Percentile Calculator](https://wps-pr-2157.apps.silver.devops.gov.bc.ca/)
[MoreCast](https://wps-pr-2157.apps.silver.devops.gov.bc.ca/morecast)
[C-Haines](https://wps-pr-2157.apps.silver.devops.gov.bc.ca/c-haines)
[FireBat](https://wps-pr-2157.apps.silver.devops.gov.bc.ca/fire-behaviour-calculator)
[FireBat bookmark](https://wps-pr-2157.apps.silver.devops.gov.bc.ca/fire-behaviour-calculator?s=266&f=c5&c=NaN&w=20,s=286&f=c7&c=NaN&w=16,s=1055&f=c7&c=NaN&w=NaN,s=305&f=c7&c=NaN&w=NaN,s=344&f=c5&c=NaN&w=NaN,s=346&f=c7&c=NaN&w=NaN,s=328&f=c7&c=NaN&w=NaN,s=1399&f=c7&c=NaN&w=NaN,s=334&f=c7&c=NaN&w=NaN,s=1082&f=c3&c=NaN&w=NaN,s=388&f=c7&c=NaN&w=NaN,s=309&f=c7&c=NaN&w=16,s=306&f=c7&c=NaN&w=NaN,s=1029&f=c7&c=NaN&w=NaN,s=298&f=c7&c=NaN&w=NaN,s=836&f=c7&c=NaN&w=NaN,s=9999&f=c7&c=NaN&w=NaN)
[Fire Behaviour Advisory](https://wps-pr-2157.apps.silver.devops.gov.bc.ca/fire-behaviour-advisory)
[HFI Calculator](https://wps-pr-2157.apps.silver.devops.gov.bc.ca/hfi-calculator)
[FWI Calculator](https://wps-pr-2157.apps.silver.devops.gov.bc.ca/fwi-calculator)


api/app/hfi/hfi_admin.py
```diff
@@ -149,11 +149,19 @@ def get_next_order(updated_stations: List[PlanningWeatherStation], other_station
...
+    existing_orders = [
+        station.order_of_appearance_in_planning_area_list for station in other_stations
+        if station.order_of_appearance_in_planning_area_list is not None]
 
     if len(updated_orders) == 0:
         return max(existing_orders) + 1
     if len(existing_orders) == 0:
         return max(updated_orders) + 1
 
     return min(max(updated_orders) + 1, max(existing_orders) + 1)
```
andrea-williams, line 163: I'd prefer to have a comment here explaining the logic behind this. I'm trying to wrap my head around it but it's a bit complicated and not immediately obvious what's going on here.

conbrad: I simplified the logic here after some further testing and updated the comment.

api/app/db/crud/hfi_calc.py
```diff
@@ -1,11 +1,12 @@
 """ CRUD operations relating to HFI Calculator
 """
-from typing import List
+from typing import List, Optional
 from sqlalchemy.engine.cursor import CursorResult
 from sqlalchemy.orm import Session
-from sqlalchemy import desc, insert
+from sqlalchemy import and_, desc, exists, insert, update
+from sqlalchemy.dialects import postgresql
```
Sybrand, line 7: Where is this being used?

api/app/db/crud/hfi_calc.py
```diff
@@ -1,11 +1,12 @@
 """ CRUD operations relating to HFI Calculator
 """
-from typing import List
+from typing import List, Optional
 from sqlalchemy.engine.cursor import CursorResult
 from sqlalchemy.orm import Session
-from sqlalchemy import desc, insert
+from sqlalchemy import and_, desc, exists, insert, update
```
Sybrand, line 6:
```suggestion
from sqlalchemy import desc, insert, update
```
and_ , exists ? not being used?

api/app/schemas/hfi_calc.py
```diff
@@ -191,8 +191,10 @@ class HFIAdminRemovedStation(BaseModel):
 
 class HFIAdminStationUpdateRequest(BaseModel):
     """ Request input for updating a batch of HFI stations. """
+    fire_centre_id: int
     added: List[HFIAdminAddedStation]
     removed: List[HFIAdminRemovedStation]
+    date_range: Optional[DateRange]
```
Sybrand, line 197: Maybe I'm missing it - but where is this date range being used?

The only place I can find is in unready_planning_areas, and there, only the start_date of the range is being used.

api/app/db/crud/hfi_calc.py
```diff
@@ -175,6 +178,24 @@ def get_stations_for_affected_planning_areas(session: Session, request: HFIAdmin
...
+                           planning_area_ids):
+    if date_range is None:
+        return
+    now = get_utc_now()
+    query = update(HFIReady).values(
+        {HFIReady.ready: False, HFIReady.update_user: username, HFIReady.update_timestamp: now})\
+        .where(HFIReady.planning_area_id.in_(planning_area_ids))\
+        .where(HFIReady.hfi_request_id == HFIRequest.id)\
+        .where(HFIRequest.fire_centre_id == fire_centre_id)\
+        .where(HFIRequest.prep_start_day >= date_range.start_date).\
```
Sybrand, line 194: I think we need to add some explicit comments around **why** this is happening. Looking at the code it isn't clear why we're only changing requests after the "start_date", as opposed to all requests. I would guess it's business logic, there's some rule that states that we don't "unready" old requests. But what's to stop me from changing my prep period to 1997, and then editing the stations?
Why discriminate? Why not just mark them all as unready - it should be a fairly fast database operation?

Sybrand (changes requested):
@conbrad some minor requests around the toggle. Also - question - when I delete all the stations from a zone, I get a 500 error. I don't know if it's worth worrying about? I don't see why one would ever want to delete all stations?

api/app/db/crud/hfi_calc.py
conbrad: Sure. I guess the behaviour we need to accept is that all "readied" HFI prep periods in the past are now unready. The model in my head is that a station modification affects current a future prep periods, hence the `gte` start date discrimination.

conbrad:
> @conbrad some minor requests around the toggle. Also - question - when I delete all the stations from a zone, I get a 500 error. I don't know if it's worth worrying about? I don't see why one would ever want to delete all stations?

Unlikely but I can address this case in a follow up PR.

api/app/db/crud/hfi_calc.py
Sybrand: I would go with whatever Tess thinks is more appropriate on this.

tessmgood: I can't test this in prod, so this is the logic I'd advocate for: 
A - if the station changes are applied historically, un-ready all history as well - be consistent (I think this is the reality?)
or
B - if the station changes are only applied today onwards, and historical station lists are saved, then only un-ready today forward
- _If there were a stronger use case for recording history in the product this would have made sense, but because of the actuals context, we rely on PDF records instead as an MVP_

conbrad: Ok sounds good, I'll make the changes to unready all states for a planning area for added/removed stations.

PR #2202, opened 2022-08-09
Switch to using an async driver

Sybrand:
Switch to using an async driver - putting things in place for only using async going forward - and possibly taking some time to re-factor old stuff to running async.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -63,6 +63,16 @@
             "console": "integratedTerminal"
         },
         {
+            "name": "advisory calculate",
+            "type": "python",
+            "request": "launch",
+            "module": "advisory.calculate_advisories",
+            "console": "integratedTerminal",
+            "args": [
+                "2022-08-09"
+            ]
+        },
+        {
             "name": "Chrome",
             "type": "pwa-chrome",
             "request": "launch",
--- a/api/advisory/db/crud.py
+++ b/api/advisory/db/crud.py
@@ -1,22 +1,34 @@
 """ Create/Read/Update/Delete """
 from datetime import date
-from sqlalchemy.orm import Session
+from sqlalchemy import select
+from sqlalchemy.ext.asyncio import AsyncSession
 from advisory.db.models.tileserver import (Hfi, FireZone)
 
 
-def get_hfi_area_percentages(session: Session, for_date: date):
+async def get_hfi_area_percentages(session: AsyncSession, for_date: date):
     """ This is terribly slow!
 
     For each fire zone, it gives you the area of the fire zone, and the area of hfi polygons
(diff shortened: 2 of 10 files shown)


.vscode/launch.json
```diff
@@ -62,6 +62,16 @@
             "module": "api.advisory.polygonize_hfi",
             "console": "integratedTerminal"
         },
+        {
+            "name": "advisory calculate",
```
conbrad, line 66: ❤️

api/advisory/db/crud.py
```diff
@@ -1,22 +1,27 @@
...
+                  FireZone.mof_fire_zone_id,
+                  FireZone.mof_fire_zone_name,
+                  FireZone.geom.ST_Transform(3005).ST_Area().label('zone_area'),
+                  Hfi.wkb_geometry.ST_Union().ST_Intersection(FireZone.geom)
+                  .ST_Transform(3005).ST_Area().label('hfi_area'))\
         .join(Hfi, Hfi.wkb_geometry.ST_Intersects(FireZone.geom))\
-        .filter(Hfi.date == for_date)\
+        .where(Hfi.date == for_date)\
         .group_by(FireZone.id)
+    result = await session.execute(stmt)
```
conbrad, line 26: Is it possible to return an `await`ed statement in python like in Typescript?
e.g.

`return await session.execute(stmt)`

or

`return await session.query(....)`

conbrad (approved):

api/advisory/db/crud.py
Sybrand: Yes - in this cause though, I'm throwing it into `result` so that I can throw it into `result.all()` - I suppose we could throw that all into one line - but I'd prefer not to - chaining like that just ends up making debugging more difficult.

api/app/db/database.py
```diff
@@ -69,9 +95,11 @@ def get_read_session_scope() -> Generator[Session, None, None]:
         session.close()
 
 
-@ contextmanager
+@contextmanager
 def get_write_session_scope() -> Generator[Session, None, None]:
-    """Provide a transactional scope around a series of operations."""
+    """Provide a transactional scope around a series of operations.
+    THIS METHOD IS DEPRECATED! PLEASE MOVE TO USING: get_async_write_session_scope
```
andrea-williams, line 101: is it worthwhile creating a Task in ZenHub to remind ourselves to remove the sync session scope()s at some point? Or are you confident you/we will remember?

andrea-williams (approved):

api/app/db/database.py
Sybrand: It's going to be around a LOOOOONG time.... But sure - I'll create a dev task for migrating.

Issue #1579, opened 2021-11-25
UTC-7? What's up with that!!!


Sybrand:
Problem:
Have the wrong time zones sneak into the code is horrible and disastrous. This needs to be fixed!

Make it so:
- We should only be concerned with UTC-8 (pacific standard time)
- All references to -7 to be removed
- All hard coded -8 values to be removed - we should be using the -8 constants everywhere
- The reason for this should be extremely well documented in code

Sybrand:
Uh oh. So - sometimes PDT does matter. For example, if you want to know on what DAY something happened, in the summer, then you need to be concerned with PDT. 23h50 could fall on a different day depending if it's PDT, PST or UTC.

PR #2246, opened 2022-08-24
SFMS follow up

Sybrand:
- Fix folder date bug (5pm runs would push the process date ahead because of UTC)
- Handling ; for comments in .ini file
# Test Links:
[Percentile Calculator](https://wps-pr-2246.apps.silver.devops.gov.bc.ca/)
[MoreCast](https://wps-pr-2246.apps.silver.devops.gov.bc.ca/morecast)
[C-Haines](https://wps-pr-2246.apps.silver.devops.gov.bc.ca/c-haines)
[FireBat](https://wps-pr-2246.apps.silver.devops.gov.bc.ca/fire-behaviour-calculator)
[FireBat bookmark](https://wps-pr-2246.apps.silver.devops.gov.bc.ca/fire-behaviour-calculator?s=266&f=c5&c=NaN&w=20,s=286&f=c7&c=NaN&w=16,s=1055&f=c7&c=NaN&w=NaN,s=305&f=c7&c=NaN&w=NaN,s=344&f=c5&c=NaN&w=NaN,s=346&f=c7&c=NaN&w=NaN,s=328&f=c7&c=NaN&w=NaN,s=1399&f=c7&c=NaN&w=NaN,s=334&f=c7&c=NaN&w=NaN,s=1082&f=c3&c=NaN&w=NaN,s=388&f=c7&c=NaN&w=NaN,s=309&f=c7&c=NaN&w=16,s=306&f=c7&c=NaN&w=NaN,s=1029&f=c7&c=NaN&w=NaN,s=298&f=c7&c=NaN&w=NaN,s=836&f=c7&c=NaN&w=NaN,s=9999&f=c7&c=NaN&w=NaN)
[Fire Behaviour Advisory](https://wps-pr-2246.apps.silver.devops.gov.bc.ca/fire-behaviour-advisory)
[HFI Calculator](https://wps-pr-2246.apps.silver.devops.gov.bc.ca/hfi-calculator)
[FWI Calculator](https://wps-pr-2246.apps.silver.devops.gov.bc.ca/fwi-calculator)

### DIFF
--- a/api/app/routers/sfms.py
+++ b/api/app/routers/sfms.py
@@ -7,7 +7,7 @@ from tempfile import SpooledTemporaryFile
 from fastapi import APIRouter, UploadFile, Response, Request
 from app.utils.s3 import get_client
 from app import config
-from app.utils.time import get_hour_20, get_utc_now, get_pst_now
+from app.utils.time import get_hour_20, get_pdt_now
 
 logger = logging.getLogger(__name__)
 
@@ -52,7 +52,7 @@ def is_actual(filename: str) -> bool:
         year=int(file_date_string[:4]),
         month=int(file_date_string[4:6]),
         day=int(file_date_string[6:8]))
-    now = get_pst_now()
+    now = get_pdt_now()
     now_date = now.date()
 
     if file_date < now_date:
@@ -82,7 +82,7 @@ def get_target_filename(filename: str) -> str:
     # We are assuming that the local server time, matches the issue date. We assume that
     # right after a file is generated, this API is called - and as such the current
     # time IS the issue date.
-    issue_date = get_utc_now()
+    issue_date = get_pdt_now()
     # depending on the issue date, we decide if it's a forecast or actual.
     prefix = get_prefix(filename)
     # create the filename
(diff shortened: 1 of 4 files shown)


api/app/routers/sfms.py
```diff
@@ -52,7 +52,7 @@ def is_actual(filename: str) -> bool:
         year=int(file_date_string[:4]),
         month=int(file_date_string[4:6]),
         day=int(file_date_string[6:8]))
-    now = get_pst_now()
+    now = get_pdt_now()
```
andrea-williams, line 55: I'm confused - why the use of PDT here? Based on https://github.com/bcgov/wps/issues/1579, I thought using PDT was a big no-no?

Sybrand: I know right! Timezone nightmares.
So. In this case. Here's the deal:
- If it's say, the 25th of August. Then it means we (citizens of bc) are using PDT.
- If we have a server, that makes a prediction, on the 25th of August, then we refer to that prediction as having run on the 25 of August.
- If we use a different timezone, then we may end up saying it was on a different date! So using PST, we may misattribute the date something happened on.


Such garbage. BC should just switch to PST as Glod intended.

andrea-williams: I'm still not able to wrap my head around this, but I don't want to have to think about it anymore. I partially trust your logic on this, and if it turns out to be wrong, well then I'll just blame you.

andrea-williams (approved):

api/app/routers/sfms.py
Sybrand: Ok - switched to using "america/vancouver" - if you care to look again.

PR #2250, opened 2022-08-30
Get docker working for M1

Sybrand:
- Change Dockerfile base image - buster didn't have M1 compatible image  (also easer i.m.h.o. to work with Ubuntu for local development)
- Make a number of changes to get python to run as non root
- Updated some "docker compose" commands - "docker-compose" is now redundant, compose is build into latest versions of docker.
- Allowing a more recent version of python, we need to update to the latest version of python across our base anyway!
# Test Links:
[Percentile Calculator](https://wps-pr-2250.apps.silver.devops.gov.bc.ca/)
[MoreCast](https://wps-pr-2250.apps.silver.devops.gov.bc.ca/morecast)
[C-Haines](https://wps-pr-2250.apps.silver.devops.gov.bc.ca/c-haines)
[FireBat](https://wps-pr-2250.apps.silver.devops.gov.bc.ca/fire-behaviour-calculator)
[FireBat bookmark](https://wps-pr-2250.apps.silver.devops.gov.bc.ca/fire-behaviour-calculator?s=266&f=c5&c=NaN&w=20,s=286&f=c7&c=NaN&w=16,s=1055&f=c7&c=NaN&w=NaN,s=305&f=c7&c=NaN&w=NaN,s=344&f=c5&c=NaN&w=NaN,s=346&f=c7&c=NaN&w=NaN,s=328&f=c7&c=NaN&w=NaN,s=1399&f=c7&c=NaN&w=NaN,s=334&f=c7&c=NaN&w=NaN,s=1082&f=c3&c=NaN&w=NaN,s=388&f=c7&c=NaN&w=NaN,s=309&f=c7&c=NaN&w=16,s=306&f=c7&c=NaN&w=NaN,s=1029&f=c7&c=NaN&w=NaN,s=298&f=c7&c=NaN&w=NaN,s=836&f=c7&c=NaN&w=NaN,s=9999&f=c7&c=NaN&w=NaN)
[Fire Behaviour Advisory](https://wps-pr-2250.apps.silver.devops.gov.bc.ca/fire-behaviour-advisory)
[HFI Calculator](https://wps-pr-2250.apps.silver.devops.gov.bc.ca/hfi-calculator)
[FWI Calculator](https://wps-pr-2250.apps.silver.devops.gov.bc.ca/fwi-calculator)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,11 +3,11 @@ docker-build:
 	# Build dev docker images.
 	# Having issues? try: docker volume prune
 	# Still having issues? try: docker system prune
-	docker-compose build
+	docker compose build
 
 docker-run:
 	# Run docker in dev mode.
-	docker-compose up
+	docker compose up
 
 docker-db:
 	# Run the database
@@ -22,10 +22,10 @@ docker-web-server:
 	docker run -p 3000:3000 wps/web
 
 docker-shell-api:
-	# Shell into the dev container.
-	docker-compose run --rm api bash
+	# Shell into the running dev container.
+	docker compose exec api bash
 
 docker-shell-web:
-	# Shell into the dev container.
+	# Shell into the running dev container.
 	# docker run -it --env-file app/.env --entrypoint bash wps-api_api:latest
-	docker-compose run --rm web bash
+	docker compose exec web bash
--- a/Makefile
+++ b/Makefile
@@ -11,7 +11,7 @@ docker-run:
 
 docker-db:
 	# Run the database
-	docker-compose up db
+	docker compose up db
 
 docker-web-server-build:
 	# Build the web in server mode (same dockerfile as used in production)
--- a/api/Dockerfile
+++ b/api/Dockerfile
@@ -1,12 +1,22 @@
 # NOTE:
 # This Dockerfile is for local development only!
 
-FROM python:3.8.12-buster
+# debian would match more closely what we have in production, and would probably be ideal,
+# but it's also a pain working with because debian is so old.
+FROM ubuntu:22.04
 
-# wildfire one may have an old certificate, so we have to make debian more forgiving.
-# RUN sed -i 's/TLSv1.2/TLSv1.0/g' /etc/ssl/openssl.cnf
+ARG USERNAME=worker
+ARG USER_UID=1000
+ARG USER_GID=$USER_UID
 
-# Install dependancies need by python developer packages
+# Tell r-base not to wait for interactive input.
+ENV DEBIAN_FRONTEND=noninteractive
+
+# Install dependancies needed by python developer packages
+# One should really run all these installs and the update in one go - for a consistent install
+# but ease of development trumps consistency in this instance: it's easer to have more
+# faster running steps that can fail, that one big monster install that takes forever 
+# and fails.
 # NOTE: Once we no longer need pyodbc, please remove the apt-get update and install commands below.
 RUN apt-get -y update
 RUN apt-get -y install unixodbc-dev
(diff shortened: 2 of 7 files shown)


Makefile
```diff
@@ -22,10 +22,10 @@ docker-web-server:
 	docker run -p 3000:3000 wps/web
 
 docker-shell-api:
-	# Shell into the dev container.
-	docker-compose run --rm api bash
+	# Shell into the running dev container.
+	docker compose exec api bash
```
conbrad, line 26: This is not a command on my machine but the original `docker-compose` is

conbrad (changes requested):

api/Dockerfile
```diff
@@ -19,44 +29,50 @@ RUN apt-get update --fix-missing && apt-get -y install r-base
...
-RUN cd /tmp && \
-    poetry install --no-root
+COPY --chown=worker:worker poetry.lock pyproject.toml ./
+RUN python3 -m pip install cachecontrol
+
+RUN poetry install
+
+# We can't have this inside pyproject.toml because the gdal version differs from platform to platform.
+# To figure out what version of pygdal you need, run gdal-config
+RUN poetry run python -m pip install pygdal==3.4.1.10
```
conbrad, line 68: Does the latest `pygdal` (3.5.1.11) work?
https://pypi.org/project/pygdal/3.5.1.11/

Makefile
Sybrand: `docker-compose` is an implementation of Docker Compose V1, and has been deprecated. If you don't have `docker compose` it means you're running an older version of docker, and you should upgrade. `docker-compose` is an additional dependency, so apart from everything else, it simplifies the development environment.

I don't think we should continue to use a deprecated version, and should transition to v2.
https://docs.docker.com/compose/#compose-v2-and-the-new-docker-compose-command
https://www.docker.com/blog/announcing-compose-v2-general-availability/

api/Dockerfile
Sybrand: No. Latest version that comes with Ubuntu 22.04 is 3.4.1

Makefile
conbrad: Huh weird. I'm running: `Docker version 20.10.16, build aa7e414fdc`

But I use the `limactl` wrapper which I guess just supports the old command.

conbrad (approved):

Makefile
Sybrand: Maybe your `limactl` needs an update? lima with nerdctl seems to support it - so it's strange that it wouldn't work with docker?

conbrad: I needed to manually symlink the `docker compose` command as a plugin:

Compose is now a Docker plugin. For Docker to find this plugin, symlink it:    mkdir -p ~/.docker/cli-plugins    ln -sfn $(brew --prefix)/opt/docker-compose/bin/docker-compose ~/.docker/cli-plugins/docker-compose
--

https://formulae.brew.sh/formula/docker-compose

Makefile
```diff
@@ -3,11 +3,11 @@ docker-build:
...
 	# Still having issues? try: docker system prune
-	docker-compose build
+	docker compose build
 
 docker-run:
 	# Run docker in dev mode.
-	docker-compose up
+	docker compose up
 
 docker-db:
```
conbrad, line 12: There's an old `docker-compose` in this command

andrea-williams (approved):
After much pain, the dockerfile builds and runs on my machine 👍

Makefile
Sybrand: In that case - updating all the other references too.

PR #2252, opened 2022-09-01
Simplified zone geometry

Sybrand:
- Add generalised "shape" object (to stick zones and things into)
- Importing simplified zones
- Added files/config for developing api inside dev container
- Updated some readme's
- Ran `poetry update` - and added some timeouts where lint was complaining.
# Test Links:
[Percentile Calculator](https://wps-pr-2252.apps.silver.devops.gov.bc.ca/)
[MoreCast](https://wps-pr-2252.apps.silver.devops.gov.bc.ca/morecast)
[C-Haines](https://wps-pr-2252.apps.silver.devops.gov.bc.ca/c-haines)
[FireBat](https://wps-pr-2252.apps.silver.devops.gov.bc.ca/fire-behaviour-calculator)
[FireBat bookmark](https://wps-pr-2252.apps.silver.devops.gov.bc.ca/fire-behaviour-calculator?s=266&f=c5&c=NaN&w=20,s=286&f=c7&c=NaN&w=16,s=1055&f=c7&c=NaN&w=NaN,s=305&f=c7&c=NaN&w=NaN,s=344&f=c5&c=NaN&w=NaN,s=346&f=c7&c=NaN&w=NaN,s=328&f=c7&c=NaN&w=NaN,s=1399&f=c7&c=NaN&w=NaN,s=334&f=c7&c=NaN&w=NaN,s=1082&f=c3&c=NaN&w=NaN,s=388&f=c7&c=NaN&w=NaN,s=309&f=c7&c=NaN&w=16,s=306&f=c7&c=NaN&w=NaN,s=1029&f=c7&c=NaN&w=NaN,s=298&f=c7&c=NaN&w=NaN,s=836&f=c7&c=NaN&w=NaN,s=9999&f=c7&c=NaN&w=NaN)
[Fire Behaviour Advisory](https://wps-pr-2252.apps.silver.devops.gov.bc.ca/fire-behaviour-advisory)
[HFI Calculator](https://wps-pr-2252.apps.silver.devops.gov.bc.ca/hfi-calculator)
[FWI Calculator](https://wps-pr-2252.apps.silver.devops.gov.bc.ca/fwi-calculator)


api/app/db/models/advisory.py
```diff
@@ -0,0 +1,44 @@
...
+        # we may have to re-visit this constraint - but for the time being, the idea is
+        # that for any given type of area, it has to be unique for the kind of thing that
+        # it is. e.g. a zone has some id.
+        UniqueConstraint('external_identifier', 'shape_type'),
+        {'comment': 'Record identifying some area of interest with respect to advisories'}
+    )
+
+    id = Column(Integer, primary_key=True)
+    # An area is uniquely identified, e.g. a zone has a number, so does a fire.
+    external_identifier = Column(String, nullable=False, index=True)
```
conbrad, line 38: Just want to suggest naming this `origin_identifier`, external isn't really clear to me, `origin` suggests it comes from somewhere else.

conbrad (approved):

api/app/db/models/advisory.py
Sybrand: I like it. Don't know that origin captures it either - but better than "external".

api/alembic/versions/c04f22e31997_import_zones.py
```diff
@@ -0,0 +1,83 @@
...
+    for object_id in zone_ids:
+        # Fetch each object in turn.
+        obj = esri.fetch_object(object_id, zone_url)
+        for feature in obj.get('features', []):
+            attributes = feature.get('attributes', {})
+            # Each zone is uniquely identified by a fire zone id.
+            mof_fire_zone_id = attributes.get('MOF_FIRE_ZONE_ID')
+            fire_zone_id = str(int(mof_fire_zone_id))
+            geometry = feature.get('geometry', {})
+            rings = geometry.get('rings', [[]])
```
andrea-williams, line 56: I think I missed something - what are these "rings"?

api/app/db/models/advisory.py
Sybrand: Changed to "source_identifier"

api/app/health.py
```diff
@@ -22,7 +22,7 @@ def patroni_cluster_health_check():
     header = {
         'Authorization': 'Bearer ' + config.get('STATUS_CHECKER_SECRET')
     }
-    resp = requests.get(url, headers=header)
+    resp = requests.get(url, headers=header, timeout=10)
```
andrea-williams, line 25: just wanted to double-check that the units for timeout are seconds, not milliseconds? (bc I was kinda expecting it would be ms for consistency with other cases where time is used, but for a timeout, seconds makes more sense than milliseconds)

api/app/utils/esri.py
```diff
@@ -0,0 +1,72 @@
...
+    """
+    Fetch object list from a feature layer.
+
+    url: layer url to fetch
+    (e.g. https://maps.gov.bc.ca/arcserver/rest/services/whse/bcgw_pub_whse_legal_admin_boundaries/MapServer/2)
+    """
+    logger.info('fetching object list for %s...', url)
+
+    params = {
+        'where': '1=1',
```
andrea-williams, line 22: is this `where` clause necessary?

api/alembic/versions/c04f22e31997_import_zones.py
Sybrand: For some reason that's the convention on the ARC format. No idea why they're using that convention. It's all over their documentation: https://totalapis.github.io/api-reference/esri-geometry-Polygon.html

Sybrand: You want a comment in the code? It doesn't make it an less confusing....

api/app/health.py
Sybrand: Consistent? Python? It's seconds - from the source code - lib/python3.8/site-packages/requests/api.py:

":param timeout: (optional) How many seconds to wait for the server to send data
        before giving up, as a float, or a :ref:`(connect timeout, read
        timeout) <timeouts>` tuple."

Sybrand: Shall I throw in a comment?

api/alembic/versions/c04f22e31997_import_zones.py
andrea-williams: Their definition for "rings" is "an array of rings", which is not overly helpful, but regardless, yes I'd like that link commented in the code so at least it's documented that the terminology used there is a convention and not something you just pulled out of the dark recesses of your brain.

api/app/utils/esri.py
Sybrand: Unfortunately yes. There's no "*" equivalent. There's no direct way of asking for everything, so that hack is to pass a condition that is true for all records.

Sybrand: Oh - again - you want a comment in the code?

andrea-williams: Gotcha. (Very unusual choice on their part to require a where clause though...)

andrea-williams: Yeah I think a comment there would be wise.

andrea-williams (approved):

api/alembic/versions/c04f22e31997_import_zones.py
Sybrand: one ring to rule them all, and in the darkness bind them

PR #2236, opened 2022-08-23
Story/fba slider threshold

conbrad:
- Fire zone advisory endpoint (`/fire-zone-areas`) performs `hfi`/`fire_zones` postgis intersection lookup on the fly since we simplified the fire zones
- Adds a threshold slider on the frontend to paint advisory layer based on the dynamic hfi area % threshold number

PR #2256, opened 2022-09-07
Story/fba slider sybrand

Sybrand:
From @conbrad :
- Fire zone advisory endpoint (`/fire-zone-areas`) performs `hfi`/`fire_zones` postgis intersection lookup on the fly since we simplified the fire zones
- Adds a threshold slider on the frontend to paint advisory layer based on the dynamic hfi area % threshold number

From @Sybrand :
- fire-zone-areas endpoint attempts to classify hfi for date requested before doing intersection lookup
- fire zone intersection now down using generic "shape" table in api database
- removed required read access to tile server database
- remove "FireZoneAdvisory" model, since areas are being calculated on the fly

NOTE:
You may have trouble with your alembic migration. If you're run an `upgrade head` on #2236 you're going to be in trouble. I restored from backup to be sure the migration is compatible with production. You may have to trash your database if you can't revert easily.


api/alembic/versions/2cc32257e8bc_rethink_classification.py
```diff
@@ -0,0 +1,71 @@
...
+    # ### end Alembic commands ###
+    # Advisory shapes was missing an SRID:
+    session = Session(bind=op.get_bind())
+    session.execute("SELECT UpdateGeometrySRID('advisory_shapes','geom',3005);")
+
+
+def downgrade():
+    # ### commands (mostly) auto generated by Alembic ###
+    op.drop_index('idx_advisory_areas_geom', table_name='advisory_shapes', postgresql_using='gist')
+    op.create_index('idx_advisory_shapes_geom', 'advisory_shapes', ['geom'], unique=False, postgresql_using='gist')
```
andrea-williams, line 52: I'm confused why you're creating this index as part of the downgrade() when the exact same index is created as part of the upgrade()?

api/app/auto_spatial_advisory/classify_hfi.py
```diff
@@ -0,0 +1,48 @@
...
+    # Create an object with the same dimensions as the input, but with 8 bit unsigned values.
+    target_tiff = output_driver.Create(target_path, xsize=source_band.XSize,
+                                       ysize=source_band.YSize, bands=1, eType=gdal.GDT_Byte)
+    # Set the geotransform and projection to the same as the input.
+    target_tiff.SetGeoTransform(source_tiff.GetGeoTransform())
+    target_tiff.SetProjection(source_tiff.GetProjection())
+
+    # Write the classified data to the band.
+    target_band = target_tiff.GetRasterBand(1)
+    target_band.SetNoDataValue(0)
```
andrea-williams, line 37: Variable-izing this 0 value might be a good idea. In my branch I've been using -10000 as the NoDataValue to remain consistent with the FTL.

conbrad (commented):
Code LGTM, just gonna try running the migrations locally

api/app/auto_spatial_advisory/classify_hfi.py
```diff
@@ -0,0 +1,48 @@
...
+
+    # Important to make sure data is flushed to disk!
+    target_tiff.FlushCache()
+
+    # Explicit delete to make sure underlying resources are cleared up!
+    del source_band
+    del source_tiff
+    del target_band
+    del target_tiff
+    del output_driver
```
andrea-williams, line 48: is there a difference between `del`ing these or just setting them to `None`? `None` seemed to be more common based on what I saw online

api/app/auto_spatial_advisory/polygonize.py
```diff
@@ -0,0 +1,52 @@
...
+    spatial_reference: osr.SpatialReference = source.GetSpatialRef()
+
+    # generate mask data
+    mask_data = np.where(source_data == 0, False, True)
+    mask_ds, mask_band = _create_in_memory_band(
+        mask_data, source_band.XSize, source_band.YSize, source.GetProjection(),
+        source.GetGeoTransform())
+
+    # Create a memory OGR datasource to put results in.
+    mem_drv: ogr.Driver = ogr.GetDriverByName("Memory")
```
andrea-williams, line 39: Is there a diff between using the "Memory" driver and the "MEM" driver in line 12?

andrea-williams (commented):
Still working through the review, just saving my progress so far...

api/app/autoneal/process_hfi.py
```diff
@@ -0,0 +1,89 @@
...
+    """
+    logger.info('Processing HFI %s for run date: %s, for date: %s', run_type, run_date, for_date)
+    perf_start = perf_counter()
+
+    bucket = config.get('OBJECT_STORE_BUCKET')
+    # TODO what really has to happen, is that we grab the most recent prediction for the given date,
+    # but this method doesn't even belong here, it's just a shortcut for now!
+    for_date_string = f'{for_date.year}{for_date.month:02d}{for_date.day:02d}'
+
+    key = f'/vsis3/{bucket}/sfms/uploads/{run_type.value}/{run_date.isoformat()}/hfi{for_date_string}.tif'
```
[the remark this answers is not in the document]
Sybrand: The hfi is part if the filename as given to us by sfms. Right now we're just dumping all the outputs for a given run date into one folder. Adding a slash would imply a change there, and more processing. We;d end up with a redundant date, e.g. `{run_date}/{for_date}/hfi{for_date}.tiff` or have to rename the file as we process it, which would then stop mirroring the output. We'd also have a seperate folder for each type of output, which I don't think actually adds much clarity.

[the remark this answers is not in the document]
Sybrand: visis3: https://gdal.org/user/virtual_file_systems.html - I'll add the link into the code.

api/alembic/versions/17b1c787f420_advisory_areas.py
```diff
@@ -52,5 +52,5 @@ def downgrade():
     op.drop_table('advisory_shapes')
     op.drop_index(op.f('ix_advisory_shape_types_name'), table_name='advisory_shape_types')
     op.drop_table('advisory_shape_types')
-    sa.Enum(name='shapetypeenum').drop(op.get_bind())
     # ### end Alembic commands ###
+    sa.Enum(name='shapetypeenum').drop(op.get_bind())
```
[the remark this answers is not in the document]
Sybrand: Purely because it isn't part of he auto generated alembic stuff. For some reason alembic doesn't clean up postgres types.

api/alembic/versions/2cc32257e8bc_rethink_classification.py
```diff
@@ -0,0 +1,71 @@
...
+# revision identifiers, used by Alembic.
+revision = '2cc32257e8bc'
+down_revision = 'ef2482f08074'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic ###
+    op.create_table('advisory_classified_hfi',
```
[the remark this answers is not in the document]
Sybrand: Current is `HFI classification for some forecast/advisory run on some day, for some date` - I'm lacking imagination, and suggestions?

api/app/db/models/advisory.py
```diff
@@ -54,8 +44,30 @@ class Shape(Base):
...
+    """ HFI classified into different groups.
+    NOTE: In actual fact, forecasts and actuals can be run multiple times per day,
+    but we only care about the most recent one, so we only store the date, not the timesamp.
+    """
+    __tablename__ = 'advisory_classified_hfi'
+    __table_args__ = (
+        {'comment': 'HFI classification for some forecast/advisory run on some day, for some date'}
+    )
+    id = Column(Integer, primary_key=True, index=True)
+    # TODO: we could do this better!
```
andrea-williams, line 64: Maybe briefly explain *how* this could be done better?

api/app/main.py
```diff
@@ -107,7 +107,7 @@ async def catch_exception_middleware(request: Request, call_next):
 api.include_router(stations.router, tags=["Stations"])
 api.include_router(hfi_calc.router, tags=["HFI"])
 api.include_router(fba_calc.router, tags=["FBA Calc"])
-api.include_router(fba.router, tags=["FBA"])
+api.include_router(fba.router, tags=["Fire Behaviour Advisory"])
```
andrea-williams, line 110:
```suggestion
api.include_router(fba.router, tags=["Auto Spatial Advisory"])
```

andrea-williams: ?

api/app/main.py
```diff
@@ -107,7 +107,7 @@ async def catch_exception_middleware(request: Request, call_next):
...
 api.include_router(fwi_calc.router, tags=["FWI"])
 api.include_router(sfms.router, tags=["SFMS", "Fire Behaviour Advisory"])
```
andrea-williams, line 112:
```suggestion
api.include_router(sfms.router, tags=["SFMS", "Auto Spatial Advisory"])
```

api/app/routers/fba.py
```diff
@@ -1,14 +1,17 @@
 """ Routers for new/formal/non-tinker fba.
```
andrea-williams, line 1:
```suggestion
""" Routers for Auto Spatial Advisory
```

andrea-williams: Ugh there's too many references to "fba" etc. I'm giving up on this now. It's beyond the scope of this ticket anyway.

andrea-williams (commented):
Saving progress again...

api/alembic/versions/2cc32257e8bc_rethink_classification.py
Sybrand: Good catch - the Index in the model file is wrong! I'll try fix it.

api/app/auto_spatial_advisory/classify_hfi.py
Sybrand: It's going to have different meanings for different rasters - so I don't know about that. In this case, we're classifying into three sets 0. < 4000 1. >=4000 and <10000, 2. >=10000
With FTL (or other things) 0 may have some different meaning.

api/app/auto_spatial_advisory/classify_hfi.py
Sybrand: I don't like setting it to None, because you're assign a value to a variable that you're never going to use. The intention is to trigger garbage collection (because gdal is really bad for holding on to memory), so `None` is going to have the same effect as `del` in that respect. I'm just coming from it from the perspective that assigning a value to a variable, implies you're going to use that variable, which is not the case here. We're not doing any None checks.

andrea-williams (approved):
Gonna riskily approve this bc I haven't tested out the DB migrations on my local but I suspect it's going to cause a lot of headaches trying to resolve with the current state of my local DB. And I don't feel like dealing with it right now.

api/app/auto_spatial_advisory/polygonize.py
Sybrand: You're making me go DEEP into the horrid documentation. 

`gdal.GetDriverByName('MEM')` is asking for a in memory raster driver: https://gdal.org/drivers/raster/mem.html

If you do:
`
mem_driver = gdal.GetDriverByName('MEM')
dataset = mem_driver.Create('memory', cols, rows, 1, gdal.GDT_Byte)
`
then `memory` is just a name - no meaning.

If you do
`ogr.GetDriverByName("Memory")` 
then it's asking for an in memory vector driver: https://gdal.org/drivers/vector/memory.html#vector-memory

Sybrand: I've added comments.

api/app/db/models/advisory.py
Sybrand: I've added a comment. It's basically around not using a string. We could fix this now if you want? I'm feeling a bit non-committal at the moment, but normalizing the database is probably a good idea. Should I do it?

Sybrand: Going to normalize this...

Sybrand: Normalized.

Sybrand:
> Gonna riskily approve this bc I haven't tested out the DB migrations on my local but I suspect it's going to cause a lot of headaches trying to resolve with the current state of my local DB. And I don't feel like dealing with it right now.

@andrea-williams I've address all your comments.
@conbrad have made a fair amount of changes following Andrea's review, if you want to relook.

api/app/autoneal/process_hfi.py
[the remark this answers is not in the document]
conbrad: My bad, thought it was a path typo

api/alembic/versions/fbca68ccc9da_classification_thresholds.py
```diff
@@ -0,0 +1,47 @@
...
+classification_threshold_table = sa.Table(
+    'advisory_hfi_classification_threshold',
+    sa.MetaData(),
+    sa.Column('id', sa.Integer(), nullable=False),
+    sa.Column('description', sa.String(), nullable=False),
+    sa.Column('name', sa.String(), nullable=False),
+)
+
+
+def upgrade():
```
conbrad, line 28: This migration takes awhile to grab everything from arcserver, will this migration timeout when we deploy?

conbrad (approved):

api/alembic/versions/fbca68ccc9da_classification_thresholds.py
Sybrand: It's never happened to me - it's certainly possible - you'd have to re-run the migration if it fails.

PR #2645, opened 2023-02-14
Morecast v2: Adjusted API schema, added db schema

conbrad:
- Changes timestamp field in API schema to `for_date` for clarity
- Adds `MorecastForecast` DB schema and migration


andrea-williams (approved):

api/app/db/models/morecast_v2.py
```diff
@@ -0,0 +1,22 @@
...
+    for_date = Column(TZTimeStamp, nullable=False, index=True)
+    temp = Column(Integer, nullable=False, index=True)
+    rh = Column(Integer, nullable=False, index=True)
+    precip = Column(Integer, nullable=False, index=True)
+    wind_speed = Column(Integer, nullable=False, index=True)
+    wind_direction = Column(Integer, nullable=False, index=True)
+    create_timestamp = Column(TZTimeStamp, nullable=False, index=True)
+    create_user = Column(String, nullable=False)
+    update_timestamp = Column(TZTimeStamp, nullable=False, index=True)
+    update_user = Column(String, nullable=False)
```
dgboss, line 22: What is this field for?

dgboss (approved):

api/app/db/models/morecast_v2.py
conbrad: It's to capture the user who updated it last, assuming we allow updates to forecasts. It also implies an assumption that the creator and updater can differ but I'm not sure that's true.

PR #2709, opened 2023-03-09
Morecast2 DatePicker

conbrad:
- Pull out and make DateRangePicker generic from PrepRangeDatePicker
- Have Morecast2 use the above 
- @dgboss 's fix for the extra day caused by millisecond offset
# Test Links:
[Landing Page](https://wps-pr-2709.apps.silver.devops.gov.bc.ca/)
[MoreCast 2.0](https://wps-pr-2709.apps.silver.devops.gov.bc.ca/morecast-2)
[Percentile Calculator](https://wps-pr-2709.apps.silver.devops.gov.bc.ca/percentile-calculator)
[MoreCast](https://wps-pr-2709.apps.silver.devops.gov.bc.ca/morecast)
[C-Haines](https://wps-pr-2709.apps.silver.devops.gov.bc.ca/c-haines)
[FireBat](https://wps-pr-2709.apps.silver.devops.gov.bc.ca/fire-behaviour-calculator)
[FireBat bookmark](https://wps-pr-2709.apps.silver.devops.gov.bc.ca/fire-behaviour-calculator?s=266&f=c5&c=NaN&w=20,s=286&f=c7&c=NaN&w=16,s=1055&f=c7&c=NaN&w=NaN,s=305&f=c7&c=NaN&w=NaN,s=344&f=c5&c=NaN&w=NaN,s=346&f=c7&c=NaN&w=NaN,s=328&f=c7&c=NaN&w=NaN,s=1399&f=c7&c=NaN&w=NaN,s=334&f=c7&c=NaN&w=NaN,s=1082&f=c3&c=NaN&w=NaN,s=388&f=c7&c=NaN&w=NaN,s=309&f=c7&c=NaN&w=16,s=306&f=c7&c=NaN&w=NaN,s=1029&f=c7&c=NaN&w=NaN,s=298&f=c7&c=NaN&w=NaN,s=836&f=c7&c=NaN&w=NaN,s=9999&f=c7&c=NaN&w=NaN)
[Auto Spatial Advisory (ASA)](https://wps-pr-2709.apps.silver.devops.gov.bc.ca/auto-spatial-advisory)
[HFI Calculator](https://wps-pr-2709.apps.silver.devops.gov.bc.ca/hfi-calculator)


web/src/components/dateRangePicker/DateRangePickerWrapper.tsx
```diff
@@ -30,14 +31,14 @@ const DateRangePickerWrapper: React.FunctionComponent<DateRangePickerWrapperProp
   const handleKeyPress = (event: any) => event?.key === 'Escape' && handleToggle()
 
   const minDate = initialDateRange.startDate || new Date()
-  const maxDate = DateTime.fromJSDate(minDate).plus({ days: 6 }).toJSDate()
+  const maxDate = DateTime.fromJSDate(minDate).plus({ days: 11 }).toJSDate()
```
andrea-williams, line 34: Should technically be plus 10 days. The 11 days specified in the ticket counts Day 1 as today.

andrea-williams (changes requested):
The same off-by-1 date bug is still appearing. See screenshot (this happened when I loaded the page and selected a fire centre at random). Date range ends at the 11th, but table includes the 12th.

Screenshot 2023-03-09 at 12 02 21 PM

web/src/components/dateRangePicker/DateRangePickerWrapper.tsx
conbrad: The date range is inclusive so this is already the case

conbrad:
> The same off-by-1 date bug is still appearing. See screenshot (this happened when I loaded the page and selected a fire centre at random). Date range ends at the 11th, but table includes the 12th.
> 
> Screenshot 2023-03-09 at 12 02 21 PM

I'm not able to reproduce this

web/src/components/dateRangePicker/DateRangePickerWrapper.tsx
andrea-williams: Screenshot 2023-03-09 at 12 16 25 PM

Load page -> open datepicker without changing the dates selected - the 12th day is clickable when it shouldn't be. But really I'm not worried about this one so much, but the extra day showing up repeatedly in the table with N/A every time needs to be fixed

conbrad: Again, not able to reproduce...

andrea-williams: Are you still able to reproduce the extra-day-in-the-table bug @dgboss ?

conbrad: Day offset fixed

conbrad:
> MoreCast 2.0

Ok managed to get it on my machine -- first load, as soon as I open debug tools or refresh it goes away...

web/src/features/moreCast2/pages/MoreCast2Page.tsx
```diff
@@ -82,8 +83,10 @@ const MoreCast2Page = () => {
   const [modelType, setModelType] = useState<ModelType>(
     (localStorage.getItem(DEFAULT_MODEL_TYPE_KEY) as ModelType) || DEFAULT_MODEL_TYPE
   )
-  const [fromDate, setFromDate] = useState<DateTime>(DateTime.now())
-  const [toDate, setToDate] = useState<DateTime>(DateTime.now().plus({ days: 2 }))
+  const [fromTo, setFromTo] = useState<DateRange>({
+    startDate: DateTime.now().toJSDate(),
+    endDate: DateTime.now().plus({ days: 2 }).toJSDate()
```
dgboss, line 88: I think the off by one day bug is due to the initialization of `startDate` and `endDate` with `DateTime.now()`. `endDate` needs to be initialized exactly 2 days after `startDate` (to the millisecond).

web/src/components/dateRangePicker/DateRangePickerWrapper.tsx
dgboss: Yes, I can reproduce. I added a comment describing where it can be fixed (initialization of the endDate).

dgboss:
I kind of feel like we need a close or OK button. It might not be obvious to all users that clicking away from the calendar closes and 'saves' the date range. I do realize that the state is updated immediately after you select the end date, but this isn't necessarily obvious.

web/src/features/moreCast2/pages/MoreCast2Page.tsx
conbrad: Wow, nice catch!

web/src/features/moreCast2/pages/MoreCast2Page.tsx
```diff
@@ -82,8 +83,13 @@ const MoreCast2Page = () => {
   const [modelType, setModelType] = useState<ModelType>(
     (localStorage.getItem(DEFAULT_MODEL_TYPE_KEY) as ModelType) || DEFAULT_MODEL_TYPE
   )
-  const [fromDate, setFromDate] = useState<DateTime>(DateTime.now())
-  const [toDate, setToDate] = useState<DateTime>(DateTime.now().plus({ days: 2 }))
+
+  const startDateTime = DateTime.now().startOf('day')
```
dgboss, line 87: This will probably work, but the following logic might be easier to understand.
```  const startDateTime = DateTime.now()
  const endDateTime = startDateTime.plus({ days: 2 })
  const [fromTo, setFromTo] = useState<DateRange>({
    startDate: startDateTime.toJSDate(),
    endDate: endDateTime.toJSDate()
  })

File: web/src/components/DateRangeSelector.tsx [hunk 4]
```diff
@@ -0,0 +1,68 @@
...
+        data-testid="date-range-picker-text-field"
+        size={size}
+        variant="outlined"
+        disabled={true}
+        label={label}
+        onClick={() => setDateRangePickerOpen(!dateRangePickerOpen)}
+        value={
+          isUndefined(dateRange) || isUndefined(dateRange.startDate) || isUndefined(dateRange.endDate)
+            ? ''
+            : `${DateTime.fromJSDate(dateRange.startDate).toFormat(dateDisplayFormat).trim()} - ${DateTime.fromJSDate(
```
dgboss, line 39: Curious about .trim(). Does `toFormat()` add whitespace or line terminations? Or is the concern that the `dateDisplayFormat` might introduce these?

conbrad: No idea, I can remove it

conbrad: Removed `trim`, I didn't see any difference in how the string renders, it was probably a historical defensive measure.

dgboss:
I'm seeing this in my Chrome dev tools:
image

web/src/components/dateRangePicker/DateRangePickerWrapper.tsx
```diff
@@ -30,14 +31,14 @@ const DateRangePickerWrapper: React.FunctionComponent<DateRangePickerWrapperProp
   const handleKeyPress = (event: any) => event?.key === 'Escape' && handleToggle()
 
   const minDate = initialDateRange.startDate || new Date()
-  const maxDate = DateTime.fromJSDate(minDate).plus({ days: 6 }).toJSDate()
+  const maxDate = DateTime.fromJSDate(minDate).plus({ days: 10 }).toJSDate()
 
   return (
     <div data-testid="date-range-picker-wrapper">
-      {open && <div onKeyPress={handleKeyPress} onClick={handleToggle} />}
+      {open && <div onKeyDown={handleKeyPress} onClick={handleToggle} />}
```
dgboss, line 38: I can't seem to hit the `onClick` handler for this div, probably because it has no height. I think that can be removed. Could the `onKeyDown` handler moved to the parent div and have `handleKeyPress` include the check for `open`
`const handleKeyPress = (event: any) => open && event?.key === 'Escape' && handleToggle()`

conbrad: Nice call, fixed in: https://github.com/bcgov/wps/pull/2709/commits/99a3d48c9a38831fd667f991eb2228145048e384

conbrad:
> I'm seeing this in my Chrome dev tools: image

Fixed in: https://github.com/bcgov/wps/pull/2709/commits/99a3d48c9a38831fd667f991eb2228145048e384
Also fixes a warning for equality checks in an autocomplete component

dgboss:
Looks great!

dgboss (approved):

PR #2909, opened 2023-05-30
Unify headers

dgboss:


conbrad (approved):
🚀

andrea-williams (approved):
Morecast v1 doesn't have a Beta label in the page header, but does still have the Beta label on our landing page. I think we want to remove that label from the landing page, rather than adding it to the header. (Just guessing, though).

Otherwise looks good!

dgboss:
> 

Good catch!

PR #2934, opened 2023-06-13
upgrade react

jmetev1:
upgrade react plus changes forced by that


jmetev1:
remade on non forked branch

PR #2936, opened 2023-06-13
upgrade react

jmetev1:
upgraded react and made some changes required by new react
# Test Links:
[Landing Page](https://wps-pr-2936.apps.silver.devops.gov.bc.ca/)
[MoreCast 2.0](https://wps-pr-2936.apps.silver.devops.gov.bc.ca/morecast-2)
[Percentile Calculator](https://wps-pr-2936.apps.silver.devops.gov.bc.ca/percentile-calculator)
[MoreCast](https://wps-pr-2936.apps.silver.devops.gov.bc.ca/morecast)
[C-Haines](https://wps-pr-2936.apps.silver.devops.gov.bc.ca/c-haines)
[FireBat](https://wps-pr-2936.apps.silver.devops.gov.bc.ca/fire-behaviour-calculator)
[FireBat bookmark](https://wps-pr-2936.apps.silver.devops.gov.bc.ca/fire-behaviour-calculator?s=266&f=c5&c=NaN&w=20,s=286&f=c7&c=NaN&w=16,s=1055&f=c7&c=NaN&w=NaN,s=305&f=c7&c=NaN&w=NaN,s=344&f=c5&c=NaN&w=NaN,s=346&f=c7&c=NaN&w=NaN,s=328&f=c7&c=NaN&w=NaN,s=1399&f=c7&c=NaN&w=NaN,s=334&f=c7&c=NaN&w=NaN,s=1082&f=c3&c=NaN&w=NaN,s=388&f=c7&c=NaN&w=NaN,s=309&f=c7&c=NaN&w=16,s=306&f=c7&c=NaN&w=NaN,s=1029&f=c7&c=NaN&w=NaN,s=298&f=c7&c=NaN&w=NaN,s=836&f=c7&c=NaN&w=NaN,s=9999&f=c7&c=NaN&w=NaN)
[Auto Spatial Advisory (ASA)](https://wps-pr-2936.apps.silver.devops.gov.bc.ca/auto-spatial-advisory)
[HFI Calculator](https://wps-pr-2936.apps.silver.devops.gov.bc.ca/hfi-calculator)


conbrad:
Thanks @jmetev1 , looks like here's where further compilation errors are showing up, are you comfortable running/debugging cypress tests?

https://github.com/bcgov/wps/actions/runs/5261144500/jobs/9508938748?pr=2936

jmetev1:
this was just kind of a placeholder. i was going to merge in main after merging my other pr which will take care of those ts errors.

conbrad:
> this was just kind of a placeholder. i was going to merge in main after merging my other pr which will take care of those ts errors.

Perfect, makes sense, thanks.

dgboss:
I'm seeing infinite re-renders on every route that uses the AuthWrapper. It looks like the `init` function in keycloak-js doesn't like being called more than once. We might be able to guard against the second call to `init` in a manner similar to this post: https://github.com/keycloak/keycloak/issues/19452#issuecomment-1503483795.

We'd have to place the ref and check in `AuthWrapper.tsx` and use it to prevent a second call to `dispatch(authenticate())`.

conbrad:
> I'm seeing infinite re-renders on every route that uses the AuthWrapper. It looks like the `init` function in keycloak-js doesn't like being called more than once. We might be able to guard against the second call to `init` in a manner similar to this post: [keycloak/keycloak#19452 (comment)](https://github.com/keycloak/keycloak/issues/19452#issuecomment-1503483795).
> 
> We'd have to place the ref and check in `AuthWrapper.tsx` and use it to prevent a second call to `dispatch(authenticate())`.

Oh interesting! Yes it looks like we may have intended keycloak to be a singleton based on the function name `getKeycloakInstance` but that is not what it's doing.

dgboss:
Temporarily closing to free resources

dgboss:
.

jmetev1:
is it related to react 18 and strict mode and the way it renders twice? https://stackoverflow.com/questions/72489140/react-18-strict-mode-causing-component-to-render-twice 
I can't use auth so i can't see these kind of problems. Feel free to push any commits if someone wants to change anything.

dgboss:
> is it related to react 18 and strict mode and the way it renders twice? https://stackoverflow.com/questions/72489140/react-18-strict-mode-causing-component-to-render-twice I can't use auth so i can't see these kind of problems. Feel free to push any commits if someone wants to change anything.

The issue is related to the double rendering with strict mode, but not exactly a problem with React 18. We use keycloak-js to handle some of our authorization logic. The keycloak instance is supposed to be a singleton and only supports calling the `init()` function once. I've pushed out a fix for this issue, but we have a bigger challenge if we move forward with this PR. i'll add another comment with a screenshot.

dgboss:
After dealing with the re-rendering issue in https://github.com/bcgov/wps/pull/2936/commits/31bedda2916573dd41707da1d543e79f6299a6e0, I am seeing multiple styling issues. Here's an example of what I'm seeing for the header in the Percentile Calculator (no auth required).

image

The expected appearance can be seen at https://psu.nrs.gov.bc.ca/percentile-calculator.

The root of the issue is that @mui v5 moved away from JSS based styling towards the use of [emotion](https://emotion.sh/docs/introduction). @mui v5 maintained backwards compatibility for JSS styling with React 17, but this is not compatible with React 18. 

Details about these changes can be found at https://mui.com/material-ui/migration/migration-v4/.

Our apps rely heavily on JSS based styling patterns (ie `import { makeStyles} from '@mui/styles'`) which are not supported by React 18, which is why I'm seeing styling issues.

I'll chat with @conbrad about how to move forward.

conbrad:
> After dealing with the re-rendering issue in [31bedda](https://github.com/bcgov/wps/commit/31bedda2916573dd41707da1d543e79f6299a6e0), I am seeing multiple styling issues. Here's an example of what I'm seeing for the header in the Percentile Calculator (no auth required).
> 
> image
> 
> The expected appearance can be seen at https://psu.nrs.gov.bc.ca/percentile-calculator.
> 
> The root of the issue is that @mui v5 moved away from JSS based styling towards the use of [emotion](https://emotion.sh/docs/introduction). @mui v5 maintained backwards compatibility for JSS styling with React 17, but this is not compatible with React 18.
> 
> Details about these changes can be found at https://mui.com/material-ui/migration/migration-v4/.
> 
> Our apps rely heavily on JSS based styling patterns (ie `import { makeStyles} from '@mui/styles'`) which are not supported by React 18, which is why I'm seeing styling issues.
> 
> I'll chat with @conbrad about how to move forward.

Good find, this makes sense. We "halfway" migrated to material ui v5 but never quite finished it. Looks like we should prioritize this task next: https://app.zenhub.com/workspaces/bcws---agile-psu-5e321393e038fba5bbe203b8/issues/gh/bcgov/wps/1898

I've also added the react upgrade as blocked by the above.

web/src/features/auth/components/AuthWrapper.tsx
```diff
@@ -25,11 +25,13 @@ const setAxiosRequestInterceptors = (): AppThunk => (_, getState) => {
 const AuthWrapper = ({ children }: Props) => {
   const dispatch: AppDispatch = useDispatch()
   const { isAuthenticated, authenticating, error } = useSelector(selectAuthentication)
+  const didInit = useRef(false)
 
   useEffect(() => {
     if (TEST_AUTH || window.Cypress) {
       dispatch(testAuthenticate(true, 'test token', 'test id token'))
-    } else {
+    } else if (!didInit.current) {
```
conbrad, line 33: We may want to consider a singleton pattern less dependent on react useRef, something like: https://refactoring.guru/design-patterns/singleton/typescript/example

Something like we did with the `MoreCast2Column`

PR #2997, opened 2023-07-13
Bug/asa map overflow

brettedw:
Fixed map overflow issues and added a scroll bar to the side panel
# Test Links:
[Landing Page](https://wps-pr-2997.apps.silver.devops.gov.bc.ca/)
[MoreCast 2.0](https://wps-pr-2997.apps.silver.devops.gov.bc.ca/morecast-2)
[Percentile Calculator](https://wps-pr-2997.apps.silver.devops.gov.bc.ca/percentile-calculator)
[MoreCast](https://wps-pr-2997.apps.silver.devops.gov.bc.ca/morecast)
[C-Haines](https://wps-pr-2997.apps.silver.devops.gov.bc.ca/c-haines)
[FireBat](https://wps-pr-2997.apps.silver.devops.gov.bc.ca/fire-behaviour-calculator)
[FireBat bookmark](https://wps-pr-2997.apps.silver.devops.gov.bc.ca/fire-behaviour-calculator?s=266&f=c5&c=NaN&w=20,s=286&f=c7&c=NaN&w=16,s=1055&f=c7&c=NaN&w=NaN,s=305&f=c7&c=NaN&w=NaN,s=344&f=c5&c=NaN&w=NaN,s=346&f=c7&c=NaN&w=NaN,s=328&f=c7&c=NaN&w=NaN,s=1399&f=c7&c=NaN&w=NaN,s=334&f=c7&c=NaN&w=NaN,s=1082&f=c3&c=NaN&w=NaN,s=388&f=c7&c=NaN&w=NaN,s=309&f=c7&c=NaN&w=16,s=306&f=c7&c=NaN&w=NaN,s=1029&f=c7&c=NaN&w=NaN,s=298&f=c7&c=NaN&w=NaN,s=836&f=c7&c=NaN&w=NaN,s=9999&f=c7&c=NaN&w=NaN)
[Auto Spatial Advisory (ASA)](https://wps-pr-2997.apps.silver.devops.gov.bc.ca/auto-spatial-advisory)
[HFI Calculator](https://wps-pr-2997.apps.silver.devops.gov.bc.ca/hfi-calculator)

### DIFF
--- a/web/src/features/fba/components/ZoneSummaryPanel.tsx
+++ b/web/src/features/fba/components/ZoneSummaryPanel.tsx
@@ -45,23 +45,23 @@ const ZoneSummaryPanel = React.forwardRef((props: Props, ref: React.ForwardedRef
     return (
       <Container ref={ref} className={`${props.className} ${classes.wrapper}`}>
         <Grid container alignItems={'center'} direction={'column'}>
-          <div>
+          <Grid item>
             <Typography className={classes.zoneName}>{props.selectedFireZone.mof_fire_zone_name}</Typography>
             <Typography className={classes.centreName}>{props.selectedFireZone.mof_fire_centre_name}</Typography>
-          </div>
-          <div>
+          </Grid>
+          <Grid item>
             <CombustibleAreaViz
               fireZoneAreas={props.fireZoneAreas.filter(
                 area => area.mof_fire_zone_id == props.selectedFireZone?.mof_fire_zone_id
               )}
             />
-          </div>
-          <div>
+          </Grid>
+          <Grid item>
             <FuelTypesBreakdown selectedFireZone={props.selectedFireZone} fuelTypeInfo={props.fuelTypeInfo} />
-          </div>
-          <div>
+          </Grid>
+          <Grid item>
             <ElevationInfoViz selectedFireZone={props.selectedFireZone} hfiElevationInfo={props.hfiElevationInfo} />
-          </div>
+          </Grid>
         </Grid>
       </Container>
     )
--- a/web/src/features/fba/components/ZoneSummaryPanel.tsx
+++ b/web/src/features/fba/components/ZoneSummaryPanel.tsx
@@ -1,7 +1,7 @@
 import React from 'react'
 import makeStyles from '@mui/styles/makeStyles'
 import CombustibleAreaViz from 'features/fba/components/viz/CombustibleAreaViz'
-import { Grid, Typography } from '@mui/material'
+import { Container, Typography } from '@mui/material'
 import { isUndefined } from 'lodash'
 import { ElevationInfoByThreshold, FireZone, FireZoneArea, FireZoneThresholdFuelTypeArea } from 'api/fbaAPI'
 import ElevationInfoViz from 'features/fba/components/viz/ElevationInfoViz'
@@ -34,40 +34,41 @@ interface Props {
   fireZoneAreas: FireZoneArea[]
 }
 
-const ZoneSummaryPanel = (props: Props) => {
+const ZoneSummaryPanel = React.forwardRef((props: Props, ref: any) => {
   const classes = useStyles()
+  ZoneSummaryPanel.displayName = 'ZoneSummaryPanel'
 
   if (isUndefined(props.selectedFireZone)) {
     return <div></div>
   } else {
     return (
-      <Grid
-        container
+      <Container
+        ref={ref}
         alignItems={'center'}
         direction={'column'}
         spacing={2}
         className={`${props.className} ${classes.wrapper}`}
       >
-        <Grid item>
+        <div>
           <Typography className={classes.zoneName}>{props.selectedFireZone.mof_fire_zone_name}</Typography>
           <Typography className={classes.centreName}>{props.selectedFireZone.mof_fire_centre_name}</Typography>
-        </Grid>
-        <Grid item>
+        </div>
+        <div>
           <CombustibleAreaViz
             fireZoneAreas={props.fireZoneAreas.filter(
               area => area.mof_fire_zone_id == props.selectedFireZone?.mof_fire_zone_id
             )}
           />
-        </Grid>
-        <Grid item>
+        </div>
+        <div>
           <FuelTypesBreakdown selectedFireZone={props.selectedFireZone} fuelTypeInfo={props.fuelTypeInfo} />
-        </Grid>
-        <Grid item>
+        </div>
+        <div>
           <ElevationInfoViz selectedFireZone={props.selectedFireZone} hfiElevationInfo={props.hfiElevationInfo} />
-        </Grid>
-      </Grid>
+        </div>
+      </Container>
     )
   }
-}
+})
 
 export default React.memo(ZoneSummaryPanel)
--- a/web/src/features/fba/components/ZoneSummaryPanel.tsx
+++ b/web/src/features/fba/components/ZoneSummaryPanel.tsx
@@ -35,7 +35,7 @@ interface Props {
   fireZoneAreas: FireZoneArea[]
 }
 
-const ZoneSummaryPanel = React.forwardRef((props: Props, ref: any) => {
+const ZoneSummaryPanel = React.forwardRef((props: Props, ref: React.ForwardedRef<HTMLDivElement>) => {
   const classes = useStyles()
   ZoneSummaryPanel.displayName = 'ZoneSummaryPanel'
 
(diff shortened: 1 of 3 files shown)


web/src/features/fba/components/ZoneSummaryPanel.tsx
```diff
@@ -34,40 +35,37 @@ interface Props {
   fireZoneAreas: FireZoneArea[]
 }
 
-const ZoneSummaryPanel = (props: Props) => {
+const ZoneSummaryPanel = React.forwardRef((props: Props, ref: any) => {
```
dgboss, line 38: I think you can use `React.ForwardedRef<HTMLDivElement>` as the type instead of `any`.

brettedw: > 

Thanks, good call

web/src/features/fba/pages/FireBehaviourAdvisoryPage.tsx
```diff
@@ -152,17 +154,55 @@ const FireBehaviourAdvisoryPage: React.FunctionComponent = () => {
...
+  const sidePanelRef = useRef<HTMLDivElement>(null)
+  const [formControlHeight, setFormControlHeight] = useState<number>(0)
+  const [navRefHeight, setNavRefHeight] = useState<number>(0)
+
+  useEffect(() => {
+    if (navRef.current) {
+      const height = navRef.current.clientHeight
+      setNavRefHeight(height)
+    } // eslint-disable-line react-hooks/exhaustive-deps
+  })
```
dgboss, line 169: I led you astray a little bit here. We can gain some minor efficiency adding a dependency to this useEffect hook. We can't use an empty `[]` as that will only run on the first render, but we should be able to use`[navRef.current?.clientHeight]` so that this hook only gets called when the nav height actually changes instead of at each render. You can do the same thing below with the useEffect hook that set the `formControlHeight.

web/src/features/fba/pages/FireBehaviourAdvisoryPage.tsx
```diff
@@ -152,17 +154,55 @@ const FireBehaviourAdvisoryPage: React.FunctionComponent = () => {
...
+  }, [navRef.current?.clientHeight])
+
+  useEffect(() => {
+    if (formControlRef.current) {
+      const height = formControlRef.current.clientHeight
+      setFormControlHeight(height)
+    } // eslint-disable-line react-hooks/exhaustive-deps
+  }, [formControlRef.current?.clientHeight])
+
+  useEffect(() => {
```
conbrad, line 178: I think this `useEffect` can be merged with the one below it since they both don't specify a dependency array. Also is the intention for this to run once or on every render? If the former we should specify an empty dependency array: https://stackoverflow.com/a/57760937

conbrad (commented):
Looking really good! Just one comment, more of a question

web/src/features/fba/pages/FireBehaviourAdvisoryPage.tsx
brettedw: I can refactor it a bit to make it cleaner. My intention was to have that effect to run on every render, it was the only way I could get the map and panel to resize correctly when the form control changes sizes. 
I started with an empty dependency array which I quickly discovered didn't work with the way the page functions, I'm definitely open to there being a better way to do it though

conbrad: Got it! Makes sense thanks!

web/src/features/fba/components/ZoneSummaryPanel.tsx
```diff
@@ -34,40 +35,37 @@ interface Props {
...
-        direction={'column'}
-        spacing={2}
-        className={`${props.className} ${classes.wrapper}`}
-      >
-        <Grid item>
-          <Typography className={classes.zoneName}>{props.selectedFireZone.mof_fire_zone_name}</Typography>
-          <Typography className={classes.centreName}>{props.selectedFireZone.mof_fire_centre_name}</Typography>
+      <Container ref={ref} className={`${props.className} ${classes.wrapper}`}>
+        <Grid container alignItems={'center'} direction={'column'}>
+          <div>
```
conbrad, line 48: We'll want to maintain the existing `GridItem` components here instead of `div`s, it'll give us a bit more responsive flexibility: https://mui.com/material-ui/react-grid/

brettedw: Ok, I changed those to `div`s after talking with Darren about a display issue, but I think I've found a better way and can change them back to `Grid` items now

web/src/features/fba/components/ZoneSummaryPanel.tsx
```diff
@@ -34,40 +35,37 @@ interface Props {
   fireZoneAreas: FireZoneArea[]
 }
 
-const ZoneSummaryPanel = (props: Props) => {
+const ZoneSummaryPanel = React.forwardRef((props: Props, ref: React.ForwardedRef<HTMLDivElement>) => {
   const classes = useStyles()
+  ZoneSummaryPanel.displayName = 'ZoneSummaryPanel'
```
conbrad, line 40: I think we can delete this
```suggestion
```

conbrad (approved):
Nice job! 🚀

web/src/features/fba/components/ZoneSummaryPanel.tsx
brettedw: I wasn't confident about what I did here, but it was a solution to prevent the linter complaining hard about the Component definition missing a display name, which from what I've read could be helpful for debugging?

Another solution would be add a `/* eslint-disable react/display-name */` 

Or maybe there's a better solution...

conbrad: Huh weird, nevermind I think this is correct. It's expected by React forward ref components I think

web/src/features/fba/pages/FireBehaviourAdvisoryPage.tsx
```diff
@@ -211,13 +245,15 @@ const FireBehaviourAdvisoryPage: React.FunctionComponent = () => {
...
+              ref={sidePanelRef}
               selectedFireZone={selectedFireZone}
               fuelTypeInfo={hfiThresholdsFuelTypes}
               hfiElevationInfo={fireZoneElevationInfo}
               fireZoneAreas={fireZoneAreas}
+              className={classes.scrollablePanel}
             />
           </Grid>
-          <Grid className={classes.flex} item>
+          <Grid className={classes.flex} ref={mapRef} item>
```
dgboss, line 256: Continuing on from my comment above about an alternate approach to setting the height of the component.

```
<Grid
  className={classes.flex}
  sx={{height: `calc(100vh - ${formControlHeight + navRefHeight}px)`}}
  item
>
```

web/src/features/fba/pages/FireBehaviourAdvisoryPage.tsx
```diff
@@ -152,17 +154,49 @@ const FireBehaviourAdvisoryPage: React.FunctionComponent = () => {
...
   return (
     <div className={classes.root}>
-      <GeneralHeader
-        isBeta={true}
-        spacing={1}
-        title={FIRE_BEHAVIOUR_ADVISORY_NAME}
-        productName={FIRE_BEHAVIOUR_ADVISORY_NAME}
-      />
+      <Container disableGutters maxWidth={'xl'}>
+        <GeneralHeader
```
dgboss, line 189: Continuing on from my comment above about an alternate approach to setting the height of the component.

You could add a `height` prop to the `<GeneralHeader>`. The value would be `calc(100vh - ${formControlHeight + navRefHeight}px)` and this could be assigned to the top level element of the GeneralHeader component like this ```<Container sx={{height: props.height}} className=`${props.className} ${classes.wrapper}`>```

web/src/features/fba/pages/FireBehaviourAdvisoryPage.tsx
```diff
@@ -152,17 +154,49 @@ const FireBehaviourAdvisoryPage: React.FunctionComponent = () => {
...
+      setFormControlHeight(formControlRef.current.clientHeight)
+    }
+  }, [formControlRef.current?.clientHeight])
+
+  useEffect(() => {
+    const sidePanelElement = sidePanelRef.current
+    const mapElement = mapRef.current
+    if (sidePanelElement && mapElement && formControlHeight && navRefHeight) {
+      const height = `calc(100vh - ${formControlHeight + navRefHeight}px)`
+      sidePanelElement.style.height = height
```
dgboss, line 181: This works great, but there is a more 'Reactful' way of setting the height of the map and side panel elements. I think setting properties on components by directly accessing the dom attributes is a bit of an anti-pattern in React, though perhaps it's just stylistic. I'll leave the choice of how you do this up to you, but I just wanted to point out another option since you're still relatively new to React. See my comments below which would allow you to get rid of this useEffect hook, the mapRef and the sidePanelRef.

PR #3133, opened 2023-09-22
Single values from weather model  rasters

conbrad:
…essor


brettedw (approved):
Looks great, nice work!

PR #3183, opened 2023-10-24
Imports zone units to advisory_shapes table

conbrad:
- Grabs fire zone units geojson from s3, adds `fire_zone_unit` shape type, adds fire zone units to `advisory_shapes` table


dgboss (approved):
LGTM!

PR #3202, opened 2023-10-30
Bias adjusted precip

dgboss:
- Add `bias_adjusted_precip_24h` to `WeatherModelStationPredictions` and the materialized view
- Add crud functions to get actual and predicted 24 hour precip
- Add `PrecipModel` class for handling precip specific linear regression (a unique `add_samples()` function is required
# Test Links:
[Landing Page](https://wps-pr-3202.apps.silver.devops.gov.bc.ca/)
[MoreCast 2.0](https://wps-pr-3202.apps.silver.devops.gov.bc.ca/morecast-2)
[Percentile Calculator](https://wps-pr-3202.apps.silver.devops.gov.bc.ca/percentile-calculator)
[MoreCast](https://wps-pr-3202.apps.silver.devops.gov.bc.ca/morecast)
[C-Haines](https://wps-pr-3202.apps.silver.devops.gov.bc.ca/c-haines)
[FireBat](https://wps-pr-3202.apps.silver.devops.gov.bc.ca/fire-behaviour-calculator)
[FireBat bookmark](https://wps-pr-3202.apps.silver.devops.gov.bc.ca/fire-behaviour-calculator?s=266&f=c5&c=NaN&w=20,s=286&f=c7&c=NaN&w=16,s=1055&f=c7&c=NaN&w=NaN,s=305&f=c7&c=NaN&w=NaN,s=344&f=c5&c=NaN&w=NaN,s=346&f=c7&c=NaN&w=NaN,s=328&f=c7&c=NaN&w=NaN,s=1399&f=c7&c=NaN&w=NaN,s=334&f=c7&c=NaN&w=NaN,s=1082&f=c3&c=NaN&w=NaN,s=388&f=c7&c=NaN&w=NaN,s=309&f=c7&c=NaN&w=16,s=306&f=c7&c=NaN&w=NaN,s=1029&f=c7&c=NaN&w=NaN,s=298&f=c7&c=NaN&w=NaN,s=836&f=c7&c=NaN&w=NaN,s=9999&f=c7&c=NaN&w=NaN)
[Auto Spatial Advisory (ASA)](https://wps-pr-3202.apps.silver.devops.gov.bc.ca/auto-spatial-advisory)
[HFI Calculator](https://wps-pr-3202.apps.silver.devops.gov.bc.ca/hfi-calculator)


api/app/weather_models/machine_learning.py
```diff
@@ -171,6 +174,21 @@ def learn(self):
         self.regression_models_v2.collect_data(query)
         self.regression_models_v2.train()
 
+        # Precip is based on 24 hour periods at 20:00 hours UTC. Use the start_date
```
conbrad, line 177: I think we can encapsulate this chunk of logic into it's own class and have it managed by `RegressionModelsV2` similar what was done for `WindDirectionModel`.

api/app/weather_models/precip_model.py
```diff
@@ -0,0 +1,46 @@
+import logging
+import math
+from datetime import datetime
+from typing import List
+from app.weather_models.linear_model import LinearModel
+from app.weather_models.utils import construct_dictionary_from_list_by_property
+
+logger = logging.getLogger(__name__)
+
+class PrecipModel():
```
conbrad, line 10: Should be an `RegressionModelProto` like `WindDirectionModel`

api/app/db/crud/observations.py
```diff
@@ -69,3 +70,48 @@ def get_accumulated_precipitation(session: Session, station_code: int, start_dat
...
+    :param session: The ORM/database session.
+    :param station_code: The numeric code identifying the weather station of interest.
+    :param start_datetime: The earliest date and time of interest.
+    :param end_datetime: The latest date and time of interest.
+    """
+    # Note: I couldn't construct this query in SQLAlchemy, hence the need for the 'text' based query.
+    stmt = """
+        SELECT day, station_code, sum(precipitation) actual_precip_24h
+        FROM
+            generate_series(\'{}\', \'{}\', '24 hours'::interval) day
```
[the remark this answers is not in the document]
dgboss: Haha, you have the gist of it. I'm glad you asked this question, I think I found an off-by-one day logic error.

I'll add to the comment with a description of what this is doing. 

`generate_series(\'{}\', \'{}\', '24 hours'::interval)`
This gives us a one column table of dates separated by 24 hours between the start and end dates. For example, if start and end dates are 2023-10-31 20:00:00 to 2023-11-03 20:00:00 we would have a table like:
| day |
|-----|
|2023-10-31 20:00:00|
|2023-11-01 20:00:00|
|2023-11-02 20:00:00|
|2023-11-03 20:00:00|

We then join the HourlyActuals table so that we can sum hourly precip in a 24 hour period. The join is based on the `weather_date` in the `HourlyActuals` table  field being in a 24 hour range using this odd looking syntax:

`weather_date <@ tstzrange(day, day + '24 hours', '(]')`

Using 2023-10-31 20:00:00 as an example, rows with the following dates would match. The (] syntax means the lower bound is excluded but the upper bound is included.

2023-10-31 21:00:00
2023-10-31 22:00:00
2023-10-31 23:00:00
2023-11-01 00:00:00
2023-11-01 01:00:00
....
2023-11-01 19:00:00
2023-11-01 20:00:00

My off by one error is that right now I am attributing this accumulated precipitation to October 31, but it should actually be for Nov 1 (we need the preceding 24 hour precip).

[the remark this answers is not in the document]
dgboss: An index on weather_date is a great idea!

api/app/weather_models/machine_learning.py
dgboss: I could pull it out into a separate function, I'm not sure it makes sense to throw all of it into its own class though? I think I still need to query for the data using `self.session', unless I pass the session to my class.
I can add the `PrecipModel` to `RegressionModelV2' though.

api/app/weather_models/precip_model.py
dgboss: The `add_sample` function of this class requires a unique implementation and does not accept the `HourlyActual` and `ModelRunPrediction` parameters required by the function signature in `RegressionModelProto`. I could just remove the `add_sample` abstract function from `RegressionModelProto`?

conbrad: Yeah, I like the latter (removing `add_sample` from `RegressionModelProto`), I think this interface will evolve as we see what the different weather parameters need.

conbrad: Oh that may not work since there's a few places that expect to be calling `add_sample` on a model -- feel free to disregard this comment, precip is way different than others.

api/app/weather_models/machine_learning.py
conbrad: Yeah true, I think the interface I introduce doesn't generalize to precip -- it's too different. Feel free to disregard this comment.

conbrad (approved):

api/app/weather_models/machine_learning.py
```diff
@@ -170,6 +174,24 @@ def learn(self):
             self.session, self.model.id, self.station_code, start_date, self.max_learn_date)
         self.regression_models_v2.collect_data(query)
         self.regression_models_v2.train()
+        
+    def learn_precip_model(self, start_date):
```
conbrad, line 178: The last thing I'd consider is adding some tests for this function

dgboss: I incorporated testing into the machine learning tests of the other bias adjusted weather parameters in `test_machine_learning.py`.

PR #3224, opened 2023-11-08
Fix cypress tests

dgboss:


brettedw (approved):
Interesting find on the fix!

PR #3449, opened 2024-03-05
Forecast versus actual comparison icons

dgboss:
- Show a +/- icon depending on if the forecasted value is greater than or less than the actual
- Don't show for WindDirection
- Add darker border around editable Forecast cells

# Test Links:
[Landing Page](https://wps-pr-3449.apps.silver.devops.gov.bc.ca/)
[MoreCast 2.0](https://wps-pr-3449.apps.silver.devops.gov.bc.ca/morecast-2)
[Percentile Calculator](https://wps-pr-3449.apps.silver.devops.gov.bc.ca/percentile-calculator)
[MoreCast](https://wps-pr-3449.apps.silver.devops.gov.bc.ca/morecast)
[C-Haines](https://wps-pr-3449.apps.silver.devops.gov.bc.ca/c-haines)
[FireBat](https://wps-pr-3449.apps.silver.devops.gov.bc.ca/fire-behaviour-calculator)
[FireBat bookmark](https://wps-pr-3449.apps.silver.devops.gov.bc.ca/fire-behaviour-calculator?s=266&f=c5&c=NaN&w=20,s=286&f=c7&c=NaN&w=16,s=1055&f=c7&c=NaN&w=NaN,s=305&f=c7&c=NaN&w=NaN,s=344&f=c5&c=NaN&w=NaN,s=346&f=c7&c=NaN&w=NaN,s=328&f=c7&c=NaN&w=NaN,s=1399&f=c7&c=NaN&w=NaN,s=334&f=c7&c=NaN&w=NaN,s=1082&f=c3&c=NaN&w=NaN,s=388&f=c7&c=NaN&w=NaN,s=309&f=c7&c=NaN&w=16,s=306&f=c7&c=NaN&w=NaN,s=1029&f=c7&c=NaN&w=NaN,s=298&f=c7&c=NaN&w=NaN,s=836&f=c7&c=NaN&w=NaN,s=9999&f=c7&c=NaN&w=NaN)
[Auto Spatial Advisory (ASA)](https://wps-pr-3449.apps.silver.devops.gov.bc.ca/auto-spatial-advisory)
[HFI Calculator](https://wps-pr-3449.apps.silver.devops.gov.bc.ca/hfi-calculator)


web/src/features/moreCast2/components/colDefBuilder.test.tsx
```diff
@@ -231,4 +196,29 @@ describe('ColDefBuilder', () => {
       ).toEqual({ testField: { choice: ModelChoice.MANUAL, value: 2 } })
     })
   })
+  it('should generate grass curing cold def correctly', () => {
```
conbrad, line 199:
```suggestion
  it('should generate grass curing column definition correctly', () => {
```

web/src/features/moreCast2/components/GridComponentRenderer.tsx
```diff
@@ -104,20 +107,90 @@ export class GridComponentRenderer {
...
+    let showGreaterThan = false
+    // Only show + and - icons if an actual value exists, a forecast value exists and this is not a windDirection
+    // field.
+    if (!isNaN(actualValue) && isNumber(actualValue) && isNumber(formattedValue) && !field.includes('windDirection')) {
+      showLessThan = formattedValue < actualValue
+      showGreaterThan = params.formattedValue > actualValue
+    }
+
+    // The grass curing 'forecast' field is rendered differently
+    if (isGrassField) {
```
conbrad, line 125: I think we can pull this out into a component and get some test coverage around it

brettedw (approved):
Nice work, looks good!

web/src/features/moreCast2/components/GridComponentRenderer.tsx
dgboss: Good call. I pulled out a ForecastCell component and created some tests around it.

web/src/features/moreCast2/components/forecastCell.test.tsx
```diff
@@ -0,0 +1,160 @@
...
+        showGreaterThan={false}
+        showLessThan={false}
+        value={params.formattedValue}
+      />
+    )
+
+    const element = queryByTestId('forecast-cell-less-than-icon')
+    expect(element).not.toBeInTheDocument()
+  })
+  it('should show greater than icon when showGreaterThan is true', () => {
```
conbrad, line 70: What if `showGreaterThan` and `showLessThan` are both true?

dgboss: Technically both icons would display which I could test for, but logic in the parent component prevents this from happening.

```      showLessThan = formattedValue < actualValue
      showGreaterThan = formattedValue > actualValue
```

Any thoughts on what we should do/test?

conbrad: I think we just need to show one or the other, so we could just use one boolean input, but we also need to consider the unlikely equal case though, which is what is achieved when both booleans are false. 

I would say disregard if the parent guarantees one or both are false.

conbrad (approved):

web/src/features/moreCast2/components/forecastCell.test.tsx
dgboss: A single boolean won't quite cover it as no icon is rendered when forecast === actual. I suppose we could pass both the forecast and actual values as props but the component doesn't really need to know about both values as it only displays the forecast value if it exists.

I've added one more test to explicitly cover the case where `showLessThan` and `showGreaterThan` are both false.

dgboss:
Note that the code duplication is due to tests.

PR #3461, opened 2024-03-12
Save Draft Forecast

brettedw:
- saves draft forecast in localStorage
- adds reset button
# Test Links:
[Landing Page](https://wps-pr-3461.apps.silver.devops.gov.bc.ca/)
[MoreCast 2.0](https://wps-pr-3461.apps.silver.devops.gov.bc.ca/morecast-2)
[Percentile Calculator](https://wps-pr-3461.apps.silver.devops.gov.bc.ca/percentile-calculator)
[MoreCast](https://wps-pr-3461.apps.silver.devops.gov.bc.ca/morecast)
[C-Haines](https://wps-pr-3461.apps.silver.devops.gov.bc.ca/c-haines)
[FireBat](https://wps-pr-3461.apps.silver.devops.gov.bc.ca/fire-behaviour-calculator)
[FireBat bookmark](https://wps-pr-3461.apps.silver.devops.gov.bc.ca/fire-behaviour-calculator?s=266&f=c5&c=NaN&w=20,s=286&f=c7&c=NaN&w=16,s=1055&f=c7&c=NaN&w=NaN,s=305&f=c7&c=NaN&w=NaN,s=344&f=c5&c=NaN&w=NaN,s=346&f=c7&c=NaN&w=NaN,s=328&f=c7&c=NaN&w=NaN,s=1399&f=c7&c=NaN&w=NaN,s=334&f=c7&c=NaN&w=NaN,s=1082&f=c3&c=NaN&w=NaN,s=388&f=c7&c=NaN&w=NaN,s=309&f=c7&c=NaN&w=16,s=306&f=c7&c=NaN&w=NaN,s=1029&f=c7&c=NaN&w=NaN,s=298&f=c7&c=NaN&w=NaN,s=836&f=c7&c=NaN&w=NaN,s=9999&f=c7&c=NaN&w=NaN)
[Auto Spatial Advisory (ASA)](https://wps-pr-3461.apps.silver.devops.gov.bc.ca/auto-spatial-advisory)
[HFI Calculator](https://wps-pr-3461.apps.silver.devops.gov.bc.ca/hfi-calculator)


web/src/features/moreCast2/components/TabbedDataGrid.tsx
```diff
@@ -104,6 +101,8 @@ const TabbedDataGrid = ({ morecast2Rows, fromTo, setFromTo }: TabbedDataGridProp
     mouseY: number
   } | null>(null)
 
+  const storedDraftForecast = new MorecastDraftForecast(localStorage)
+
```
brettedw, line 105: @dgboss Curious how you would write this into a local state var when the useState setter isn't needed

web/src/features/moreCast2/forecastDraft.ts
```diff
@@ -0,0 +1,70 @@
...
+    const storedRowsMap = getRowsMap(storedForecastsToUpdate.rows)
+
+    rowsToStore.forEach(row => {
+      storedRowsMap.set(row.id, row)
+    })
+    // we only need to store rows that are 'Forecast' rows
+    storedForecastsToUpdate.rows = Array.from(storedRowsMap.values()).filter(row => {
+      return isForecastRow(row)
+    })
+    storedForecastsToUpdate.lastEdited = Date.now()
```
conbrad, line 43: It's not specified in the task but we should use a timezone with the date. For now we can stick to PST until we implement smart date.

You can use a `luxon` `DateTime` and pass it to the `pstFormatter` in `date.ts` since we've used that with HFI Calculator.

web/src/features/moreCast2/util.ts
```diff
@@ -102,6 +102,10 @@ export const validForecastPredicate = (row: MoreCast2Row) =>
   !isUndefined(row.windSpeedForecast) &&
   !isNaN(row.windSpeedForecast.value)
 
+export const isForecastRow = (row: MoreCast2Row) => {
+  return !rowContainsActual(row) && !isPreviousToToday(row.forDate)
```
conbrad, line 106: How do you feel about renaming `isPreviousToToday` to `isBeforeToday`?

web/src/features/moreCast2/components/TabbedDataGrid.tsx
brettedw: Would defining it this way but outside the component be a reasonable solution? It shouldn't instantiate a new object on every rerender then I don't think?

web/src/features/moreCast2/util.ts
```diff
@@ -123,6 +127,32 @@ export const mapForecastChoiceLabels = (newRows: MoreCast2Row[], storedRows: Mor
   return newRows
 }
 
+export const fillForecastsFromRows = (rowsToFill: MoreCast2Row[], savedRows: MoreCast2Row[]): MoreCast2Row[] => {
+  for (const row of rowsToFill) {
```
conbrad, line 131: I think this could be simplified a bit:
- `if(!savedRows)` then return `rowsToFill` right away as an if guard
- Can we also add that `savedRows` is potentially `undefined` or `null` in the parameter type?
- Can rewrite the `for of` loop to a filter: `rowsToFill.filter(row => isForecastRow(row)).map(forecastRow => { // ... do the things })`

web/src/features/moreCast2/util.ts
```diff
@@ -224,3 +255,23 @@ export const isPreviousToToday = (datetime: DateTime): boolean => {
 
   return datetime < today
 }
+
+export const rowContainsActual = (row: MoreCast2Row): boolean => {
+  for (const key in row) {
```
conbrad, line 260: Can simplify loop by using `some` from `lodash`: https://lodash.com/docs/4.17.15#some

conbrad (commented):
Looking really good! 😎

web/src/features/moreCast2/util.ts
brettedw: I feel slightly embarrassed that it was ever named `isPreviousToToday` in the first place haha. I'll definitely rename it

conbrad: Lol no worries

web/src/features/moreCast2/forecastDraft.ts
brettedw: `pstFormatter` doesn't handle times, and it turns out the `fromObject` method used in it doesn't convert time zones, it interprets the time being passed to it to have the time zone specified in `fromObject`. So if I gave it a DateTime.now(), it would return the wrong time since we're currently in PDT
Just a note for the future smart dates

brettedw: Maybe that's already obvious in the name `pstFormatter`

web/src/features/moreCast2/forecastDraft.test.ts
```diff
@@ -0,0 +1,70 @@
+import { MorecastDraftForecast } from 'features/moreCast2/forecastDraft'
+import { DraftMorecast2Rows } from 'features/moreCast2/interfaces'
+import { buildValidActualRow, buildValidForecastRow } from 'features/moreCast2/rowFilters.test'
+import { DateTime, Settings } from 'luxon'
+
+// Temporarily set DateTime.now() to return the same DateTime when called
+const TEST_DATE = DateTime.fromISO('2024-01-01T00:00:00.000-08:00')
+Settings.now = () => TEST_DATE.toMillis()
```
conbrad, line 8: We can just mock `getDateTimeNowPST` to return `TEST_DATE` instead. I'm way of the scope of this setting change.

conbrad: Take a look at `web/src/features/auth/slices/authenticationSlice.test.ts` for how to mock with `sinon`

web/src/features/moreCast2/components/ResetForecastButton.tsx
```diff
@@ -0,0 +1,53 @@
+import React from 'react'
+import { Button } from '@mui/material'
+import { fillGrassCuringForecast } from 'features/moreCast2/util'
+import { MoreCast2Row, PredictionItem } from 'features/moreCast2/interfaces'
+import { ModelChoice, WeatherDeterminate } from 'api/moreCast2API'
+// import { MorecastDraftForecast } from 'features/moreCast2/forecastDraft'
```
dgboss, line 6: Leftover import

web/src/features/moreCast2/forecastDraft.test.ts
```diff
@@ -0,0 +1,73 @@
...
+    buildValidForecastRow(111, TEST_DATE.plus({ days: 1 })),
+    buildValidForecastRow(111, TEST_DATE.plus({ days: 2 })),
+    buildValidForecastRow(222, TEST_DATE.plus({ days: 1 })),
+    buildValidForecastRow(222, TEST_DATE.plus({ days: 2 })),
+    buildValidActualRow(222, TEST_DATE.minus({ days: 1 })),
+    buildValidActualRow(222, TEST_DATE.minus({ days: 2 }))
+  ]
+
+  it('should only store forecast rows', () => {
+    jest.spyOn(DateUtils, 'getDateTimeNowPST').mockReturnValue(TEST_DATE)
```
brettedw, line 34: @conbrad Is this an ok solution? Still have some things to learn about ways to mock things...
It was actually `isBeforeToday` that was causing the issue, I changed the `forecastDraft` functions to not use `DateTime.now()` this morning like you suggested.
I just changed `isBeforeToday()` with this commit and mocked it with jest, since it seemed like I could do the mock the same way as sinon

web/src/features/moreCast2/interfaces.ts
```diff
@@ -143,3 +143,8 @@ export interface ForecastMorecast2Row extends BaseRow {
   windDirection?: PredictionItem
   windSpeed?: PredictionItem
 }
+
+export interface DraftMorecast2Rows {
+  rows: MoreCast2Row[]
+  lastEdited: string | null
```
dgboss, line 149: In my experience I find it better to handle dates as objects and leave them as objects right up until the point they are displayed in the UI. Once you start working with dates as strings you can run into all sorts of pain related to timezones. 

I'd recommend that `lastEdited` be a luxon DateTime.

web/src/features/moreCast2/forecastDraft.test.ts
conbrad: Yes! Nice!

conbrad: You'll actually want to clean up the spy after the test though -- with `sinon` this can be done with a sandbox, with `jest` you'll want something like this:

```
afterEach(() => {
  // restore the spy created with spyOn
  jest.restoreAllMocks();
});
```

If this is the only function that spies, you can just call `jest.restoreAllMocks();` at the end of the test.

web/src/features/moreCast2/components/TabbedDataGrid.tsx
dgboss: Yes, this is perfectly reasonable and some would argue a better approach I'm sure. The alternative is to do `const [storedDraftForecast] = useState(new MorecastDraftForecast(localStorage))`

web/src/features/moreCast2/interfaces.ts
dgboss: Happy to discuss though!

web/src/features/moreCast2/forecastDraft.test.ts
brettedw: Ah I was using `clearAllMocks`, just spent a little more time reading about the differences. Thanks!

PR #3563, opened 2024-04-25
Limit hourlies endpoint to only store ffmc files

conbrad:
- Checks that the received file on the hourlies endpoint is an ffmc and only puts it on S3 if it is


api/app/tests/sfms/test_sfms_upload.py
```diff
@@ -221,4 +221,29 @@ async def _mock_get_client_for_router():
...
+    client = TestClient(app)
+    response = client.post(HOURLY_FFMC_URL,
+                           files={'file': ('hfi20220904.tiff', b'')},
+                           headers={
+                               'Secret': config.get('SFMS_SECRET'),
+                               'Last-modified': datetime.now().isoformat(),
+                               'Create-time': datetime.now().isoformat()})
+    # We should get a 200 response if the file is uploaded successfully.
+    assert response.status_code == 200
+    # We should have called put_object once.
```
dgboss, line 246: Comment should say put_object shouldn't be called.

dgboss (approved):

PR #3690, opened 2024-05-30
Interpolation test start

brettedw:


api/app/weather_models/__init__.py
```diff
@@ -1,10 +1,6 @@
 """ Code common to app.weather_models.fetch """
-from datetime import datetime
 from enum import Enum
 import logging
-from scipy.interpolate import interp1d
-
-from app.db.models.weather_models import ModelRunPrediction
 
 logger = logging.getLogger(__name__)
```
dgboss, line 5: I think we can get rid of the logging code.

dgboss: Or is it needed because this is __init__.py?

brettedw: Good call, don't think it's needed. We have many `__init__` files with no logger, but logging in individual modules

api/app/tests/weather_models/test_interpolation.py
```diff
@@ -0,0 +1,30 @@
...
+    prediction_a = generate_model_run_prediction(prediction_a_timestamp, 1.1)
+    prediction_b = generate_model_run_prediction(prediction_b_timestamp, 4.4)
+    
+    noon_prediction = construct_interpolated_noon_prediction(prediction_a, prediction_b, SCALAR_MODEL_VALUE_KEYS_FOR_INTERPOLATION)
+    assert noon_prediction.prediction_timestamp.hour == 20
+    assert noon_prediction.apcp_sfc_0 is not None
+    assert noon_prediction.rh_tgl_2 is not None
+    assert noon_prediction.wdir_tgl_10 is not None
+    assert noon_prediction.wind_tgl_10 is not None
+    assert noon_prediction.tmp_tgl_2 is not None
```
dgboss, line 30: Might be worth considering adding asserts for the actual values to make sure we're doing the interpolation as expected. Or alternatively, add a test for `construct_interpolated_noon_prediction`. We could also add tests for the `interpolate.py` file at a later date.

dgboss (approved):
Looks good to me!

conbrad (approved):
Nice work!

PR #3692, opened 2024-06-03
Limit frontend perf tracing, set user in backend

conbrad:
Limit our perf credit spending on frontend performance tracking for sentry, set sentry user email in backend

### DIFF
--- a/api/app/auth.py
+++ b/api/app/auth.py
@@ -4,6 +4,7 @@ from fastapi import Depends, HTTPException, status, Request
 from fastapi.security import OAuth2PasswordBearer
 import jwt
 from jwt import InvalidTokenError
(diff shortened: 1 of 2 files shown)


api/app/auth.py
```diff
@@ -55,12 +56,13 @@ async def audit(request: Request, token=Depends(authenticate)):
 
 
 async def authentication_required(token=Depends(authenticate)):
-    """ Raises HTTPExcecption with status code 401 if authentation fails."""
+    """ Raises HTTPException with status code 401 if authentation fails."""
```
dgboss, line 59:
```suggestion
    """ Raises HTTPException with status code 401 if authentication fails."""
```

dgboss (approved):
LGTM!

PR #3800, opened 2024-07-24
Task/tpi summarize hfi

conbrad:
Summarizes advisory area based on TPI. Uses the snow-masked pmtiles hfi layer for the given date to mask contributing TPI pixels.

After finding out the initial implementation uses up to 20GB of memory, commit https://github.com/bcgov/wps/pull/3800/commits/ab72307a28f6797afb3df2d50212e6c722ae492a and onwards introduce a more memory optimized implementation. Basically all datasets are loaded from S3, but data is only read in and transformed a chunk at a time. This implementation stays under 4.5GB of memory usage.

- Tests: https://github.com/bcgov/wps/issues/3821

# Test Links:
[Landing Page](https://wps-pr-3800-e1e498-dev.apps.silver.devops.gov.bc.ca/)
[MoreCast](https://wps-pr-3800-e1e498-dev.apps.silver.devops.gov.bc.ca/morecast)
[Percentile Calculator](https://wps-pr-3800-e1e498-dev.apps.silver.devops.gov.bc.ca/percentile-calculator)
[C-Haines](https://wps-pr-3800-e1e498-dev.apps.silver.devops.gov.bc.ca/c-haines)
[FireBat](https://wps-pr-3800-e1e498-dev.apps.silver.devops.gov.bc.ca/fire-behaviour-calculator)
[FireBat bookmark](https://wps-pr-3800-e1e498-dev.apps.silver.devops.gov.bc.ca/fire-behaviour-calculator?s=266&f=c5&c=NaN&w=20,s=286&f=c7&c=NaN&w=16,s=1055&f=c7&c=NaN&w=NaN,s=305&f=c7&c=NaN&w=NaN,s=344&f=c5&c=NaN&w=NaN,s=346&f=c7&c=NaN&w=NaN,s=328&f=c7&c=NaN&w=NaN,s=1399&f=c7&c=NaN&w=NaN,s=334&f=c7&c=NaN&w=NaN,s=1082&f=c3&c=NaN&w=NaN,s=388&f=c7&c=NaN&w=NaN,s=309&f=c7&c=NaN&w=16,s=306&f=c7&c=NaN&w=NaN,s=1029&f=c7&c=NaN&w=NaN,s=298&f=c7&c=NaN&w=NaN,s=836&f=c7&c=NaN&w=NaN,s=9999&f=c7&c=NaN&w=NaN)
[Auto Spatial Advisory (ASA)](https://wps-pr-3800-e1e498-dev.apps.silver.devops.gov.bc.ca/auto-spatial-advisory)
[HFI Calculator](https://wps-pr-3800-e1e498-dev.apps.silver.devops.gov.bc.ca/hfi-calculator)


api/app/utils/geospatial.py
```diff
@@ -0,0 +1,165 @@
...
+    :param data_source: input raster to extract metadata from
+    :param options: defines the metadata the caller is interested in extracting from the raster
+    :return: optional metadata of geotransform, project, x and y sizes from the raster
+    """
+    geotransform = None
+    projection = None
+    x_size = None
+    y_size = None
+    if options is None:
+        return (geotransform, projection.x_size, y_size)
```
dgboss, line 63:
```suggestion
        return (geotransform, projection, x_size, y_size)
```

api/app/utils/geospatial.py
```diff
@@ -0,0 +1,165 @@
...
+
+
+def read_raster_data(data_source):
+    """
+    Read raster data and return as array from the data source
+
+    :param data_source: data source to read the data from
+    :return: raster data as array
+    """
+    data_band = data_source.GetRasterBand(1)
```
dgboss, line 87: Some of the grib files we work with (NAM and GFS I think) have all the weather parameters in a single file, each in their own band. We could consider adding an optional `band` parameter to this function with a default of 1.
`def read_raster_data(data_source, band=1)`

api/app/utils/geospatial.py
```diff
@@ -0,0 +1,165 @@
...
+
+def _get_raster_data_source(key: str, s3_data, options: Optional[GeospatialOptions]):
+    """
+    Retrieves geospatial file from S3. If the options include Vector2RasterOptions, it will transform
+    the vector into a raster and return the raster datasource. If no Vector2RasterOptions are included,
+    it will assume the file is a raster and try to read it as a raster. The caller is expected to know whether
+    the geospatial file is a raster or vector and supply the appropriate options.
+
+    :param key: the object store key pointing to the desired geospatial file
+    :param s3_data: the object store content blob
```
dgboss, line 115: is s3_data the 'destination' for the file to be opened? Is it a variable or a file path?

api/app/utils/geospatial.py
```diff
@@ -0,0 +1,165 @@
...
+    :param options: options defining the desired metadata as well as vector options if the file is a vector
+    :return: raster data source
+    """
+    mem_path = f"/vsimem/{key}"
+    gdal.FileFromMemBuffer(mem_path, s3_data)
```
dgboss, line 120: Is this line needed? It looks like we open the mem_path here, in the `if` block with `gdal.OpenEx` and in the `else` block with `gdal.Open`.

dgboss: Nevermind, I think I figured it out. I was a little confused by the 'dual use' of the key parameter.

api/app/auto_spatial_advisory/tpi_area_summary.py
```diff
@@ -0,0 +1,45 @@
...
+    iso_date = for_date.isoformat()
+
+    async with get_client() as (client, bucket):
+        tpi_result = await read_raster_into_memory(
+            client,
+            bucket,
+            f'dem/tpi/{config.get("CLASSIFIED_TPI_DEM_NAME")}',
+            GeospatialOptions(include_geotransform=True, include_projection=True, include_x_size=True, include_y_size=True),
+        )
+        hfi_result = await read_raster_into_memory(
```
dgboss, line 25: It looks like we're massaging the HFI layer to match the geotransform, projection and dimensions of the tpi layer? We might be able to avoid this extra processing if we massage the tpi layer to match the HFI layer when we initially create the tpi layer. This could help to preserve the 'original` HFI pixel values which we loose when we do the reprojection. We could check in with Brett on Monday to see if my thoughts make sense.

dgboss (commented):
I'm loving the abstraction of opening rasters in `geospatial.py`. It's a great idea!

api/app/auto_spatial_advisory/tpi_area_summary.py
conbrad: > It looks like we're massaging the HFI layer to match the geotransform, projection and dimensions of the tpi layer?
Yeah that's the direction I received from Neal, unless I misconstrued it but perhaps we should revisit that.

conbrad:
> I'm loving the abstraction of opening rasters in `geospatial.py`. It's a great idea!

Naming and abstraction boundaries could be improved I think, but short of recreating `rasterio` I think it works for now. We can adjust as we go.

api/app/utils/geospatial.py
conbrad: Nice, done in: https://github.com/bcgov/wps/pull/3800/commits/615e2b1bb8966d2ca1986253abfe559532bf5796

api/app/utils/geospatial.py
conbrad: Hmm maybe I should create separate `vsimem` paths to avoid the confusion

conbrad: I adjusted the code to use separate path variables for in memory s3 blob and in memory generated raster here: https://github.com/bcgov/wps/pull/3800/commits/7e10bbd8b280a8d1abc84b66a30a863c98d75126

api/app/auto_spatial_advisory/tpi_area_summary.py
brettedw: Ya I didn't think about projections but I guess what we're doing with the HFI is going from lambert (original) to WGS84 (pmtiles) then re-projecting again back to the DEM crs. Probably not ideal.
Curious to see how the HFI data looks at a pixel level after all this when compared to an original tif in an area where there's no snow

conbrad: Maybe I should spend the time to adjust our HFI geospatial pipeline to save the snow-masked HFI as rasters as well as pmtimes.

api/app/db/models/auto_spatial_advisory.py
```diff
@@ -175,20 +171,38 @@ class AdvisoryElevationStats(Base):
...
+    __tablename__ = "advisory_tpi_stats"
+    __table_args__ = {"comment": "Elevation TPI stats per fire shape"}
+    id = Column(Integer, primary_key=True, index=True)
+    advisory_shape_id = Column(Integer, ForeignKey(Shape.id), nullable=False, index=True)
+    threshold = Column(Integer, ForeignKey(HfiClassificationThreshold.id), nullable=False)
+    run_parameters = Column(Integer, ForeignKey(RunParameters.id), nullable=False, index=True)
+    valley_bottom = Column(Integer, nullable=False, index=True)
+    mid_slope = Column(Integer, nullable=False, index=True)
+    upper_slope = Column(Integer, nullable=False, index=True)
+    raster_height = Column(Integer, nullable=False, index=True)
```
dgboss, line 206: What do `raster_height` and `raster_width` represent?

api/app/auto_spatial_advisory/elevation.py
```diff
@@ -178,6 +180,61 @@ def apply_threshold_mask_to_dem(threshold: int, mask_path: str, temp_dir: str):
...
+        async with get_async_write_session_scope() as session:
+            stmt = text("SELECT id, source_identifier FROM advisory_shapes;")
+            result = await session.execute(stmt)
+            rows = result.all()
+
+            for row in rows:
+                cut_tpi_result = cut_raster_by_shape_id(row[0], row[1], tpi_result.data_source, raster_options)
+                cut_hfi_result = cut_raster_by_shape_id(row[0], row[1], hfi_result.data_source, raster_options)
+                # transforms all pixels that are 255 to 1, this is what we get from pmtiles
+                hfi_4k_or_above = np.where(cut_hfi_result.data_array == 255, 1, cut_hfi_result.data_array)
```
dgboss, line 225: Is this classification still necessary?

conbrad: Not once the snow master raster tif PR is merged, will update this PR against main once that's in

dgboss: Oh, I see this is still a bit of a WIP

api/app/db/models/auto_spatial_advisory.py
conbrad: I took the wording from QGIS here, I think they represent # of cells in column and row of the raster
wdithheight

dgboss: I'm wondering if we need these values? They relate to the underling raster, not to the fire zone unit we're storing data for so I'm not clear on the potential value they provide.

conbrad: Ah good point, added the table before I had broken zone the raster by fire zone. I'll delete it!

api/app/auto_spatial_advisory/tpi_area_summary.py
conbrad: Ok going to adjust this based on the work merged here: https://github.com/bcgov/wps/pull/3809

api/app/db/models/auto_spatial_advisory.py
```diff
@@ -175,20 +171,35 @@ class AdvisoryElevationStats(Base):
...
+    Each record includes the raster pixel count of the above classifications as well as the raster pixel size and resolution in metres
+    and an advisory shape the stats are related to.
+    """
+
+    __tablename__ = "advisory_tpi_stats"
+    __table_args__ = {"comment": "Elevation TPI stats per fire shape"}
+    id = Column(Integer, primary_key=True, index=True)
+    advisory_shape_id = Column(Integer, ForeignKey(Shape.id), nullable=False, index=True)
+    run_parameters = Column(Integer, ForeignKey(RunParameters.id), nullable=False, index=True)
+    valley_bottom = Column(Integer, nullable=False, index=True)
```
dgboss, line 202: I don't think we need indexes on the valley bottom, mid slope, upper slope and pixel_size_metres columns.

api/app/auto_spatial_advisory/elevation.py
```diff
@@ -178,6 +210,63 @@ def apply_threshold_mask_to_dem(threshold: int, mask_path: str, temp_dir: str):
...
+            ),
+        )
+
+        async with get_async_write_session_scope() as session:
+            stmt = text("SELECT id, source_identifier FROM advisory_shapes;")
+            result = await session.execute(stmt)
+            rows = result.all()
+
+            for row in rows:
+                cut_tpi_result = cut_raster_by_shape_id(row[0], row[1], tpi_result.data_source, raster_options)
```
dgboss, line 256: Not a big deal, but might be slightly simpler computationally to multiply the two rasters first outside of the for loop, then cut out the advisory shapes from the result.

dgboss (approved):
I haven't tested it locally but it looks good!

api/app/auto_spatial_advisory/elevation.py
conbrad: Good call, thanks, I'll make that change.

api/app/db/models/auto_spatial_advisory.py
conbrad: Done in: https://github.com/bcgov/wps/pull/3800/commits/31ed215e807ef4e7fbb5e4d483ca53b06d785cad

brettedw (approved):
Looks great, nice work!

api/app/auto_spatial_advisory/elevation.py
```diff
@@ -178,6 +211,75 @@ def apply_threshold_mask_to_dem(threshold: int, mask_path: str, temp_dir: str):
...
+
+    gdal.SetConfigOption("AWS_SECRET_ACCESS_KEY", config.get("OBJECT_STORE_SECRET"))
+    gdal.SetConfigOption("AWS_ACCESS_KEY_ID", config.get("OBJECT_STORE_USER_ID"))
+    gdal.SetConfigOption("AWS_S3_ENDPOINT", config.get("OBJECT_STORE_SERVER"))
+    gdal.SetConfigOption("AWS_VIRTUAL_HOSTING", "FALSE")
+    bucket = config.get("OBJECT_STORE_BUCKET")
+    dem_file = config.get("CLASSIFIED_TPI_DEM_NAME")
+
+    key = f"/vsis3/{bucket}/dem/tpi/{dem_file}"
+    tpi_source: gdal.Dataset = gdal.Open(key, gdal.GA_ReadOnly)
```
dgboss, line 248: `tpi_source` should also be closed/set to `None` when we're done with it.

api/app/auto_spatial_advisory/elevation.py
```diff
@@ -178,6 +211,75 @@ def apply_threshold_mask_to_dem(threshold: int, mask_path: str, temp_dir: str):
...
+    pixel_size_metres = int(tpi_source.GetGeoTransform()[1])
+
+    hfi_raster_filename = get_raster_tif_filename(for_date)
+    hfi_raster_key = get_raster_filepath(run_date, run_type, hfi_raster_filename)
+    hfi_key = f"/vsis3/{bucket}/{hfi_raster_key}"
+    hfi_source: gdal.Dataset = gdal.Open(hfi_key, gdal.GA_ReadOnly)
```
dgboss, line 254: `hfi_source` can also be closed/set to `None`

api/app/utils/geospatial.py
```diff
@@ -0,0 +1,84 @@
...
+
+    # Process in chunks
+    for y in range(0, y_size, chunk_size):
+        y_chunk_size = min(chunk_size, y_size - y)
+
+        for x in range(0, x_size, chunk_size):
+            x_chunk_size = min(chunk_size, x_size - x)
+
+            # Read chunks from both rasters
+            tpi_chunk = tpi_raster.GetRasterBand(1).ReadAsArray(x, y, x_chunk_size, y_chunk_size)
```
dgboss, line 70: Could move these `GetRasterBand(1)` calls outside the double for loop to prevent the n x m calls.

dgboss (approved):
Nice re-work! Just a few comments/thoughts.

api/app/auto_spatial_advisory/elevation.py
```diff
@@ -178,6 +211,75 @@ def apply_threshold_mask_to_dem(threshold: int, mask_path: str, temp_dir: str):
...
+
+    warped_mem_path = f"/vsimem/warp_{hfi_raster_filename}"
+    resized_hfi_source: gdal.Dataset = warp_to_match_extent(hfi_source, tpi_source, warped_mem_path)
+    hfi_masked_tpi = raster_mul(tpi_source, resized_hfi_source)
+    resized_hfi_source = None
+
+    fire_zone_stats: Dict[int, Dict[int, int]] = {}
+    async with get_async_write_session_scope() as session:
```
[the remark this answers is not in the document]
conbrad: We need to to do the writes for the advisory tpi stats, erred on the side of reusing a single session with the widest necessary permissions over created 2 sessions. Not sure how the behavior works with leader pod, etc. I can change  it though.

api/app/auto_spatial_advisory/elevation.py
```diff
@@ -13,16 +15,47 @@
...
+
+    :param hfi_s3_key: the object store key pointing to the hfi tif to intersect with tpi layer
+    :param run_type: The type of run to process. (is it a forecast or actual run?)
+    :param run_datetime: The date and time of the run to process. (when was the hfi file created?)
+    :param for_date: The date of the hfi to process. (when is the hfi for?)
+    """
+    logger.info("Processing elevation stats %s for run date: %s, for date: %s", run_type, run_datetime, for_date)
+    perf_start = perf_counter()
+    # Get the id from run_parameters associated with the provided run_type, for_date and for_datetime
+    async with get_async_write_session_scope() as session:
```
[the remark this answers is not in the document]
conbrad: Not sure either, opted to just use the single write session for the whole transaction. I think there's a risk in opening multiple sessions that we lose the transactional view of read data that informs the writes. WDYT?

api/app/utils/geospatial.py
conbrad: Done in https://github.com/bcgov/wps/pull/3800/commits/ed06548a2a71ab77888dae7b5829a7470f2f4a9a

api/app/auto_spatial_advisory/elevation.py
```diff
@@ -178,6 +211,75 @@ def apply_threshold_mask_to_dem(threshold: int, mask_path: str, temp_dir: str):
     return masked_dem_path
 
 
+@dataclass(frozen=True)
+class FireZoneTPIStats:
+    """
+    Captures fire zone stats of TPI pixels hitting >4K HFI threshold via
+    a dictionary, fire_zone_stats, of {source_identifier: {1: X, 2: Y, 3: Z}}, where 1 = valley bottom, 2 = mid slope, 3 = upper slope
```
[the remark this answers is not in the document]
conbrad: It's kinda awkward because we rely on `np.unique` for classified counts. So we'd have to add an extra bunch of code to make the integers enums. Not sure if it's worth it.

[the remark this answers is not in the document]
dgboss: Ok, I'm fine with leaving it as is.

api/app/utils/geospatial.py
```diff
@@ -0,0 +1,87 @@
...
+    x_res = source_geotransform[1]
+    y_res = -source_geotransform[5]
+    minx = source_geotransform[0]
+    maxy = source_geotransform[3]
+    maxx = minx + source_geotransform[1] * raster_to_match.RasterXSize
+    miny = maxy + source_geotransform[5] * raster_to_match.RasterYSize
+    extent = [minx, miny, maxx, maxy]
+
+    # Warp to match input option parameters
+    return gdal.Warp(output_path, source_raster, dstSRS=raster_to_match.GetProjection(), outputBounds=extent, xRes=x_res, yRes=y_res, resampleAlg=gdal.GRA_NearestNeighbour)
```
brettedw, line 29: I don't know that it matters too much at this scale, but I think for the next TPI we should reproject the DEM/TPI data to Lambert and store it that way, using something like bilinear/bicubic

conbrad: Is that because Lambert, bilinear/bicubic is less lossy when projected?

brettedw: Nothing better about Lambert, it's just what the HFI data is in and what our other DEMs are in, so it would avoid having to do any reprojecting during processing. I think just this new 50m DEM they gave us is in Albers. 
Bilinear/bicubic is best for continuous data, like elevation data.

api/app/auto_spatial_advisory/elevation.py
conbrad: Done in: https://github.com/bcgov/wps/pull/3800/commits/45c827fb54f8fe721ef04d6485c5ee8e05b40841

api/app/auto_spatial_advisory/elevation.py
conbrad: Done in: https://github.com/bcgov/wps/pull/3800/commits/45c827fb54f8fe721ef04d6485c5ee8e05b40841

PR #3849, opened 2024-08-14
Critical Hours

dgboss:
- Critical hours are calculated for each station within a fire zone unit. A representative start and end time is then selected for the fire zone unit.
- Critical hours are saved in a new critical_hours table.
- This is a temporary implementation that will be replaced once we are able to perform critical hours calculations spatially. As this is temporary there are no tests.
- A lot of functionality is borrowed/copied from FireBat.
- Back-end work only. API endpoint and front-end integration coming soon!
# Test Links:
[Landing Page](https://wps-pr-3849-e1e498-dev.apps.silver.devops.gov.bc.ca/)
[MoreCast](https://wps-pr-3849-e1e498-dev.apps.silver.devops.gov.bc.ca/morecast)
[Percentile Calculator](https://wps-pr-3849-e1e498-dev.apps.silver.devops.gov.bc.ca/percentile-calculator)
[C-Haines](https://wps-pr-3849-e1e498-dev.apps.silver.devops.gov.bc.ca/c-haines)
[FireBat](https://wps-pr-3849-e1e498-dev.apps.silver.devops.gov.bc.ca/fire-behaviour-calculator)
[FireBat bookmark](https://wps-pr-3849-e1e498-dev.apps.silver.devops.gov.bc.ca/fire-behaviour-calculator?s=266&f=c5&c=NaN&w=20,s=286&f=c7&c=NaN&w=16,s=1055&f=c7&c=NaN&w=NaN,s=305&f=c7&c=NaN&w=NaN,s=344&f=c5&c=NaN&w=NaN,s=346&f=c7&c=NaN&w=NaN,s=328&f=c7&c=NaN&w=NaN,s=1399&f=c7&c=NaN&w=NaN,s=334&f=c7&c=NaN&w=NaN,s=1082&f=c3&c=NaN&w=NaN,s=388&f=c7&c=NaN&w=NaN,s=309&f=c7&c=NaN&w=16,s=306&f=c7&c=NaN&w=NaN,s=1029&f=c7&c=NaN&w=NaN,s=298&f=c7&c=NaN&w=NaN,s=836&f=c7&c=NaN&w=NaN,s=9999&f=c7&c=NaN&w=NaN)
[Auto Spatial Advisory (ASA)](https://wps-pr-3849-e1e498-dev.apps.silver.devops.gov.bc.ca/auto-spatial-advisory)
[HFI Calculator](https://wps-pr-3849-e1e498-dev.apps.silver.devops.gov.bc.ca/hfi-calculator)


api/app/db/crud/auto_spatial_advisory.py
```diff
@@ -386,3 +423,29 @@ async def get_provincial_rollup(session: AsyncSession, run_type: RunTypeEnum, ru
     )
     result = await session.execute(stmt)
     return result.all()
+
+
+async def get_containing_zone(session: AsyncSession, geometry: str, srid: int):
+    geom = func.ST_GeomFromText(geometry, srid)
+    stmt = select(Shape.id).filter(func.ST_Contains(Shape.geom, geom))
```
brettedw, line 430: I think there might be a potential projection issue here if `srid` isn't the same projection as `Shape`'s srid. Could possibly make it more generic with an `ST_Transform`

api/app/utils/geospatial.py
```diff
@@ -83,3 +84,21 @@ def raster_mul(tpi_ds: gdal.Dataset, hfi_ds: gdal.Dataset, chunk_size=256) -> gd
...
 
     return out_ds
+
+
+class PointTransformer:
+    """
+    Transforms the coordinates of a point from one spatial reference to another.
+    """
+
+    def __init__(self, source_srs, target_srs):
```
brettedw, line 94: nit: some type hints here might be handy

api/app/db/crud/auto_spatial_advisory.py
conbrad: You mean something like this?

`geom = func.ST_Transform(func.ST_GeomFromText(geometry, srid), 3005)`

Since 3005 is what we're storing in `advisory_shapes`.

brettedw: Ya, could even do the transform conditionally if srid != 3005 🤷🏻‍♂️

api/alembic/versions/c9e46d098c73_critical_hours.py
```diff
@@ -0,0 +1,57 @@
...
+
+# revision identifiers, used by Alembic.
+revision = "c9e46d098c73"
+down_revision = "6910d017b626"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
```
dgboss, line 21:
```suggestion
```

api/app/utils/geospatial.py
conbrad: Done in: https://github.com/bcgov/wps/pull/3849/commits/af4bd2af7d9680afcf95ab7b6e8537fcb87cdbe9

api/alembic/versions/c9e46d098c73_critical_hours.py
conbrad: Done in: https://github.com/bcgov/wps/pull/3849/commits/af4bd2af7d9680afcf95ab7b6e8537fcb87cdbe9

api/app/tests/auto_spatial_advisory/test_critical_hours.py
```diff
@@ -0,0 +1,140 @@
...
+                mock_station,
+                critical_hours_inputs=CriticalHoursInputs(
+                    dailies_by_station_id=dailies_by_station_id, yesterday_dailies_by_station_id={}, hourly_observations_by_station_code=hourlies_by_station_code
+                ),
+            )
+            == False
+        )
+
+
+def test_check_station_invalid_missing_daily():
```
dgboss, line 59: It looks like this test might also be checking for missing hourly instead of missing dailies?

api/app/tests/auto_spatial_advisory/test_critical_hours.py
```diff
@@ -0,0 +1,140 @@
...
+@pytest.mark.parametrize(
+    "start_times, expected_start_time",
+    [
+        ([1, 2], 1),
+        ([1, 2, 3], math.floor(np.percentile([1, 2, 3], 25))),
+    ],
+)
+def test_determine_start_time(start_times, expected_start_time):
+    """
+    Given a list of start times, choose them minimum if less than 3, otherwise the 25th percentile
```
dgboss, line 106:
```suggestion
    Given a list of start times, choose the minimum if less than 3, otherwise the 25th percentile
```

api/app/db/crud/auto_spatial_advisory.py
dgboss: If we're going to do the transform here (which is a great idea), I think we can get rid of the `PointTransformer` class in `geospatial.py`.

api/app/tests/utils/test_geospatial.py
```diff
@@ -3,7 +3,7 @@
 from osgeo import gdal
 import numpy as np
 
-from app.utils.geospatial import raster_mul, warp_to_match_extent
+from app.utils.geospatial import PointTransformer, raster_mul, warp_to_match_extent
```
dgboss, line 6: Unused import: `PointTransformer'

api/app/auto_spatial_advisory/critical_hours.py
```diff
@@ -0,0 +1,457 @@
...
+    logger.info(f"delta count before and after calculating critical hours: {delta}")
+
+
+#### - Helper functions for local testing of critical hours calculations.
+
+
+async def start_critical_hours():
+    async with get_async_write_session_scope() as db_session:
+        result = await get_most_recent_run_parameters(db_session, RunTypeEnum.actual, date(2024, 8, 8))
+        await calculate_critical_hours(result[0].run_type, result[0].run_datetime, result[0].for_date)
```
dgboss, line 435: My bad here. The call to `calculate_critical_hours` doesn't need to be nested within this block unless we want to pass the `db_session` into `calculate_critical_hours` and save ourselves from having to create another session in that function. Sorry.

dgboss: Oh wait, we do need to create the session in `calculate_critical_hours` as that gets called from the nats consumer.

api/app/auto_spatial_advisory/critical_hours.py
```diff
@@ -0,0 +1,457 @@
...
+            for_date,
+        )
+
+        if len(critical_hours_by_fuel_type) > 0:
+            critical_hours_by_zone_and_fuel_type[zone_key] = critical_hours_by_fuel_type
+
+    for zone_id, critical_hours_by_fuel_type in critical_hours_by_zone_and_fuel_type.items():
+        await save_critical_hours(db_session, zone_id, critical_hours_by_fuel_type, run_parameters_id)
+
+    return critical_hours_by_zone_and_fuel_type
```
dgboss, line 385: Again my bad, no need to return anything from this function.

api/app/auto_spatial_advisory/critical_hours.py
```diff
@@ -0,0 +1,457 @@
...
+async def calculate_critical_hours_by_zone(db_session: AsyncSession, header: dict, stations_by_zone: Dict[int, List[WFWXWeatherStation]], run_parameters_id: int, for_date: date):
+    """
+    Calculates critical hours for fire zone units by heuristically determining critical hours for each station in the fire zone unit that are under advisory conditions (>4k HFI).
+
+    :param db_session: An async database session.
+    :param header: An authorization header for making requests to WF1.
+    :param stations_by_zone: A dictionary of lists of stations in fire zone units keyed by fire zone unit id.
+    :param run_parameters_id: The RunParameters object (ie. the SFMS run).
+    :param for_date: The date critical hours are being calculated for.
+    :return: A dictionary of critical hours by fuel type per fire zone unit keyed by fire zone unit id.
```
dgboss, line 364: Can remove this docstring about what is returned as this function doesn't actually need to return anything.

api/app/auto_spatial_advisory/critical_hours.py
conbrad: We need it to be nested here because of the way sqlalchemy lazily loads data when records are accessed: https://docs.sqlalchemy.org/en/20/errors.html#error-bhk3

Without nesting we get an error `Instance <RunParameters at 0x308e28190> is not bound to a Session; attribute refresh operation cannot proceed.`

api/app/tests/auto_spatial_advisory/test_critical_hours.py
conbrad: Updated tests, hopefully they read more clearly now.

api/app/auto_spatial_advisory/critical_hours.py
```diff
@@ -0,0 +1,457 @@
...
+from app.utils.geospatial import PointTransformer
+from app.utils.time import get_hour_20_from_date, get_julian_date
+from app.wildfire_one import wfwx_api
+from app.wildfire_one.schema_parsers import WFWXWeatherStation
+
+logger = logging.getLogger(__name__)
+
+
+@dataclass(frozen=True)
+class CriticalHoursInputs:
```
dgboss, line 43: ❤️

dgboss (commented):
I can't actually approve this as I opened the original PR, but I think it looks good with your changes. Thanks again for picking this up!

brettedw (approved):
Nice work to both of you!

PR #3860, opened 2024-08-17
frontend to vite

### LINKED ISSUE #3785: Migrate frontend build to vite
**Describe the task**
To speed up builds and reloading we can migrate to vite and away from crac: https://vitejs.dev/

**Acceptance Criteria**
- [x] Frontend builds and deploys with vite as the build tool
- [x] All tests pass (may require migrating from jest to vitest: https://vitest.dev/)
- [x] All code coverage from jest and cypress tests are reported to codecov still

**Additional context**
- You can take a look at a previous attempt here: https://github.com/bcgov/wps/tree/experiment/migrate-to-vite, it's missing the last AC above requiring code coverage.

jmetev1:
Things I did
1. changed to vite and vitest
1.  removed some unneeded stuff from integration workflow
2. moved to node alpine bc couldn’t make it work with corepack and new type= module nature of vite
3. renamed env variables for vite
4. add global browser fetch in tests which was happening before behind the scenes with CRA
5. not running       "src/features/auth/slices/authenticationSlice.test.ts" bc ES Modules cannot be stubbed so need to find other solution for sinon.

Outcomes
1. can’t use require anymore unless you name file .cjs or .cts. Everything is default import export now. 
closes #3785

# Test Links:
[Landing Page](https://wps-pr-3860-e1e498-dev.apps.silver.devops.gov.bc.ca/)
[MoreCast](https://wps-pr-3860-e1e498-dev.apps.silver.devops.gov.bc.ca/morecast)
[Percentile Calculator](https://wps-pr-3860-e1e498-dev.apps.silver.devops.gov.bc.ca/percentile-calculator)
[C-Haines](https://wps-pr-3860-e1e498-dev.apps.silver.devops.gov.bc.ca/c-haines)
[FireBat](https://wps-pr-3860-e1e498-dev.apps.silver.devops.gov.bc.ca/fire-behaviour-calculator)
[FireBat bookmark](https://wps-pr-3860-e1e498-dev.apps.silver.devops.gov.bc.ca/fire-behaviour-calculator?s=266&f=c5&c=NaN&w=20,s=286&f=c7&c=NaN&w=16,s=1055&f=c7&c=NaN&w=NaN,s=305&f=c7&c=NaN&w=NaN,s=344&f=c5&c=NaN&w=NaN,s=346&f=c7&c=NaN&w=NaN,s=328&f=c7&c=NaN&w=NaN,s=1399&f=c7&c=NaN&w=NaN,s=334&f=c7&c=NaN&w=NaN,s=1082&f=c3&c=NaN&w=NaN,s=388&f=c7&c=NaN&w=NaN,s=309&f=c7&c=NaN&w=16,s=306&f=c7&c=NaN&w=NaN,s=1029&f=c7&c=NaN&w=NaN,s=298&f=c7&c=NaN&w=NaN,s=836&f=c7&c=NaN&w=NaN,s=9999&f=c7&c=NaN&w=NaN)
[Auto Spatial Advisory (ASA)](https://wps-pr-3860-e1e498-dev.apps.silver.devops.gov.bc.ca/auto-spatial-advisory)
[HFI Calculator](https://wps-pr-3860-e1e498-dev.apps.silver.devops.gov.bc.ca/hfi-calculator)


web/.yarnrc.yml
```diff
@@ -1 +1,2 @@
 nodeLinker: node-modules
+enableHardenedMode: false
```
conbrad, line 2: Why this change?

web/index.html
```diff
@@ -2,10 +2,7 @@
 <html lang="en">
   <head>
     <meta name="viewport" content="width=device-width, initial-scale=1.0">
-    <script src="config.js"></script>
-    <script type="text/javascript">
-      window.env = config
-    </script>
+    <script>console.log('abandon all hope')</script>
```
conbrad, line 5:
```suggestion
```

🤣

web/index.html
```diff
@@ -21,15 +18,7 @@
...
-      This HTML file is a template.
-      If you open it directly in the browser, you will see an empty page.
-
-      You can add webfonts, meta tags, or analytics to this file.
-      The build step will place the bundled scripts into the <body> tag.
-
-      To begin the development, run `npm start` or `yarn start`.
-      To create a production bundle, use `npm run build` or `yarn build`.
-    -->
+    <div id="i be trippin"></div>
```
conbrad, line 21:
```suggestion
```

web/index.html
```diff
@@ -21,15 +18,7 @@
...
+    <script type="module" src="/src/index.tsx"></script>
```
conbrad, line 22: Is this a change required by vite?

web/package.json
```diff
@@ -51,73 +53,55 @@
...
-    "react-dom": "^18.2.0",
+    "react": "^18.3.1",
+    "react-dom": "^18.3.1",
     "react-redux": "^8.0.0",
     "react-router-dom": "^6.22.3",
     "recharts": "^2.1.8",
-    "typescript": "^5.0.0"
+    "typescript": "^5.2.2",
+    "vitest": "^2.0.5",
+    "whatwg-fetch": "^3.6.20"
```
conbrad, line 63: What's this dependency used for?

web/package.json
```diff
@@ -51,73 +53,55 @@
...
   "scripts": {
-    "start": "CI=true craco start",
-    "start:cypress": "export $(cat .env.cypress | xargs) && craco --max_old_space_size=4096 -r @cypress/instrument-cra start",
-    "build": "craco --max_old_space_size=2048 build",
-    "build:prod": "craco --max_old_space_size=2048 build && yarn sentry:sourcemaps",
-    "test": "craco test --transformIgnorePatterns \"node_modules/(?!ol)/\"",
+    "dev": "vite",
+    "preview": "vite preview",
+    "dev:cypress": "export $(cat .env.cypress | xargs) && yarn dev",
+    "build:prod": "tsc -b && vite build",
```
conbrad, line 69: I think we need to keep sentry source map generation and upload here?

web/package.json
```diff
@@ -51,73 +53,55 @@
...
     "eslint": "^9.7.0",
     "eslint-plugin-react-hooks": "^4.6.2",
+    "eslint-plugin-react-refresh": "^0.4.7",
     "globals": "^15.8.0",
-    "react-scripts": "^5.0.1",
     "start-server-and-test": "^2.0.0",
-    "ts-sinon": "^2.0.2"
-  },
-  "jest": {
-    "collectCoverageFrom": [
```
conbrad, line 114: This is also probably why coverage dropped, does vitest not have a similar configuration option?

web/src/app/store.ts
```diff
@@ -1,6 +1,6 @@
 import { configureStore, AnyAction, ThunkAction } from '@reduxjs/toolkit'
 import thunk, { ThunkMiddleware } from 'redux-thunk'
-import rootReducer, { RootState } from 'app/rootReducer'
+import rootReducer, { RootState } from 'app/rootReducer.ts'
```
conbrad, line 3: Is this change required?

web/src/features/moreCast2/forecastDraft.test.ts
```diff
@@ -3,21 +3,22 @@ import { DraftMorecast2Rows } from 'features/moreCast2/interfaces'
 import { buildValidActualRow, buildValidForecastRow } from 'features/moreCast2/rowFilters.test'
 import { DateTime } from 'luxon'
 import * as DateUtils from 'utils/date'
+import { vi } from 'vitest'
```
conbrad, line 6: Should we also import `expect`, `it`, etc also?

web/src/index.tsx
```diff
@@ -1,12 +1,11 @@
 import React from 'react'
 import * as ReactDOMClient from 'react-dom/client'
 import { Provider } from 'react-redux'
-import App from 'app/App'
+import App from 'app/App.tsx'
```
conbrad, line 4: Is this change required?

web/src/types/window.d.ts
conbrad, line 1: Should these be `VITE_` prefixes instead of `REACT_APP`?

web/src/utils/env.ts
```diff
@@ -27,6 +27,7 @@ if (process.env.NODE_ENV === 'production') {
     KC_AUTH_URL: window.env.REACT_APP_KEYCLOAK_AUTH_URL,
     KC_REALM: window.env.REACT_APP_KEYCLOAK_REALM,
     KC_CLIENT: window.env.REACT_APP_KEYCLOAK_CLIENT,
+    WF1_AUTH_URL: window.env.REACT_APP_WF1_AUTH_URL,
```
conbrad, line 30:
```suggestion
```
I don't think we need `WF1_AUTH_URL` anymore

web/tsconfig.node.json
```diff
@@ -0,0 +1,13 @@
+{
+  "compilerOptions": {
+    "composite": true,
+    "tsBuildInfoFile": "./node_modules/.tmp/tsconfig.node.tsbuildinfo",
```
conbrad, line 4: Is this option required? Seems to reference a temporary file

web/vite.config.ts
```diff
@@ -0,0 +1,79 @@
...
+      },
+    }),
+    , svgr(),
+    istanbul({
+      include: 'src/*',
+      exclude: ['node_modules', 'test/'],
+      extension: ['.js', '.ts'],
+      cypress: true
+    }),
+    sentryVitePlugin({
```
conbrad, line 46: Does this generated and upload source maps for us?

web/vite.config.ts
```diff
@@ -0,0 +1,79 @@
...
+      authToken: process.env.SENTRY_AUTH_TOKEN,
+      sourcemaps: {
+        filesToDeleteAfterUpload: ['**/*.map'],
+      },
+    }),
+  ],
+  server: {
+    port: Number(process.env.PORT) || 3000,
+    watch: {
+      ignored: [
```
conbrad, line 58: Should we also ignore `*.test.*` files?

conbrad (commented):
Looking good! Mostly questions. Would you mind also updating the PR description to include the summary of the changes?

Top level question, is it required that we move ts config files to common js files?

Dockerfile.web
```diff
@@ -9,10 +9,10 @@ ARG NODE_OPTIONS="--v8-pool-size=4"
 # Pull from local registry - we can't pull from docker due to limits.
 # see https://catalog.redhat.com/software/containers/ubi9/nodejs-20/64770ac7a835530172eee6a9 for
 # details
-FROM registry.access.redhat.com/ubi9/nodejs-20 as static
-
+FROM node:20-alpine AS static
```
conbrad, line 12: Is this required? Will the web still build with the old base image from redhat?

web/.yarnrc.yml
jmetev1: temp thing for to make ci run faster. taking out now.

web/index.html
jmetev1: yep. it's ok?

conbrad: Yeah just curious if you have a link to the docs about it or some background info.

conbrad: I think so, just wondering if there's any background info on why

web/src/features/moreCast2/forecastDraft.test.ts
jmetev1: i agree this looks a little funky. but expect and it are coming from setuptests.ts already

web/package.json
conbrad: Unless there's a vite plugin doing it now.

web/src/features/moreCast2/forecastDraft.test.ts
conbrad: I see, why do other tests require imports?

Dockerfile.web
jmetev1: Nope. There's just something about the redhat node images. I've tried a lot of things. The error is "Dynamic require of "util" is not supported". And googling it is not very helpful. So we can't use docker bc we don't have a paid account and since it's not cached we will exceed rate limit? It seems like somehow yarn/corepack is part of the problem. Want to switch to npm instead of yarn? I don't have a lot more ideas.

web/index.html
jmetev1: the beginning of this is good intro. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Modules Vite is all about using modern native  esm support.

Dockerfile.web
conbrad: All good just wanted to clarify. Let's keep this and adjust if we run into rate limits.

web/vite.config.ts
jmetev1: i added these for cypress tests running. When that is running these directories get changed and so dev server restarts. But the test files won't get changed so I think we're ok.

conbrad: Does this config only apply to cypress tests though? If a developer is running the vite server locally and making changes to component tests will this trigger dev server restarts?

conbrad:
There's a bunch of sonarcloud-flagged unused imports from `vitest` in test files, e.g. https://sonarcloud.io/project/issues?pullRequest=3860&open=AZFsFi30uttMHqCURf1j&id=bcgov_wps

Would you mind going through those and removing unused imports?

.github/workflows/integration.yml
```diff
@@ -191,8 +191,6 @@ jobs:
         with:
           # For sonar-scanner to work properly we can't use a shallow fetch.
           fetch-depth: 0
-      - name: Setup kernel for react, increase watchers
```
jmetev1, line 194: i don't think this is needed.

.vscode/settings.json
```diff
@@ -8,7 +8,6 @@
         ".github": false,
         "**/.pytest_cache": true,
         "**/__pycache__**": true,
-        "web/node_modules": true,
```
jmetev1, line 11: just a personal preference. I'm always going in node modules. But other people want it hidden?

web/.env.cypress
```diff
@@ -1,9 +1,9 @@
 BROWSER=none
 PORT=3030
-REACT_APP_API_BASE_URL=http://localhost:8080/api
-REACT_APP_HIDE_DISCLAIMER=false
-REACT_APP_KEYCLOAK_AUTH_URL=https://keycloak-auth-url.ca
-REACT_APP_SM_LOGOUT_URL=https://sm-logout-url.ca
-REACT_APP_KEYCLOAK_REALM=8wl6x4cp
-REACT_APP_KEYCLOAK_CLIENT=wps-web
-REACT_APP_PMTILES_BUCKET=https://nrs.objectstore.gov.bc.ca/gpdqha/psu/pmtiles/
```
jmetev1, line 9: i hadn't noticed this. so cypress is using the real bucket but not other tests

.vscode/settings.json
conbrad: I think it's fine. Where it can kind of get annoying is if you're doing a global string search in the repo and node modules are indexed those results can cloud what you're looking for.

web/.env.cypress
conbrad: Other tests could use the real bucket (these are public, I forgot)

web/package.json
jmetev1: yep it's in vitest.config test/coverage

web/src/features/moreCast2/forecastDraft.test.ts
jmetev1: hmmmm. good question. will look.

web/src/types/window.d.ts
jmetev1: these are the ones that are coming from the config file that is replaced when deployed. so we would have to rename them there and here. we can do that if you like that better

jmetev1:
> Looking good! Mostly questions. Would you mind also updating the PR description to include the summary of the changes?
> 
> Top level question, is it required that we move ts config files to common js files?

Yep I can add more details.
If config or all other files use require then they have to either be rewritten or they have to be renamed to .cjs. Bc package.json now has type:module, all .js or .ts files are now assumed to be esm compliant. I rewrote some and others like cypress config I couldn't make work so just renamed to cjs.

jmetev1:
My numbers are improving now that one less test is disabled!
image

.vscode/settings.json
jmetev1: do you mean when you're searching using vscode? on my end at least if this button is clicked then it will ignore node_modules 
image

web/vite.config.ts
jmetev1: yea. is that a problem? it's like instantatenous. but i can add tests to ignored.

web/vite.config.ts
jmetev1: i went back to old way

web/tsconfig.node.json
jmetev1: removed

web/src/app/store.ts
jmetev1: removed

web/package.json
jmetev1: will add to summary notes

web/src/features/fba/components/map/fbaMap.test.tsx
```diff
@@ -3,10 +3,16 @@ import store from 'app/store'
...
 import { RunType } from 'features/fba/pages/FireBehaviourAdvisoryPage'
 import { DateTime } from 'luxon'
-import React from 'react'
+
 import { Provider } from 'react-redux'
+// @ts-ignore
+import { fetch } from 'whatwg-fetch'
 
 describe('FBAMap', () => {
+  beforeEach(() => {
```
conbrad, line 12: Why do we need to overwrite the global fetch function?

jmetev1: in create react app version this was also being done, just under the hood. without this, when olpmtiles does a fetch to that s3 bucket it will expect browser fetch but will actually get node built in fetch and error out. this was a fascinating discovery that took me so many hours to figure out. other options would be to mock some sub components i suppose. or the whole fetch response.

conbrad: Interesting... would it make sense to replace the node fetch in an `after` function? Or perhaps just set this as the global fetch for all tests in `setupTests`?

.vscode/settings.json
conbrad: Yeah in vscode, interesting ok, will check that button out.

web/vite.config.ts
conbrad: Not sure if it's a problem, we could add that after if it becomes one.

web/src/types/window.d.ts
conbrad: Actually if we could remove this `window.d.ts` file altogether that'd be nice, one less place to add env vars.

web/src/features/fba/components/map/fbaMap.test.tsx
jmetev1: sure thing. done

web/.env.cypress
jmetev1: done

web/src/utils/env.ts
jmetev1: done

conbrad:
> My numbers are improving now that one less test is disabled! image

Nice! How many tests are disabled? The code coverage reports make sense when I'm looking at the ones with coverage or partial coverage, but a bunch of missing coverage around HFI Calculator signals that either those cypress tests are disabled or the coverage report isn't being merged properly.

jmetev1:
> > My numbers are improving now that one less test is disabled! image
> 
> Nice! How many tests are disabled? The code coverage reports make sense when I'm looking at the ones with coverage or partial coverage, but a bunch of missing coverage around HFI Calculator signals that either those cypress tests are disabled or the coverage report isn't being merged properly.

image
whoops!

web/src/whatwg-fetch.d.ts
```diff
@@ -0,0 +1 @@
+declare module "whatwg-fetch";
```
jmetev1, line 1: sometimes typescript just wears me out. i don't even know what this does. i just want to polyfill fetch in the test. other suggestions accepted.

conbrad: I imagine `whatwg-fetch` doesn't ship with type definitions, I think this is fine since we're just using it for tests. What errors were you seeing?

conbrad: https://www.npmjs.com/package/@types/whatwg-fetch says "fetch types are now provided by '--lib dom'"

web/src/types/window.d.ts
jmetev1: like right now? than typescript errors out. i do really want to fix the whole env situation but i'm trying not to avoid mission creep.

conbrad: Ok yeah no problem.

web/package.json
```diff
@@ -34,13 +35,15 @@
     "@types/webpack-env": "^1.15.1",
     "@typescript-eslint/eslint-plugin": "^8.0.0",
     "@typescript-eslint/parser": "^8.0.0",
+    "@vitest/coverage-istanbul": "^2.0.5",
```
conbrad, line 38: Nit: can this go in `devDependencies`?

web/vite.config.ts
```diff
@@ -0,0 +1,68 @@
...
+import svgr from 'vite-plugin-svgr'
+import istanbul from 'vite-plugin-istanbul'
+
+export default defineConfig({
+  build: {
+    sourcemap: true,
+    outDir: 'build'
+  },
+  optimizeDeps: {
+    include: ['@mui/material/Tooltip', '@emotion/styled/base', '@mui/material/Unstable_Grid2']
```
conbrad, line 14: Were these added to avoid the vite build errors you mentioned earlier?

jmetev1: yep

conbrad: Ok cool, is it possible to add a comment in this config ts file explaining that a little bit with links to the issue?

web/package.json
jmetev1: indulge a digression. I have this conviction that there is no distinction between dependencies and dev dependencies in a front end non library project. like the whole thing gets built. there are no real dependencies after that. right? also by any measure i feel like most of the classification here is wrong. should typescript, and sentry cli and all the types also be moved to dev deps?

conbrad: Digression indulged. Yes seems like `devDependencies` also get installed since we don't install in the web image with the `--production` flag. Perhaps we should add that but can be deferred in a follow up PR.

conbrad (approved):
Looks great, amazing work thanks for the considerable amount of effort that went into this! Builds are faster and we can sever our dependency on `cra`. 

On my side just the couple comments I added this morning to be resolved, and the injected error to test sentry source maps and I'm happy. I'll let @dgboss and @brettedw carve out some time to review further.

web/src/app/store.ts
```diff
@@ -14,16 +14,6 @@ const store = configureStore({
     getDefaultMiddleware({ immutableCheck: false, serializableCheck: false }).concat(thunkMiddleware)
 })
 
-if (process.env.NODE_ENV === 'development' && module.hot) {
```
dgboss, line 17: Do you know if we get the equivalent behaviour with vite out of the box? Or will we have to re-start our debugging session if we are making changes to a reducer?

web/src/features/landingPage/components/Sidebar.tsx
```diff
@@ -15,8 +16,8 @@ import Typography from '@mui/material/Typography'
 import useMediaQuery from '@mui/material/useMediaQuery'
 import SidebarToolList from 'features/landingPage/components/SidebarToolList'
 import Subheading from 'features/landingPage/components/Subheading'
-import { ReactComponent as MsTeamsIcon } from 'features/landingPage/images/msTeams.svg'
-import { ReactComponent as MiroIcon } from 'features/landingPage/images/miro.svg'
+import MsTeamsIcon from 'features/landingPage/images/msTeams.svg?react'
```
dgboss, line 19: Interesting syntax! I haven't seen this before. Nice!

dgboss (approved):
Amazing work! Thank you for all the time you put into this!

web/src/app/store.ts
jmetev1: hmmm. I feel like vite should handle this well since a big focus of it is not having to rebundle stuff https://vitejs.dev/guide/why.html#why-vite . But i haven't done much reducer stuff lately. Can you break down for me what problem this was solving? you don't want to lose some state in redux?

web/src/features/landingPage/components/Sidebar.tsx
jmetev1: it's a little freaky eh

web/package.json
jmetev1: ahh but that wouldn't work. to do the build we need most of the stuff that is in dev dependencies. i think it's a distinction that makes sense in backend projects and just got carried into frontend somehow.

jmetev1: but the multistage build means node modules get's left behind anyway so all good.

web/vite.config.ts
jmetev1: done

web/src/whatwg-fetch.d.ts
jmetev1: that seems to mean i need tsconfig compileroptions: [dom]. but i already had that and still got warnings.

web/src/app/store.ts
dgboss: I was curious more than anything so not a big deal. I think you are correct in that it prevented loss of state in redux. I don't think we need to do anything or look into this any further. Thanks!

web/src/whatwg-fetch.d.ts
conbrad: Ok np

brettedw (approved):
Thank you for the effort you put into this @jmetev1, love it!

PR #3918, opened 2024-09-11
Add fuel types and critical hours to advisory

dgboss:
- advisory bulletin only displays info for selected fire zone unit
- bulletin includes up to top 3 fuel types
- critical hours start time is the earliest start time from the top three fuel types
- critical hours end time is the latest end time from the three top fuel types

### DIFF
--- a/web/src/features/fba/components/infoPanel/AdvisoryText.tsx
+++ b/web/src/features/fba/components/infoPanel/AdvisoryText.tsx
@@ -1,11 +1,12 @@
 import { Box, Typography } from '@mui/material'
-import { FireCenter, FireShapeAreaDetail } from 'api/fbaAPI'
+import { FireCenter, FireShape, FireZoneFuelStats } from 'api/fbaAPI'
 import { DateTime } from 'luxon'
-import React from 'react'
+import React, { useEffect, useState } from 'react'
 import { useSelector } from 'react-redux'
 import { selectProvincialSummary } from 'features/fba/slices/provincialSummarySlice'
+import { selectFireCentreHFIFuelStats } from '@/app/rootReducer'
 import { AdvisoryStatus } from 'utils/constants'
-import { groupBy } from 'lodash'
+import { isEmpty, isUndefined } from 'lodash'
 import { calculateStatusText } from '@/features/fba/calculateZoneStatus'
 
 interface AdvisoryTextProps {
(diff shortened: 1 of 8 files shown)


web/src/features/fba/components/infoPanel/AdvisoryText.tsx
```diff
@@ -1,39 +1,99 @@
...
+      setSelectedFireZoneUnitTopFuels([])
+      setMinStartTime(undefined)
+      setMaxEndTime(undefined)
+      return
+    }
+    const allZoneUnitFuelStats = fireCentreHFIFuelStats?.[selectedFireCenter.name]
+    const selectedZoneUnitFuelStats = allZoneUnitFuelStats?.[selectedFireZoneUnit.fire_shape_id] ?? []
+    const sortedFuelStats = [...selectedZoneUnitFuelStats].sort(sortByArea)
+    let topFuels: FireZoneFuelStats[] = []
+    for (let i = 0; i < 3 && i < sortedFuelStats.length; i++) {
```
conbrad, line 60: Could simplify with `take(sortedFuelStats, 3)`

https://lodash.com/docs/4.17.15#take

conbrad (approved):

brettedw:
Looking awesome, learned a few things. Only thing I noticed was that if there is fuel stats but no critical hours for some reason the text reads something like: 
> There is a fire behaviour Advisory in effect for N2-Invermere Fire Zone between null:00 and null:00 for fuel types M-1/M-2, C-3 and C-7.

dgboss:
> Looking awesome, learned a few things. Only thing I noticed was that if there is fuel stats but no critical hours for some reason the text reads something like:
> 
> > There is a fire behaviour Advisory in effect for N2-Invermere Fire Zone between null:00 and null:00 for fuel types M-1/M-2, C-3 and C-7.

Huh? It's not supposed to do that! I'll re-visit the logic. Might be a null check versus an undefined check thing. Thanks!

web/src/features/fba/components/infoPanel/AdvisoryText.tsx
dgboss: Thanks for the tip! Changed in https://github.com/bcgov/wps/pull/3918/commits/26fbecdf625ce33c0fc99a4d700823ce92f6a1ce

dgboss:
> > Looking awesome, learned a few things. Only thing I noticed was that if there is fuel stats but no critical hours for some reason the text reads something like:
> > > There is a fire behaviour Advisory in effect for N2-Invermere Fire Zone between null:00 and null:00 for fuel types M-1/M-2, C-3 and C-7.
> 
> Huh? It's not supposed to do that! I'll re-visit the logic. Might be a null check versus an undefined check thing. Thanks!

Yep, was a null vs. undefined check. Fixed in https://github.com/bcgov/wps/pull/3918/commits/26fbecdf625ce33c0fc99a4d700823ce92f6a1ce

brettedw (approved):

PR #3919, opened 2024-09-11
Morecast: Validation on editable fields

### LINKED ISSUE #3571: MoreCast: Range limits on weather values
**As a** Weather Forecaster

**I need** range limits on weather values

**So That** data entry errors are flagged and it is not possible to provide out of range values in my forecast

**Acceptance Criteria**
- [x] Given I enter a weather value out of range, When I leave the cell, Then the cell is highlighted and a warning message appears in a hover tooltip.
- [ ] Drafts still save out of range inputs but above highlight and warning hold
- [x] A user cannot submit a forecast with any inputs that are out of range, clicking the the submit forecast button will provide an alert using the same toast popup that advises the user to fix inputs in any highlighted cells: "Please review the highlighted fields—XX errors found."

**Additional Context**
- Temperature -60 to +60 deg C
- Relative Humidity 0 to 100%
- Precipitation 0 to 200 mm
- Wind Speed 0 to 120 km/h
- Wind Direction 0 to 360 degrees

- Behaviour mimics FireBAT input validation (only 1 tooltip though!)

**Definition of Done**
https://github.com/bcgov/wps/wiki/Definition-of-Done

conbrad:
Pulls in the `preProcessEditCellProps` callback for column definitions to set the error prop if validation doesn't pass for forecast fields.

Adds `EditInputCell` as the render component for cell editing that reacts to the error flag being set with a red tooltip and border.

TODO:
- Make the tooltip error less ambiguous
- Add tests

Closes #3571

### DIFF
--- a/web/src/features/moreCast2/components/MoreCast2Column.tsx
+++ b/web/src/features/moreCast2/components/MoreCast2Column.tsx
@@ -1,4 +1,4 @@
-import { GridValueFormatterParams } from '@mui/x-data-grid-pro'
+import { GridEditCellProps, GridPreProcessEditCellProps, GridValueFormatterParams } from '@mui/x-data-grid-pro'
 import { DateTime } from 'luxon'
 import {
   ColDefGenerator,
@@ -136,7 +136,10 @@ export class IndeterminateField implements ColDefGenerator, ForecastColDefGenera
     readonly headerName: string,
     readonly type: 'string' | 'number',
     readonly precision: number,
-    readonly includeBias: boolean
+    readonly includeBias: boolean,
+    readonly preProcessEditCellProps?: (
+      params: GridPreProcessEditCellProps
+    ) => GridEditCellProps | Promise<GridEditCellProps>
   ) {
     this.colDefBuilder = new ColumnDefBuilder(
       this.field,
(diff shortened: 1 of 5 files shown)


web/src/features/moreCast2/components/MoreCast2Column.tsx
```diff
@@ -152,28 +155,86 @@ export class IndeterminateField implements ColDefGenerator, ForecastColDefGenera
...
-export const windDirForecastField = new IndeterminateField('windDirection', WIND_DIR_HEADER, 'number', 0, true)
+export const tempForecastField = new IndeterminateField(
+  'temp',
+  TEMP_HEADER,
+  'number',
+  0,
+  true,
+  (params: GridPreProcessEditCellProps) => {
+    const hasError =
+      params.props.value < -60 || params.props.value > 60 ? 'Temp must be between -60 and 60 degrees C' : false
```
dgboss, line 198: So `hasError`/`error` is interpreted as truthy when assigned a string instead of a boolean and that string gets displayed in the tooltip?

conbrad: Haha yes, hacky, but I'm going to refactor this whole approach I think because setting error values through `preProcessEditCellProps` doesn't persist for cells in view mode so errors are only showing up while editing.

dgboss: I was just about to drop a comment on that!

web/src/features/moreCast2/components/MoreCast2Column.tsx
```diff
@@ -152,28 +155,95 @@ export class IndeterminateField implements ColDefGenerator, ForecastColDefGenera
...
-export const windSpeedForecastField = new IndeterminateField('windSpeed', WIND_SPEED_HEADER, 'number', 0, true)
-export const precipForecastField = new IndeterminateField('precip', PRECIP_HEADER, 'number', 1, true)
+export const tempForecastField = new IndeterminateField(
+  'temp',
+  TEMP_HEADER,
+  'number',
+  0,
+  true,
+  (params: GridPreProcessEditCellProps) => {
+    const error = params.props.value < -60 || params.props.value > 60 ? 'Temp must be between -60 and 60 degrees C' : ''
```
dgboss, line 197: nit: Maybe do 'Temp must be between -60°C and 60°C'

or spell out Celsisus like

'Temp must be between -60 and 60 degrees Celsius'

web/src/features/moreCast2/components/MoreCast2Column.tsx
```diff
@@ -152,28 +155,95 @@ export class IndeterminateField implements ColDefGenerator, ForecastColDefGenera
...
+)
+export const windSpeedForecastField = new IndeterminateField(
+  'windSpeed',
+  WIND_SPEED_HEADER,
+  'number',
+  0,
+  true,
+  (params: GridPreProcessEditCellProps) => {
+    const error =
+      params.props.value < 0 || params.props.value > 120 ? 'Wind speed must be between 0 and 120 degrees' : ''
```
dgboss, line 232:
```suggestion
      params.props.value < 0 || params.props.value > 120 ? 'Wind speed must be between 0 and 120 kph' : ''
```

web/src/features/moreCast2/components/editInputCell.test.tsx
```diff
@@ -0,0 +1,163 @@
...
+
+// Mock the setInputValid action creator
+vi.mock('@/features/moreCast2/slices/validInputSlice', () => ({
+  setInputValid: vi.fn().mockReturnValue({ type: 'mock/setInputValid' })
+}))
+const mockDispatch = vi.fn()
+
+beforeEach(() => {
+  vi.clearAllMocks()
+  ;(useDispatch as jest.Mock).mockReturnValue(mockDispatch)
```
dgboss, line 48:
```suggestion
  (useDispatch as jest.Mock).mockReturnValue(mockDispatch)
```

web/src/features/moreCast2/components/MoreCast2Column.tsx
conbrad: Fixed in: https://github.com/bcgov/wps/pull/3919/commits/2d168d46afd8553718057507c0268f3553f6c2e9

web/src/features/moreCast2/components/MoreCast2Column.tsx
conbrad: Fixed in: https://github.com/bcgov/wps/pull/3919/commits/2d168d46afd8553718057507c0268f3553f6c2e9

web/src/features/moreCast2/components/editInputCell.test.tsx
conbrad: Doesn't build/run without it!

dgboss:
It's looking really good! 
I noticed that validation errors/tooltip doesn't persist when you go from the weather parameter tabs to the Forecast Summary tab. Realistically this probably isn't an issue as I would expect a forecaster to correct validation errors as they enter them.

dgboss:
Oh, should we add validation for the grass curing fields as well?

conbrad:
> It's looking really good! I noticed that validation errors/tooltip doesn't persist when you go from the weather parameter tabs to the Forecast Summary tab. Realistically this probably isn't an issue as I would expect a forecaster to correct validation errors as they enter them.



> It's looking really good! I noticed that validation errors/tooltip doesn't persist when you go from the weather parameter tabs to the Forecast Summary tab. Realistically this probably isn't an issue as I would expect a forecaster to correct validation errors as they enter them.

Nice catch! I adjusted the code to share the validation logic between edit and view mode cells: https://github.com/bcgov/wps/pull/3919/commits/59a301e2531f7637258bf514448e54029830146c

brettedw:
Looks great, nice work! Only comment would be same as Darren's
> Oh, should we add validation for the grass curing fields as well?

brettedw:
Just noticed there's a number of code smells. Unused imports in a few cases, otherwise just complaining about using optional chain expressions. I'm fine either way

conbrad:
> Just noticed there's a number of code smells. Unused imports in a few cases, otherwise just complaining about using optional chain expressions. I'm fine either way

Thanks fixed, part of it was a bunch of duplicated code between forecast and grass cure cells that I deduped in a shared `ValidatedForecastCell` component.

brettedw (approved):
Looks great to me! Nice work!

dgboss (approved):
This is great! Only one tiny nit that could be addressed in a followup PR. The tooltip position becomes a little wonky when toggling between the Forecast Summary tab and the other tabs. Definitely not worth holding up the PR for.

PR #3936, opened 2024-09-16
Morecast: Add validation feedback for empty but required fields

conbrad:
Renders red border on cells when a value is required but is empty. For these cells, only when a user hovers them will it provide the prompt, "Enter a value" in a tooltip, so as not to overwhelm the datagrid.
# Test Links:
[Landing Page](https://wps-pr-3936-e1e498-dev.apps.silver.devops.gov.bc.ca/)
[MoreCast](https://wps-pr-3936-e1e498-dev.apps.silver.devops.gov.bc.ca/morecast)
[Percentile Calculator](https://wps-pr-3936-e1e498-dev.apps.silver.devops.gov.bc.ca/percentile-calculator)
[C-Haines](https://wps-pr-3936-e1e498-dev.apps.silver.devops.gov.bc.ca/c-haines)
[FireBat](https://wps-pr-3936-e1e498-dev.apps.silver.devops.gov.bc.ca/fire-behaviour-calculator)
[FireBat bookmark](https://wps-pr-3936-e1e498-dev.apps.silver.devops.gov.bc.ca/fire-behaviour-calculator?s=266&f=c5&c=NaN&w=20,s=286&f=c7&c=NaN&w=16,s=1055&f=c7&c=NaN&w=NaN,s=305&f=c7&c=NaN&w=NaN,s=344&f=c5&c=NaN&w=NaN,s=346&f=c7&c=NaN&w=NaN,s=328&f=c7&c=NaN&w=NaN,s=1399&f=c7&c=NaN&w=NaN,s=334&f=c7&c=NaN&w=NaN,s=1082&f=c3&c=NaN&w=NaN,s=388&f=c7&c=NaN&w=NaN,s=309&f=c7&c=NaN&w=16,s=306&f=c7&c=NaN&w=NaN,s=1029&f=c7&c=NaN&w=NaN,s=298&f=c7&c=NaN&w=NaN,s=836&f=c7&c=NaN&w=NaN,s=9999&f=c7&c=NaN&w=NaN)
[Auto Spatial Advisory (ASA)](https://wps-pr-3936-e1e498-dev.apps.silver.devops.gov.bc.ca/auto-spatial-advisory)
[HFI Calculator](https://wps-pr-3936-e1e498-dev.apps.silver.devops.gov.bc.ca/hfi-calculator)


brettedw:
Not sure what is causing this, but sometimes locally I don't get a popup and the cursor flickers between pointer/text cursor, along with the mouse over row flickering as well. It seems to resolve when I move my cursor to the centre of the text field?

web/src/features/moreCast2/components/MoreCast2Column.tsx
```diff
@@ -180,6 +180,9 @@ export class IndeterminateField implements ColDefGenerator, ForecastColDefGenera
 }
 
 export const tempForecastField = new IndeterminateField('temp', TEMP_HEADER, 'number', 0, true, (value: string) => {
+  if (value === '') {
+    return 'Enter a value'
+  }
```
dgboss, line 185: What do you think about extracting this out into a function that can be used across all the forecast fields instead of repeating it?

conbrad:
> Not sure what is causing this, but sometimes locally I don't get a popup and the cursor flickers between pointer/text cursor, along with the mouse over row flickering as well. It seems to resolve when I move my cursor to the centre of the text field?

Aah it's because forecast cell is rendered as part of a grid, I'm going to have to move the Tooltip to wrap the grid for ForecastCell

web/src/features/moreCast2/components/MoreCast2Column.tsx
conbrad: Mixed feelings -- normally I'd agree to dedup, but it's a very small check and adding another level of indirection in this giant monstrosity is has me thinking twice lol.

dgboss:
I'm a little torn about always showing the red borders on cells that need data entry. A more typical pattern would be to validate once the submit/save button gets pressed and then display the border around cells that need it. It might be worth checking in with RP on this.

I'd also suggest updating the text in the warning toast message to say something about required fields being empty or something to that effect.

conbrad:
> I'm a little torn about always showing the red borders on cells that need data entry. A more typical pattern would be to validate once the submit/save button gets pressed and then display the border around cells that need it. It might be worth checking in with RP on this.
> 
> I'd also suggest updating the text in the warning toast message to say something about required fields being empty or something to that effect.

Perhaps the red border is a little aggressive for drawing attention to blank cells but I'd argue some sort of feedback before having the user attempt to submit is better than only seeing errors once you attempt to submit, to get the feedback as early as possible.

brettedw:
> > I'm a little torn about always showing the red borders on cells that need data entry. A more typical pattern would be to validate once the submit/save button gets pressed and then display the border around cells that need it. It might be worth checking in with RP on this.
> > I'd also suggest updating the text in the warning toast message to say something about required fields being empty or something to that effect.
> 
> Perhaps the red border is a little aggressive for drawing attention to blank cells but I'd argue some sort of feedback before having the user attempt to submit is better than only seeing errors once you attempt to submit, to get the feedback as early as possible.
> 
> @rajpersram thoughts?

Curious to hear @rajpersram thoughts. I'd say it's definitely an improvement over what we currently have but I also see Darren's point about a more typical pattern. I wonder if users might find the tooltips on every required input cell a little much, but maybe not because it sounds like they all just tab across inputs. Code looks good to me, curious what Raj has to say about UX

rajpersram:
I agree with only showing the red outline and note _after_ an action (of submitting) has occured, as opposed to always showing it as red until a value is entered.

conbrad:
Closing to save dev resources while we hackathon

PR #3947, opened 2024-09-17
Fix critical hours when O1A/O1B present

dgboss:
The fuel grid raster doesn't differentiate between O1A and O1B fuel types, it combines them as O1A/O1B. This joint fuel type doesn't exist in our FuelTypeEnum. Being unsure of the ramifications with cffdrs, after discussion with the PO we are treating O1A/O1B as O1B for the sake of calculating critical hours.

This PR also adds the ability to re-run critical hours processing by supplying a RunParameters.id.

### DIFF
--- a/api/app/auto_spatial_advisory/critical_hours.py
+++ b/api/app/auto_spatial_advisory/critical_hours.py
@@ -1,3 +1,4 @@
+import argparse
 import asyncio
 from collections import defaultdict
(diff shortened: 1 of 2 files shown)


conbrad (approved):
Nice catch, I totally didn't realize that the fuel layer didn't treat grass types separately.

brettedw (approved):
Nice find!

PR #4032, opened 2024-10-17
Publish base container to ghcr

compscidr:
I was chatting with @conbrad this morning, and he suggested this might be a good first issue.

The ultimate goal is to update https://github.com/bcgov/wps/blob/main/Dockerfile.vscode so that it's more similar to the production container. The first step of this is having them use the same base image, however, currently the base image is in a private repository.

The solution is to publish the base image into the github container repository (ghcr), which this PR does using a workflow that runs when changes to the base image Dockerfile are pushed to main.

~There is one thing that must be done before merging this PR for it to work, A personal access token named GH_TOKEN must be added to the repository secrets. This token must have `read:packages` `write:packages` and `delete:packages` permissions.~ There is actually a built-in token that workflows have which has permission already. The PR has been updated to use this token.

After the workflow runs for the first time, someone will manually need to click on the package in GH, go to its settings and set it to public. This will let people use `docker pull` on the image without needing to login to ghcr. 
image

You can see an example of the published base image on my fork:
image

And you can docker pull it: `docker pull ghcr.io/compscidr/wps/wps-api-base`

Once this is merged, it will be pullable with `docker pull ghcr.io/bcgov/wps/wps-api-base`

One other caveat. The base image itself has a base image which is also in a private repo, so the base for this image is `ubuntu:24.04`. The actual base for that image can be used if the artifactory credentials for `artifacts.developer.gov.bc.ca` are added to the workflow so that the real base can be used to make this completely accurate.

Once this is merged, I can follow-up with another PR that updates the base image for the dev container.


.github/workflows/publish_docker_base.yml
```diff
@@ -0,0 +1,31 @@
...
+  push:
+    branches:
+      - main
+    paths:
+      - 'openshift/wps-api-base/docker/Dockerfile'
+      - '.github/workflows/publish_docker_base.yml'
+jobs:
+  build-and-publish:
+    name: Build and Publish Docker Image
+    runs-on: ubuntu-latest
```
conbrad, line 13: For consistency
```suggestion
    runs-on: ubuntu-24.04
```

.github/workflows/publish_docker_base.yml
```diff
@@ -0,0 +1,31 @@
...
+        with:
+          registry: ghcr.io
+          username: ${{github.actor}}
+          password: ${{secrets.GH_TOKEN}}
+      - name: Build and Publish Base Image
+        # note we can remove the build-arg if we are able to use credentials for artifacts.developer.gov.bc.ca
+        # so we can directly consume that as a base image instead of ubuntu:24.04.
+        run: |
+          cd openshift/wps-api-base/docker
+          docker build --build-arg DOCKER_IMAGE=ubuntu:24.04 . --tag ghcr.io/bcgov/wps/wps-api-base:latest
```
conbrad, line 30: Can we create a tag this with the current date in `mm-dd-yyyy` format, then tag it as latest so we have dated versions?

conbrad (commented):
Thanks for contributing! Just a couple comments on my side, will check with the team what they think.

conbrad:
> One other caveat. The base image itself has a base image which is also in a private repo, so the base for this image is ubuntu:24.04. The actual base for that image can be used if the artifactory credentials for artifacts.developer.gov.bc.ca are added to the workflow so that the real base can be used to make this completely accurate.

So the reason we do this is that the private repo is a pull through cache -- we can actually just use `ghcr.io/osgeo/gdal:ubuntu-small-3.9.2` since we can grab this image through the pull-through cache as well.

compscidr:
> > One other caveat. The base image itself has a base image which is also in a private repo, so the base for this image is ubuntu:24.04. The actual base for that image can be used if the artifactory credentials for artifacts.developer.gov.bc.ca are added to the workflow so that the real base can be used to make this completely accurate.
> 
> So the reason we do this is that the private repo is a pull through cache -- we can actually just use `ghcr.io/osgeo/gdal:ubuntu-small-3.9.2` since we can grab this image through the pull-through cache as well.

That would be cool, I don't think I know enough about how that works to set that up / test it though.

openshift/wps-api-base/docker/Dockerfile
```diff
@@ -10,6 +10,9 @@ ARG DOCKER_IMAGE=artifacts.developer.gov.bc.ca/github-docker-remote/osgeo/gdal:u
 # docker build --build-arg DOCKER_IMAGE=ubuntu:24.04 . --tag=wps-api-base:ubuntu.24.04-latest
 FROM ${DOCKER_IMAGE}
```
conbrad, line 11: What I mean in this [comment](https://github.com/bcgov/wps/pull/4032#issuecomment-2420384472) was just:
```suggestion
FROM ghcr.io/osgeo/gdal:ubuntu-small-3.9.2
```

compscidr: oh derp, I get what you mean now, I thought there was other custom stuff going on - didnt realize it was just using another ghcr repo!

compscidr: I can probably update the workflow too since it won't have to pass in the base image name any longer

compscidr: Do you want the open shift `build.yaml` to be update to reflect this change too?

conbrad: Yeah that'd be great

compscidr: maybe its better to leave the base image to be configurable, but just update the base image on the workflow to be: `ghcr.io/osgeo/gdal:ubuntu-small-3.9.2`

compscidr: ie) this line: https://github.com/bcgov/wps/pull/4032/files#diff-581854468d04e2aa749d259292a93a295bda3c15f20a399d85fa8e80b211f163R30

compscidr: Just because I'm not sure if anything else you guys have is using the `DOCKER_IMAGE` arg or something

conbrad: Either way works for me.

conbrad (approved):

conbrad:
All non-deployment checks pass, looks like forks can't access repo secrets in workflows. Happy to merge this if @dgboss and @brettedw approve.

dgboss:
Looks good to me!

dgboss (approved):

PR #4214, opened 2025-01-21
ASA Mobile: Capacitor Skeleton

conbrad:
- Adds a capacitor app using react/vite that renders a basic `FBAMap` with pmtiles layers as well as a header and set of tabs.
- Application code is mostly copied from web


dgboss (approved):
Love it!

brettedw (approved):
👌👌👌👌

PR #4343, opened 2025-03-12
Setup and stats processing

brettedw:
- Update setup scripts/instructions
- Make processing stats locally easier

# Test Links:
[Landing Page](https://wps-pr-4343-e1e498-dev.apps.silver.devops.gov.bc.ca/)
[MoreCast](https://wps-pr-4343-e1e498-dev.apps.silver.devops.gov.bc.ca/morecast)
[Percentile Calculator](https://wps-pr-4343-e1e498-dev.apps.silver.devops.gov.bc.ca/percentile-calculator)
[C-Haines](https://wps-pr-4343-e1e498-dev.apps.silver.devops.gov.bc.ca/c-haines)
[FireBat](https://wps-pr-4343-e1e498-dev.apps.silver.devops.gov.bc.ca/fire-behaviour-calculator)
[FireBat bookmark](https://wps-pr-4343-e1e498-dev.apps.silver.devops.gov.bc.ca/fire-behaviour-calculator?s=266&f=c5&c=NaN&w=20,s=286&f=c7&c=NaN&w=16,s=1055&f=c7&c=NaN&w=NaN,s=305&f=c7&c=NaN&w=NaN,s=344&f=c5&c=NaN&w=NaN,s=346&f=c7&c=NaN&w=NaN,s=328&f=c7&c=NaN&w=NaN,s=1399&f=c7&c=NaN&w=NaN,s=334&f=c7&c=NaN&w=NaN,s=1082&f=c3&c=NaN&w=NaN,s=388&f=c7&c=NaN&w=NaN,s=309&f=c7&c=NaN&w=16,s=306&f=c7&c=NaN&w=NaN,s=1029&f=c7&c=NaN&w=NaN,s=298&f=c7&c=NaN&w=NaN,s=836&f=c7&c=NaN&w=NaN,s=9999&f=c7&c=NaN&w=NaN)
[Auto Spatial Advisory (ASA)](https://wps-pr-4343-e1e498-dev.apps.silver.devops.gov.bc.ca/auto-spatial-advisory)
[HFI Calculator](https://wps-pr-4343-e1e498-dev.apps.silver.devops.gov.bc.ca/hfi-calculator)
[SFMS Insights](https://wps-pr-4343-e1e498-dev.apps.silver.devops.gov.bc.ca/insights)

### DIFF
--- a/api/app/auto_spatial_advisory/local/process_stats_local.py
+++ b/api/app/auto_spatial_advisory/local/process_stats_local.py
@@ -35,7 +35,6 @@ END $$;
 import sys
 import asyncio
 from datetime import date, timedelta
-from sqlalchemy import select
 from wps_shared.db.crud.auto_spatial_advisory import get_most_recent_run_parameters
 from wps_shared.db.database import get_async_read_session_scope
 from wps_shared.logging import configure_logging
--- a/api/app/auto_spatial_advisory/local/process_stats_local.py
+++ b/api/app/auto_spatial_advisory/local/process_stats_local.py
@@ -7,18 +7,76 @@ Expects
    'OBJECT_STORE_SERVER'
 
 to be set
+----------------------
+
+If you're trying to reprocess stats that have already been run, you can ensure all stats get
+re-calculated by deleting the original records from the database (excluding run_parameters)
+
+DO $$
+DECLARE run_ids INT[];  -- Declare an array variable to store the list of run_id's
+BEGIN
+    -- Populate the array dynamically from a query using a date range
+    SELECT array_agg(id) INTO run_ids
+    FROM run_parameters
+    WHERE for_date BETWEEN '2024-10-23' AND '2024-10-25';
+
+    -- Delete from all tables using the run_ids array
+    DELETE FROM advisory_fuel_stats WHERE run_parameters = ANY(run_ids);
+    DELETE FROM advisory_tpi_stats WHERE run_parameters = ANY(run_ids);
+    DELETE FROM advisory_elevation_stats WHERE run_parameters = ANY(run_ids);
+    DELETE FROM high_hfi_area WHERE run_parameters = ANY(run_ids);
+    DELETE FROM advisory_hfi_wind_speed WHERE run_parameters = ANY(run_ids);
+    DELETE FROM advisory_hfi_percent_conifer WHERE run_parameters = ANY(run_ids);
+    DELETE FROM critical_hours WHERE run_parameters = ANY(run_ids);
+
+END $$;
 """
 
 import sys
 import asyncio
-from datetime import date, datetime
+import os
+from datetime import date, datetime, timedelta, timezone, time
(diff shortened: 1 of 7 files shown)


conbrad (approved):

api/app/auto_spatial_advisory/local/process_stats_local.py
```diff
@@ -7,18 +7,78 @@
...
+            print("advisory_shape_fuels must be populated")
+            sys.exit(os.EX_SOFTWARE)
+
+        for for_date in for_dates:
+            # try to reprocess the run that ASA will look for
+            run_param = await get_most_recent_run_parameters(session, run_type, for_date)
+            if run_param:
+                run_datetime = run_param[0].run_datetime
+            elif run_type == RunType.ACTUAL:
+                # Use a default run_datetime when no run parameters are found for an actual. The run_datetime
```
dgboss, line 63: In what situation would run parameters not be found for an actual?

api/poetry_setup.sh
```diff
@@ -1,8 +0,0 @@
-#!/usr/bin/env bash
-
-poetry config virtualenvs.in-project false --local
-poetry env use 3.12.3
```
dgboss, line 4: `poetry env use 3.12.3` is missing from the new setup script. Is it not necessary to tell poetry which version of Python to use?

brettedw: It's not absolutely necessary since the version is specified in the `pyproject.toml`. That being said I just noticed the version is specified as `>=3.12.3`. So if you had 3.13.0 installed, it would use that. The highest version I have installed is 3.12.3
Good thought, I can add it back since we probably want to ensure we're using 3.12.3

api/app/auto_spatial_advisory/local/process_stats_local.py
brettedw: The main scenario I'm thinking of is when your local db hasn't been updated in some time, it won't have run parameters going forward. Say you restored your local on July 1 and it's currently Aug 1. You could still process data up to Aug 1st if we create some run_parameters that don't exist in local.

Your comment made me think some more about it though. If we want this functionality, the same logic is true for almost every forecast day. The most recent forecast data for all days except the future (after Aug 1) also have `run_datetime`'s on the same `for_date`.

This is getting a bit dangerous though....if someone runs this while their `.env` is set up to look in our prod s3 bucket for data, it might overwrite hfi data in s3. Maybe I should leave this logic out completely....thoughts?

conbrad: Yeah I would favor exploding with an error here if it's an unrecoverable situation and it could corrupt data.

dgboss: Thanks for the explanation. I agree with Conor about playing it safe.

dgboss (approved):

PR #4417, opened 2025-04-07
SFMS Calculation datatypes/max value

brettedw:
- Removes `datatype` arg from `WPSDataset`. Using gdal's `band.DataType` to use same datatype as existing dataset
- Adds logging for weather parameters to check for invalid values as defined by CFFDRS. This is useful because when doing vectorized calculations we don't get much in terms of error messaging
- Adds `max_value` to `warp_to_match`, this ensures we can set a max value when resampling is being done so that we don't get out of range values (ie RH > 100)
# Test Links:
[Landing Page](https://wps-pr-4417-e1e498-dev.apps.silver.devops.gov.bc.ca/)
[MoreCast](https://wps-pr-4417-e1e498-dev.apps.silver.devops.gov.bc.ca/morecast)
[Percentile Calculator](https://wps-pr-4417-e1e498-dev.apps.silver.devops.gov.bc.ca/percentile-calculator)
[C-Haines](https://wps-pr-4417-e1e498-dev.apps.silver.devops.gov.bc.ca/c-haines)
[FireBat](https://wps-pr-4417-e1e498-dev.apps.silver.devops.gov.bc.ca/fire-behaviour-calculator)
[FireBat bookmark](https://wps-pr-4417-e1e498-dev.apps.silver.devops.gov.bc.ca/fire-behaviour-calculator?s=266&f=c5&c=NaN&w=20,s=286&f=c7&c=NaN&w=16,s=1055&f=c7&c=NaN&w=NaN,s=305&f=c7&c=NaN&w=NaN,s=344&f=c5&c=NaN&w=NaN,s=346&f=c7&c=NaN&w=NaN,s=328&f=c7&c=NaN&w=NaN,s=1399&f=c7&c=NaN&w=NaN,s=334&f=c7&c=NaN&w=NaN,s=1082&f=c3&c=NaN&w=NaN,s=388&f=c7&c=NaN&w=NaN,s=309&f=c7&c=NaN&w=16,s=306&f=c7&c=NaN&w=NaN,s=1029&f=c7&c=NaN&w=NaN,s=298&f=c7&c=NaN&w=NaN,s=836&f=c7&c=NaN&w=NaN,s=9999&f=c7&c=NaN&w=NaN)
[Auto Spatial Advisory (ASA)](https://wps-pr-4417-e1e498-dev.apps.silver.devops.gov.bc.ca/auto-spatial-advisory)
[HFI Calculator](https://wps-pr-4417-e1e498-dev.apps.silver.devops.gov.bc.ca/hfi-calculator)
[SFMS Insights](https://wps-pr-4417-e1e498-dev.apps.silver.devops.gov.bc.ca/insights)


conbrad (approved):

dgboss (approved):

api/app/sfms/fwi_processor.py
```diff
@@ -15,6 +15,10 @@ def calculate_dc(dc_ds: WPSDataset, temp_ds: WPSDataset, rh_ds: WPSDataset, prec
     rh_array, _ = rh_ds.replace_nodata_with(0)
     precip_array, _ = precip_ds.replace_nodata_with(0)
 
+    # Due to warping of the rh dataset, rh values can exceed 100 which breaks the calculation.
```
conbrad, line 18: I wonder if we should do this correction in `warp_to_match` to keep it in one place. Perhaps it's an optional parameter only applied to rh.

api/app/sfms/fwi_processor.py
```diff
@@ -65,9 +73,9 @@ def calculate_ffmc(previous_ffmc_ds: WPSDataset, temp_ds: WPSDataset, rh_ds: WPS
     precip_array, _ = precip_ds.replace_nodata_with(0)
     wind_speed_array, _ = wind_speed_ds.replace_nodata_with(0)
 
-    # Due to warping of the rh dataset, rh values can exceed 100 which breaks the ffmc calculation.
-    # Set rh values greater than 100 to the max allowable which is 100.
-    rh_array[rh_array > 100] = 100
+    if (rh_array > 100).any():
```
dgboss, line 76: You could pull this out into a function.

wps_shared/wps_shared/geospatial/wps_dataset.py
```diff
@@ -137,7 +137,13 @@ def __mul__(self, other):
 
         return WPSDataset(ds_path=None, ds=out_ds)
 
-    def warp_to_match(self, other: "WPSDataset", output_path: str, resample_method: GDALResamplingMethod = GDALResamplingMethod.NEAREST_NEIGHBOUR):
+    def warp_to_match(
+        self,
+        other: "WPSDataset",
+        output_path: str,
+        resample_method: GDALResamplingMethod = GDALResamplingMethod.NEAREST_NEIGHBOUR,
+        max_value: float | None = None,
```
dgboss, line 145: The `WPSDataset` allows for non-float datatypes (the default is `gdal.GDT_Byte`). It might be safer and keep things more flexible if the application of a max value gets factored out into an explicit function; probably a function that returns a new `WPSDataset` with the updated data.

conbrad (approved):

wps_shared/wps_shared/geospatial/wps_dataset.py
dgboss: Ha, I just read Conor's comment about creating that logic inside this function! If we're warping a dataset chances are we're dealing with floats so I'm likely being over cautious. Feel free to ignore my comment.

dgboss (approved):

wps_shared/wps_shared/geospatial/wps_dataset.py
brettedw: I could type it as `float | int | None`. I think it should work with whichever number type is provided, but maybe I'm not thinking that through

conbrad: You can provide the max value to the `init` function alongside the `datatype`. Better yet, compose the `datatype` and `maxvalue` together in it's own class to make sure they correspond and pass that to `WPSDataset`'s `init`.

brettedw: The datatype in the `init` is kind of misleading at the moment in my mind. For example when we are initializing a `WPSDataset`, if we don't explicitly know the datatype of the gdal geotiff ahead of time, it will always be set to `GDT_Byte`. So when we're loading in a temperature geotiff, the `WPSDataset` `datatype` is `GDAL_Byte`, but the gdal dataset `datatype` is `GDT_Float64`
Maybe it's worthwhile to talk about what the intention of providing the `datatype` is?

conbrad: The original intent for `datatype` was to create a memory driver in the `__mul__` operation. The default `GDAL_Byte` was chosen to minimize memory overhead unless the caller explicitly knows they need a larger data width. 

I didn't realize you can grab the data type from a dataset raster band with `band.DataType`. We should use that if we can instead.

dgboss (approved):
Looks great!

PR #4493, opened 2025-05-13
Dashboard summary datagrid

dgboss:
A quick and dirty implementation to have something to show for sprint review!
 - adds a `DataGrid` to the dashboard to show FireWatch status
 - uses master-detail feature to show daily status for a FireWatch
- endpoint for dashboard data

# Test Links:
[Landing Page](https://wps-pr-4493-e1e498-dev.apps.silver.devops.gov.bc.ca/)
[MoreCast](https://wps-pr-4493-e1e498-dev.apps.silver.devops.gov.bc.ca/morecast)
[Percentile Calculator](https://wps-pr-4493-e1e498-dev.apps.silver.devops.gov.bc.ca/percentile-calculator)
[C-Haines](https://wps-pr-4493-e1e498-dev.apps.silver.devops.gov.bc.ca/c-haines)
[FireCalc](https://wps-pr-4493-e1e498-dev.apps.silver.devops.gov.bc.ca/fire-behaviour-calculator)
[FireCalc bookmark](https://wps-pr-4493-e1e498-dev.apps.silver.devops.gov.bc.ca/fire-behaviour-calculator?s=266&f=c5&c=NaN&w=20,s=286&f=c7&c=NaN&w=16,s=1055&f=c7&c=NaN&w=NaN,s=305&f=c7&c=NaN&w=NaN,s=344&f=c5&c=NaN&w=NaN,s=346&f=c7&c=NaN&w=NaN,s=328&f=c7&c=NaN&w=NaN,s=1399&f=c7&c=NaN&w=NaN,s=334&f=c7&c=NaN&w=NaN,s=1082&f=c3&c=NaN&w=NaN,s=388&f=c7&c=NaN&w=NaN,s=309&f=c7&c=NaN&w=16,s=306&f=c7&c=NaN&w=NaN,s=1029&f=c7&c=NaN&w=NaN,s=298&f=c7&c=NaN&w=NaN,s=836&f=c7&c=NaN&w=NaN,s=9999&f=c7&c=NaN&w=NaN)
[Auto Spatial Advisory (ASA)](https://wps-pr-4493-e1e498-dev.apps.silver.devops.gov.bc.ca/auto-spatial-advisory)
[HFI Calculator](https://wps-pr-4493-e1e498-dev.apps.silver.devops.gov.bc.ca/hfi-calculator)
[SFMS Insights](https://wps-pr-4493-e1e498-dev.apps.silver.devops.gov.bc.ca/insights)
[Fire Watch](https://wps-pr-4493-e1e498-dev.apps.silver.devops.gov.bc.ca/fire-watch)


api/app/routers/fire_watch.py
```diff
@@ -156,6 +181,43 @@ def create_fire_watch_output(db_fire_watch: DBFireWatch, fire_centre: FireWatchF
...
+    # Build a dictionary of BurnForecastOutputs keyed by fire watch id for easy lookup
+    fire_watch_weather_dict = defaultdict(list)
+    for item, prescription in fire_watch_weather:
+        fire_watch_weather_dict[item.fire_watch_id].append(create_burn_forecast_output(item, prescription))
+    # Build a dictionary of FireWatchOutputBurnForecasts keyed by fire watch id for easy lookup
+    fire_watch_burn_forecasts: Dict[int, FireWatchOutputBurnForecast] = {}
+    for fire_watch, fire_centre in fire_watches:
+        fire_watch_output = create_fire_watch_output(fire_watch, fire_centre, stations)
+        burn_forecast_outputs = fire_watch_weather_dict.get(fire_watch.id, [])
+        _ = burn_forecast_outputs.sort(key=lambda x: x.date)
```
conbrad, line 215:
```suggestion
       burn_forecast_outputs.sort(key=lambda x: x.date)
```

web/src/features/fireWatch/components/FireWatchDashboard.tsx
```diff
@@ -1,18 +1,226 @@
...
+      }
+      if (burnForecast.inPrescription === PrescriptionEnum.ALL) {
+        inPrescription = PrescriptionEnum.ALL
+        break
+      }
+    }
+    return inPrescription
+  }
+
+  const shapeRows = (values: FireWatchBurnForecast[]) => {
```
conbrad, line 69: This can be put in the `selectBurnForecasts` if it always has to shape rows this way.

web/src/features/fireWatch/fireWatchApi.ts
```diff
@@ -197,11 +129,19 @@ export const getFireCentres = async (): Promise<FireWatchFireCentresResponse> =>
   return data
 }
 
+export async function getBurnForecasts(): Promise<FireWatchBurnForecast[]> {
+  const url = '/fire-watch/burn-forecasts'
+  const { data } = await axios.get(url)
+  console.log(data)
```
conbrad, line 135:
```suggestion
```

wps_shared/wps_shared/schemas/fire_watch.py
```diff
@@ -82,3 +81,28 @@ class FireWatchListResponse(BaseModel):
...
+    in_prescription: str
+
+
+class FireWatchOutputBurnForecast(BaseModel):
+    fire_watch: FireWatchOutput
+    burn_forecasts: List[BurnForecastOutput]
+
+
+class FireWatchBurnForecastsResponse(BaseModel):
+    fire_watch_burn_forecasts: Dict[int, FireWatchOutputBurnForecast]
```
conbrad, line 108: What does the key in the dict represent here? Is it the unique `id` of a `FireWatch`? Is it possible to create a `FireWatch` schema with the `id` and a list of associated `FireWatchOutputBurnForecast` so marshalling on the frontend is simpler?

web/src/features/fireWatch/components/FireWatchDashboard.tsx
dgboss: Good call! Done in https://github.com/bcgov/wps/pull/4493/commits/1aa13ef419d718b88e38ac9e72ae157e476cb1ee

wps_shared/wps_shared/schemas/fire_watch.py
dgboss: Good idea! I'm on it.

web/src/features/fireWatch/components/FireWatchDashboard.tsx
dgboss: We can get away with it because we aren't doing any editing in the datagrid. If we eventually allow editing we may have to switch back to having the rows in local state.

wps_shared/wps_shared/schemas/fire_watch.py
dgboss: Simplified to return a list in https://github.com/bcgov/wps/pull/4493/commits/77c52106045d21ae9eda175ad9ca87ffad57eea4

brettedw (approved):
Looks good, nice job!

web/src/features/fireWatch/components/fireWatchDashboard.test.tsx
```diff
@@ -0,0 +1,236 @@
...
+      </Provider>
+    )
+    expect(dispatchSpy).toHaveBeenCalled()
+  })
+
+  it('should render', async () => {
+    const testStore = buildTestStore({
+      ...initialState
+    })
+    await act(async () =>
```
conbrad, line 59: Is `act` an alternative to `waitFor`?

brettedw:
Still looks good to me!
Only sharing because I thought it was handy when I found it, MUI has `darken` and `lighten` functions for colours. A good example of it being used for hovering rows in the first example here: https://mui.com/x/react-data-grid/style/#system-StylingRowsGrid.tsx

web/src/features/fireWatch/fireWatchApi.ts
```diff
@@ -154,7 +61,7 @@ export interface FireWatchInput {
 export interface FireWatchOutput extends FireWatchInput {
   id: number
   create_timestamp: number
```
conbrad, line 63: We use iso strings for over the wire timestamps usually, any reason why we use a number over string?

dgboss:
> Still looks good to me! Only sharing because I thought it was handy when I found it, MUI has `darken` and `lighten` functions for colours. A good example of it being used for hovering rows in the first example here: https://mui.com/x/react-data-grid/style/#system-StylingRowsGrid.tsx

Very cool! I'll check that out before merging.

web/src/features/fireWatch/components/fireWatchDashboard.test.tsx
dgboss: They are somewhat similar but there seems to be some nuance. Typically you wouldn't even need to use `act` in conjunction with the `render` call, but I believe it is necessary in this case when waiting on a `dispatch` call that eventually updates state. I don't think `act` will return until those state updates are complete. In comparison, `waitFor` implements a poling mechanism and repeatedly checks to see if the `expect` returns `true` (in this use case) and times out after 1000ms.

web/src/features/fireWatch/fireWatchApi.ts
dgboss: Force of habit from a past life, epoch times are hard to misinterpret. I think I introduced epoch time into our api somewhere else. I see what you mean about ISO strings being the standard for our API though, I'll switch things over.

conbrad: Fair, happy to revisit the precedent if it makes sense though.

web/src/features/fireWatch/components/fireWatchDashboard.test.tsx
conbrad: Interesting, where did you find those details? Looking at https://testing-library.com/docs/react-testing-library/api/#act, which circularly references: https://react.dev/reference/react/act, it looks like react-testing library wraps the behaviour of act in it's helper functions already. Does the test fail with `waitFor`?

web/src/features/fireWatch/fireWatchApi.ts
dgboss: I made the change in https://github.com/bcgov/wps/pull/4493/commits/06cdb51a472c71884e1145918861c3f7efef1232.
I can't test in dev at the moment as WF1 test seems to be down.

web/src/features/fireWatch/components/fireWatchDashboard.test.tsx
dgboss: Tests seem to pass with `waitFor`. I assume this is due to the delay it introduces versus actually waiting for rendering to happen though. `waitFor` waits for the callback function's promise to resolve. As `render` itself doesn't actually return a promise, the `async () => render()` likely returns after the first 50ms second delay introduced by `waitFor`. I believe `waitFor` actually comes from `@testing-library/dom-testing-library` and does not include a call to `act`. I'll admit I don't fully understand all the nuances. Error messages from either `react-testing-library` or `dom-testing-library` told me to use `act`!

conbrad: Interesting, yeah let's stick with `act` if there's no async promise that's being awaited.

conbrad (approved):
Looks great!

PR #4566, opened 2025-06-11
Add arrows to ASA date picker

dgboss:
- introduces a new date picker with back/forward arrows
- enables/disables the calendar and forward/back buttons according to available sfms run dates
- new endpoint to retrieve sfms min/max run dates for a given year and run type

### DIFF
--- a/web/src/features/fba/components/ASADatePicker.tsx
+++ b/web/src/features/fba/components/ASADatePicker.tsx
@@ -1,0 +1,116 @@
+import { PlayArrow } from '@mui/icons-material'
+import { IconButton, TextField } from '@mui/material'
+import {
+  BaseSingleInputFieldProps,
+  CalendarIcon,
+  DatePicker,
+  DatePickerProps,
+  DateValidationError,
+  FieldSection,
+  LocalizationProvider,
+  UseDateFieldProps,
+  usePickersContext,
+  useSplitFieldProps
+} from '@mui/x-date-pickers'
+import { AdapterLuxon } from '@mui/x-date-pickers/AdapterLuxon'
+import { isNil, isNull } from 'lodash'
+import { DateTime } from 'luxon'
+import React from 'react'
+
+interface CustomDateTextFieldProps
+  extends UseDateFieldProps<DateTime, false>,
+    BaseSingleInputFieldProps<DateTime | null, DateTime, FieldSection, false, DateValidationError> {
+  date: DateTime
+  updateDate: React.Dispatch<React.SetStateAction<DateTime>>
+  minDate: DateTime
+  maxDate: DateTime
+}
+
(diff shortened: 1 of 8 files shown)


web/src/features/fba/components/ASADatePicker.tsx
```diff
@@ -0,0 +1,116 @@
...
+        minDate={minDate}
+        maxDate={maxDate}
+        value={date}
+        onAccept={(newValue: DateTime | null) => {
+          if (!isNull(newValue) && newValue.isValid) {
+            updateDate(newValue)
+          }
+        }}
+        slots={{ ...other.slots, field: CustomDateTextField }}
+        slotProps={{ ...other.slotProps, field: { date, updateDate, minDate, maxDate } as any }}
```
dgboss, line 111: DatePicker helper functions and typings are a mess in @mui/x-date-pickers@v7. The @mui examples use `as any` rather than trying to sort it out.

brettedw:
> > This is looking great, nice job! Only thing I'm seeing is locally where I have data I'm not seeing dates before the minimum or the default fire season values become disabled `{"sfms_bounds":{"minimum":"2025-04-23","maximum":"2025-05-24"}}`
> 
> There seemed to be a collision/race condition with my custom prop names and pre-existing DatePicker prop names (minDate, maxDate). Should be smoothed out now.

I think that commit addressed the arrows, but the calendar view isn't showing dates as disabled for me

brettedw:
> > > > This is looking great, nice job! Only thing I'm seeing is locally where I have data I'm not seeing dates before the minimum or the default fire season values become disabled `{"sfms_bounds":{"minimum":"2025-04-23","maximum":"2025-05-24"}}`
> > > 
> > > 
> > > There seemed to be a collision/race condition with my custom prop names and pre-existing DatePicker prop names (minDate, maxDate). Should be smoothed out now.
> > 
> > 
> > I think that commit addressed the arrows, but the calendar view isn't showing dates as disabled for me
> 
> Yeah, there is a gotcha with the calendar view. If we specify the minDate and maxDate for the calendar view, the user loses the ability to change the year. It might be possible to figure something out but the enable/disable specified in the acceptance criteria only relate to the arrows. We could get current year min/max for the arrows and overall min/max for the calendar?

Ahh I see. I'd say this is good, for sure it addresses the acceptance criteria. I like it

Thinking maybe `shouldDisableDate` could be used if we want to do dates for every year? Hopefully that's not too costly

brettedw (approved):
Looks great, nice work! Just the one failing test at the moment

dgboss:
> Thinking maybe `shouldDisableDate` could be used if we want to do dates for every year? Hopefully that's not too costly

Yes, I think we could leverage a combination of shouldDisableDate/Month/Year to restrict selection on the calendar view.

PR #4598, opened 2025-06-16
Recalculate DMC, DC, BUI when precip changes

conbrad:
If precip changes, so must DMC, DC and BUI.


wps_shared/wps_shared/schemas/fba_calc.py
```diff
@@ -82,6 +82,8 @@ class AdjustedFWIResult(BaseModel):
     ffmc: float
     bui: float
     isi: float
+    dmc: Optional[float] = None
```
dgboss, line 85: Do these need to be optional?

dgboss (approved):

PR #4821, opened 2025-08-26
Add push notification support

conbrad:
Implement push notification setup for the app side. Requests permissions from the user to allow push notifications and registers the app and device with the notification service (currently just iOS and APNs https://en.wikipedia.org/wiki/Apple_Push_Notification_service).

### DIFF
--- a/mobile/asa-go/ios/App/App/AppDelegate.swift
+++ b/mobile/asa-go/ios/App/App/AppDelegate.swift
@@ -11,6 +11,34 @@ class AppDelegate: UIResponder, UIApplicationDelegate, KeycloakAppDelegate {
 
     func application(
         _ application: UIApplication,
+        didRegisterForRemoteNotificationsWithDeviceToken deviceToken: Data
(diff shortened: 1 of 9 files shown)


mobile/asa-go/ios/App/App/AppDelegate.swift
```diff
@@ -9,6 +9,34 @@ class AppDelegate: UIResponder, UIApplicationDelegate, KeycloakAppDelegate {
     var window: UIWindow?
     var currentAuthorizationFlow: OIDExternalUserAgentSession?
 
+    func application(
+        _ application: UIApplication,
+        didRegisterForRemoteNotificationsWithDeviceToken deviceToken: Data
+    ) {
+
+        let tokenString = deviceToken.map { String(format: "%02.2hhx", $0) }.joined()
+        print("Device Token: \(tokenString)")
```
dgboss, line 18: Should this print statement be removed?

PR #4969, opened 2025-12-17
Cache bust with workbox service worker

### LINKED ISSUE #4551: Dev: Investigate why apps become unavailable during deployments
**Actual Behaviour**

For a short period of time during deployments, our apps become unavailable - 404 

**Steps To Reproduce**

1. ...
2. 
3. 

**Screenshots**

**Expected Behaviour**

https://bcps-wps.sentry.io/issues/7117228031/events/603c39a4de04433aae2912a98ec7d579/

conbrad:
Implement Workbox service worker with automatic cache management to eliminate "Failed to fetch dynamically imported module" errors that occur when users have stale cached pages after deployments.

  Problem

  When deploying new versions of the frontend:
  1. Users load the app and receive index.html with references to JS chunks like MoreCast2Page-Bg2zRDR0.js
  2. A new deployment occurs, generating new chunk hashes (e.g., MoreCast2Page-XYZ123.js)
  3. Users with the old index.html attempt to dynamically import the old chunk name
  4. Server returns 404 because the old chunk no longer exists
  5. Application crashes with: TypeError: Failed to fetch dynamically imported module

  The previous vite:preloadError handler only catches Vite-specific preload errors and doesn't prevent the underlying issue of stale asset references.

  Solution

  Implemented a Workbox-based service worker that proactively manages asset caching and version updates:

  How it works now:

  The service worker will automatically update in the background when a new version is deployed ([autoUpdate](https://vite-pwa-org.netlify.app/guide/auto-update)). Users won't see any prompts or notifications. The updated assets will be used on their next natural page navigation or when they refresh the page on their own. The workbox configuration with clientsClaim: true and cleanupOutdatedCaches: true ensures old caches are cleaned up automatically.

  The Update Flow:

  1. New version deployed - You deploy new code to your server with updated JavaScript/CSS files
  2. Service worker detects the change - The browser periodically checks for service worker updates (or when the user navigates). It notices the service worker file has changed
  3. New service worker installs - The new service worker downloads and installs in the background, precaching the new assets
  4. Automatic activation - With autoUpdate, the new service worker immediately takes control without waiting for user action. This is combined with clientsClaim: true in your workbox config
  5. Next navigation applies updates - Here's the key part:
    - The current page continues using the old cached assets that were already loaded
    - When the user navigates to a different route/page or refreshes the page themselves, the new service worker serves the new cached assets
    - This happens seamlessly without any interruption

  Example scenario:
  - User is on /dashboard viewing old version
  - You deploy new version
  - Service worker updates in background (user doesn't notice anything)
  - User clicks link to go to /reports → new version loads
  - User eventually returns to /dashboard → new version loads

  The user never sees a prompt, never experiences a forced reload, and gets the update naturally through their normal browsing behavior. The old version doesn't "stick around" for the current page, but new navigations immediately use the new code.

  Why This Prevents Errors:
  - All chunks are precached before the update prompt appears
  - Users never attempt to fetch non-existent chunks
  - The service worker serves assets from cache, not the network
  - Updates are atomic: either fully old version or fully new version, never mixed

  Technical Implementation

  Service Worker Strategy:
  - Precaching: All build assets are cached during service worker installation
  - Update Detection: Service worker checks for updates on navigation
  - Controlled Activation: New service worker doesn't activate until user refreshes
  - Client Claim: clientsClaim: true ensures immediate takeover after activation
  - Cache Cleanup: Old versions automatically purged

https://developer.chrome.com/docs/workbox/modules/workbox-precaching

Closes #4551 

# Test Links:
[Landing Page](https://wps-pr-4969-e1e498-dev.apps.silver.devops.gov.bc.ca/)
[MoreCast](https://wps-pr-4969-e1e498-dev.apps.silver.devops.gov.bc.ca/morecast)
[Percentile Calculator](https://wps-pr-4969-e1e498-dev.apps.silver.devops.gov.bc.ca/percentile-calculator)
[C-Haines](https://wps-pr-4969-e1e498-dev.apps.silver.devops.gov.bc.ca/c-haines)
[FireCalc](https://wps-pr-4969-e1e498-dev.apps.silver.devops.gov.bc.ca/fire-behaviour-calculator)
[FireCalc bookmark](https://wps-pr-4969-e1e498-dev.apps.silver.devops.gov.bc.ca/fire-behaviour-calculator?s=266&f=c5&c=NaN&w=20,s=286&f=c7&c=NaN&w=16,s=1055&f=c7&c=NaN&w=NaN,s=305&f=c7&c=NaN&w=NaN,s=344&f=c5&c=NaN&w=NaN,s=346&f=c7&c=NaN&w=NaN,s=328&f=c7&c=NaN&w=NaN,s=1399&f=c7&c=NaN&w=NaN,s=334&f=c7&c=NaN&w=NaN,s=1082&f=c3&c=NaN&w=NaN,s=388&f=c7&c=NaN&w=NaN,s=309&f=c7&c=NaN&w=16,s=306&f=c7&c=NaN&w=NaN,s=1029&f=c7&c=NaN&w=NaN,s=298&f=c7&c=NaN&w=NaN,s=836&f=c7&c=NaN&w=NaN,s=9999&f=c7&c=NaN&w=NaN)
[Auto Spatial Advisory (ASA)](https://wps-pr-4969-e1e498-dev.apps.silver.devops.gov.bc.ca/auto-spatial-advisory)
[HFI Calculator](https://wps-pr-4969-e1e498-dev.apps.silver.devops.gov.bc.ca/hfi-calculator)
[SFMS Insights](https://wps-pr-4969-e1e498-dev.apps.silver.devops.gov.bc.ca/insights)
[Fire Watch](https://wps-pr-4969-e1e498-dev.apps.silver.devops.gov.bc.ca/fire-watch)

### DIFF
--- a/backend/packages/wps-weather/README.md
+++ b/backend/packages/wps-weather/README.md
@@ -1,0 +1,78 @@
+# wps-weather
+
+Weather maps and meteorological data processing for Wildfire Predictive Services Unit.
+
+Contains utilities for fetching, processing, and visualizing weather data including maps, forecasts, and meteorological datasets.
+
+## Standalone installation with uv
+
+### Installing uv
+
+First, install uv if you don't have it:
+
+**macOS/Linux:**
+```bash
+curl -LsSf https://astral.sh/uv/install.sh | sh
+```
+
+**Homebrew:**
+```bash
+HOMEBREW_NO_AUTO_UPDATE=1 brew install uv
+```
+
+**Windows:**
+```powershell
+powershell -c "irm https://astral.sh/uv/install.ps1 | iex"
+```
+
+**Winget:**
+```powershell
+winget install --id=astral-sh.uv -e
+```
+
+**Alternative (using pip):**
+```bash
+pip install uv
+```
+
+### Installing the package
+
+To install in a separate virtual environment:
+
+```bash
+# Navigate to the package directory
+cd packages/wps-weather
+
+# Create and activate a virtual environment
+uv venv
+source .venv/bin/activate  # On macOS/Linux
+# or: .venv\Scripts\activate  # On Windows
+
+# Install the package
+uv pip install .
+```
+
+### Using standard pip
+
+```bash
+cd packages/wps-weather
+python -m venv .venv
+source .venv/bin/activate
+pip install -e .
+```
+
+## Dependencies
+
+This package includes:
+- **Data processing**: numpy, xarray, scipy, pandas
+- **Geospatial**: geopandas, shapely, pyproj, fiona, cartopy
+- **Visualization**: matplotlib
+- **Interactive development**: jupyter, notebook, jupyterlab
+
+## Development
+
+### Adding new Python files
+
+Add new Python modules to `src/wps_weather/`. They will be automatically available for import as `from wps_weather.your_module import ...`
+
+Optionally export them from `src/wps_weather/__init__.py` for easier access.
--- a/backend/packages/wps-weather/pyproject.toml
+++ b/backend/packages/wps-weather/pyproject.toml
@@ -1,0 +1,27 @@
+[project]
+name = "wps-weather"
+version = "0.1.0"
+description = "Wildfire Predictive Services Unit Weather Maps"
+authors = []
+requires-python = ">=3.12.3,<4.0"
+dependencies = [
+    "numpy==2.3.*",
+    "xarray>=2025.3.1,<2026",
+    "matplotlib>=3.7.0,<4",
+    "cartopy>=0.22.0,<1",
+    "scipy>=1.15.2,<2",
+    "geopandas>=1.0.1,<2",
+    "shapely>=2.0.5,<3",
+    "pyproj>=3.5.0,<4",
+    "fiona>=1.9.0,<2",
+    "jupyter>=1.0.0,<2",
+    "notebook>=7.0.0,<8",
+    "jupyterlab>=4.0.0,<5",
+]
+
+[build-system]
+requires = ["hatchling"]
+build-backend = "hatchling.build"
+
+[tool.hatch.build.targets.wheel]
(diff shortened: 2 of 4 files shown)


dgboss:
I like the service worker approach but I'm not a fan of popping up a `Snackbar` and prompting the user to refresh when they don't need to in order to continue using the app they have open. Do we get the benefits of the service worker on a manual page reload?

conbrad:
> I like the service worker approach but I'm not a fan of popping up a `Snackbar` and prompting the user to refresh when they don't need to in order to continue using the app they have open. Do we get the benefits of the service worker on a manual page reload?

It's a pattern I've seen in apps recently. I think the benefit of this approach over reloading automatically is if the page is already loaded and they're filling out fields or something, they don't lose all of that without their intervention (pressing the button).

dgboss:
> > I like the service worker approach but I'm not a fan of popping up a `Snackbar` and prompting the user to refresh when they don't need to in order to continue using the app they have open. Do we get the benefits of the service worker on a manual page reload?
> 
> It's a pattern I've seen in apps recently. I think the benefit of this approach over reloading automatically is if the page is already loaded and they're filling out fields or something, they don't lose all of that without their intervention (pressing the button).
> 
> ~Edit: I guess the page doesn't necessarily reload if the service worker is pulling new javascript. It could change the page behaviour though. I'm not sure...~
> 
> Confirmed it's a window reload.

I definitely wouldn't advocate for automatic reload. I think the service worker will cache all the updated resources but they don't get loaded until the page refreshes. We deploy to prod pretty regularly and the majority of those deployments are specific to one app. Prompting every user of every app to refresh their page when they don't actually need to is something we should avoid in my opinion. To my knowledge, the only time a user encounters the error we've seen is when they are loading/re-loading one of our apps and I'm hopeful that the resource caching done by the service worker will take care of that.

Issue #5184, opened 2026-03-05
ASA Go: Settings Menu


NicolasLivanos:
**Describe the task**

Create the Settings Menu based on https://www.figma.com/design/yGPwOBayMeThQaRpkkZIGx/AutoSpatialAdvisory?node-id=1348-5598&m=dev

**Acceptance Criteria**
- [ ] Disable notification menu when offline
- [ ] Settings button goes into the navbar (to replace weather's space for now)
- [ ] Clicking on settings brings up a page to show and provide ability to subscribe/unsub to a zone's notifications.
- [ ] Nested toggles with zones under each centre in a list.

**Additional context**
- Add any other context about the task here.
- Or here

PR #5212, opened 2026-03-12
Notifications /Settings tab

dgboss:
- Adds Settings button to bottom and side nav and a settings panel
- Settings panel only contains notification settings for now
- One accordion can be pinned and moves to the top of the list
- The first accordion is expanded by default
- Accordions are disabled when device is offline with a message
- Accordions are disabled when notifications are disabled in system settings and a message is displayed
- Subscriptions are managed by storing the selected fire zone unit id (from the API `Shape` table) in redux and persisted with `@capacitor/preferences` for now. Probably end up syncing with backend.
- Fire centres and containing zones are sourced through an API call and cached locally.

Closes: #5184 

# Test Links:
[Landing Page](https://wps-pr-5212-e1e498-dev.apps.silver.devops.gov.bc.ca/)
[MoreCast](https://wps-pr-5212-e1e498-dev.apps.silver.devops.gov.bc.ca/morecast)
[Percentile Calculator](https://wps-pr-5212-e1e498-dev.apps.silver.devops.gov.bc.ca/percentile-calculator)
[C-Haines](https://wps-pr-5212-e1e498-dev.apps.silver.devops.gov.bc.ca/c-haines)
[FireCalc](https://wps-pr-5212-e1e498-dev.apps.silver.devops.gov.bc.ca/fire-behaviour-calculator)
[FireCalc bookmark](https://wps-pr-5212-e1e498-dev.apps.silver.devops.gov.bc.ca/fire-behaviour-calculator?s=266&f=c5&c=NaN&w=20,s=286&f=c7&c=NaN&w=16,s=1055&f=c7&c=NaN&w=NaN,s=305&f=c7&c=NaN&w=NaN,s=344&f=c5&c=NaN&w=NaN,s=346&f=c7&c=NaN&w=NaN,s=328&f=c7&c=NaN&w=NaN,s=1399&f=c7&c=NaN&w=NaN,s=334&f=c7&c=NaN&w=NaN,s=1082&f=c3&c=NaN&w=NaN,s=388&f=c7&c=NaN&w=NaN,s=309&f=c7&c=NaN&w=16,s=306&f=c7&c=NaN&w=NaN,s=1029&f=c7&c=NaN&w=NaN,s=298&f=c7&c=NaN&w=NaN,s=836&f=c7&c=NaN&w=NaN,s=9999&f=c7&c=NaN&w=NaN)
[Auto Spatial Advisory (ASA)](https://wps-pr-5212-e1e498-dev.apps.silver.devops.gov.bc.ca/auto-spatial-advisory)
[HFI Calculator](https://wps-pr-5212-e1e498-dev.apps.silver.devops.gov.bc.ca/hfi-calculator)
[SFMS Insights](https://wps-pr-5212-e1e498-dev.apps.silver.devops.gov.bc.ca/insights)
[Fire Watch](https://wps-pr-5212-e1e498-dev.apps.silver.devops.gov.bc.ca/fire-watch)


mobile/asa-go/src/slices/settingsSlice.ts
```diff
@@ -0,0 +1,182 @@
...
+  }
+};
+
+export const initSubscriptions = (): AppThunk => async (dispatch) => {
+  const result = await Preferences.get({
+    key: SUBSCRIPTIONS_KEY,
+  });
+  try {
+    if (result.value) {
+      const subs = JSON.parse(result.value);
```
conbrad: Nit: could validate we get back `number[]` in case settings got corrupted with `Array.isArray(subs)`

backend/packages/wps-shared/src/wps_shared/db/crud/auto_spatial_advisory.py
```diff
@@ -959,3 +959,16 @@ async def gather_zone_status_inputs(
...
+
+
+async def get_fire_centre_info(db_session: AsyncSession):
+    stmt = (
+        select(
+            Shape.id, Shape.label, FireCentre.name
+        )
+        .join(ShapeType, ShapeType.id == Shape.shape_type)
+        .join(FireCentre, FireCentre.id == Shape.fire_centre)
+        .where(ShapeType.name == "fire_zone_unit")
```
conbrad: Should we order the results with `.order_by(FireCentre.name, Shape.label)`?

mobile/asa-go/src/slices/settingsSlice.ts
```diff
@@ -0,0 +1,182 @@
...
+  getFireCenterInfoSuccess,
+  setPinnedFireCentre,
+  setSubscriptions,
+} = settingsSlice.actions;
+
+export default settingsSlice.reducer;
+
+export const initPinnedFireCentre = (): AppThunk => async (dispatch) => {
+  // Read the pinned fire centre from @capacitor/preferences and store in redux state if it exists.
+  const pinnedFireCentre = await Preferences.get({
```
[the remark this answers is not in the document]
dgboss: `Preferences.get()` doesn't appear to throw, it returns either a string value or null. The underlying native implementations it accesses also return string or null without throwing.

mobile/asa-go/src/components/settings/Settings.tsx
```diff
@@ -0,0 +1,216 @@
...
+  // Check push notification permission
+  const refreshNotificationPermission = useCallback(async () => {
+    try {
+      setCheckingPerm(true);
+      const res = await FirebaseMessaging.checkPermissions();
+      setReceivePermission(res.receive ?? "unknown");
+    } catch (err) {
+      // If the plugin throws (rare), mark as unknown but don't break the UI
+      setReceivePermission("unknown");
+      console.error("checkPermissions failed", err);
```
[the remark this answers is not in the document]
dgboss: This really should never happen. I believe it can only occur there is a breakdown of the bridge between capacitor and the native device code. We have much bigger problems if this happens!

mobile/asa-go/src/components/settings/Settings.tsx
```diff
@@ -0,0 +1,264 @@
...
+        >
+          <AlertTitle>Push notifications disabled</AlertTitle>
+          Notifications are currently disabled in your system settings. To
+          receive alerts, enable notifications for this app in your device
+          settings.
+          <Box mt={1} display="flex" gap={1}>
+            <Button
+              size="small"
+              variant="outlined"
+              onClick={refreshNotificationPermission}
```
brettedw: This might be specific to ios, but I recently learned this while working on notifications:
>The first time your app makes this authorization request, the system prompts the person to grant or deny the request and records that response. Subsequent authorization requests don’t prompt the person.

https://developer.apple.com/documentation/usernotifications/asking-permission-to-use-notifications

So on ios this button doesn't actually do anything. On ios, instead of re-prompting, apps will either inform the user of their settings, or have a button that will take the user to the notification settings for that app, which is likely far more difficult

dgboss: The permission check combined with useCallback and local state was causing me grief. I offloaded the permission check logic into a thunk which gets called based on its dependencies (`isVisible` for tab visibility and `isActive` when the app is foregrounded).
https://github.com/bcgov/wps/pull/5212/changes/1419b64833566a53f07022f9b11f899141479b4a

backend/packages/wps-shared/src/wps_shared/db/crud/auto_spatial_advisory.py
```diff
@@ -959,3 +959,16 @@ async def gather_zone_status_inputs(
     )
 
     return thresholds_lut, hfi_rows
+
+
+async def get_fire_centre_info(db_session: AsyncSession):
+    stmt = (
+        select(
+            Shape.id, Shape.label, FireCentre.name
```
brettedw: Do you think we should use Shape's `id` or `source_identifier` here? I think the rest of the ASA endpoints use `Shape.source_identifier` for their identifier

brettedw: Ah the endpoint Conor put up uses `Shape.id` in the db model, can't decide if it makes that much of a difference. I think `source_identifier` would be more stable if we ever had to rebuild the `Shape` table, which seems unlikely....

conbrad: Hmm good point, I think we should change it to `source_identifier` if that's what's used elsewhere.

PR #5373, opened 2026-05-01
ASA Go: Auth & Public API support

### LINKED ISSUE #5363: ASA Go: Support auth'd routes for IDIR users, public routes for public users
**Describe the task**

Isolate the public route/service for public access from the existing authenticated one.

**Acceptance Criteria**
- [ ] Public users hit the public API routes
- [ ] IDIR users hit the auth'd API routes

**Additional context**
- Flag which base url to use based on whether user is logged in or not
- Or here

brettedw:
Adds logic in ASA Go to use either the public or private API depending on authentication.

`Continue as guest` link is temporary for testing purposes, to be reverted before merging

closes #5363 

# Test Links:
[Landing Page](https://wps-pr-5373-e1e498-dev.apps.silver.devops.gov.bc.ca/)
[MoreCast](https://wps-pr-5373-e1e498-dev.apps.silver.devops.gov.bc.ca/morecast)
[Percentile Calculator](https://wps-pr-5373-e1e498-dev.apps.silver.devops.gov.bc.ca/percentile-calculator)
[C-Haines](https://wps-pr-5373-e1e498-dev.apps.silver.devops.gov.bc.ca/c-haines)
[FireCalc](https://wps-pr-5373-e1e498-dev.apps.silver.devops.gov.bc.ca/fire-behaviour-calculator)
[FireCalc bookmark](https://wps-pr-5373-e1e498-dev.apps.silver.devops.gov.bc.ca/fire-behaviour-calculator?s=266&f=c5&c=NaN&w=20,s=286&f=c7&c=NaN&w=16,s=1055&f=c7&c=NaN&w=NaN,s=305&f=c7&c=NaN&w=NaN,s=344&f=c5&c=NaN&w=NaN,s=346&f=c7&c=NaN&w=NaN,s=328&f=c7&c=NaN&w=NaN,s=1399&f=c7&c=NaN&w=NaN,s=334&f=c7&c=NaN&w=NaN,s=1082&f=c3&c=NaN&w=NaN,s=388&f=c7&c=NaN&w=NaN,s=309&f=c7&c=NaN&w=16,s=306&f=c7&c=NaN&w=NaN,s=1029&f=c7&c=NaN&w=NaN,s=298&f=c7&c=NaN&w=NaN,s=836&f=c7&c=NaN&w=NaN,s=9999&f=c7&c=NaN&w=NaN)
[Auto Spatial Advisory (ASA)](https://wps-pr-5373-e1e498-dev.apps.silver.devops.gov.bc.ca/auto-spatial-advisory)
[HFI Calculator](https://wps-pr-5373-e1e498-dev.apps.silver.devops.gov.bc.ca/hfi-calculator)
[SFMS Insights](https://wps-pr-5373-e1e498-dev.apps.silver.devops.gov.bc.ca/insights)
[Fire Watch](https://wps-pr-5373-e1e498-dev.apps.silver.devops.gov.bc.ca/fire-watch)
[Weather Toolkit](https://wps-pr-5373-e1e498-dev.apps.silver.devops.gov.bc.ca/weather-toolkit)


.github/workflows/asa_go_ios_build_deploy.yml
```diff
@@ -2,7 +2,7 @@ name: Publish ASA Go iOS
 
 env:
   appBuildNumber: ${{ github.run_number }}
-  appBuildVersion: "1.1.1"
+  appBuildVersion: ${{ vars.ASA_GO_VERSION }}
```
dgboss: Does this var autoincrement?

mobile/asa-go/src/api/axios.test.ts
```diff
@@ -0,0 +1,54 @@
+// @vitest-environment node
+
+import { beforeEach, describe, expect, it, vi } from "vitest";
+
+const loadAxiosApi = async () => {
+  vi.resetModules();
```
dgboss: Do we need both calls to `vi.resetModules()`? One here and one in the `beforeEach`.

.github/workflows/asa_go_ios_build_deploy.yml
brettedw: No, it's stored as a variable in github actions so we can change the value without needing a PR, if that's desired. I also put the API URL's in `vars` so that we can see what these variables are set to. It's just an alternative to using GH secrets if something doesn't really need to be a secret

conbrad: I would prefer if we keep the version number version controlled for visibility and lessening dependence on GH.

dgboss: I see. Could you please update the `Distributing` section of the readme to mention this?

mobile/asa-go/src/api/axios.test.ts
```diff
@@ -0,0 +1,54 @@
...
+  vi.doMock("@/utils/env", () => ({
+    API_BASE_URL: "https://auth-api.example.com/api",
```
conbrad: nit: why `auth-api.example.com` for base url?

.github/workflows/asa_go_ios_build_deploy.yml
brettedw: I've reverted the version variable

mobile/asa-go/src/utils/axiosInterceptor.ts
```diff
@@ -1,30 +1,37 @@
...
+export const configureAuthenticatedApiInterceptors = () => {
+  if (interceptorsConfigured) {
+    return;
+  }
+
+  interceptorsConfigured = true;
+
+  // add auth headers on the authenticated client
+  authenticatedApi.interceptors.request.use((config) => {
+    const token = selectToken(store.getState());
```
conbrad: Can we simplify by just dynamically setting `config.baseURL` inside the existing authenticatedApi request interceptor to `API_BASE_URL` if `token` else `API_PUBLIC_BASE_URL`?

conbrad: e.g.

```
       const token = selectToken(store.getState());
       if (!isNil(token)) {
            config.headers.Authorization = `Bearer ${token}`;
            config.baseURL = API_BASE_URL;
        }  else  {
            config.baseURL = API_PUBLIC_BASE_URL;
        }
        return config;
```

brettedw: We could possibly go that way, you want the interceptor to handle both auth and route selection? I believe that second path would also have to delete any `config.headers.Authorization`

I'm trying to think if that could cause any issues down the road. Like a user ends up back on the splash/login screen with a stale/expired token in redux and select `Continue as guest`. I'm not sure if that's possible but I guess I'm trying to think if existence of a token choosing the route could lead to any bugs down the road

mobile/asa-go/src/api/axios.test.ts
brettedw: Good call, I just kept the one in `beforeEach`: [e7dd1f8](https://github.com/bcgov/wps/pull/5373/commits/e7dd1f8d73092750b547f2d8d61f05aae9b2f7f3)

mobile/asa-go/src/api/axios.test.ts
brettedw: authenticated vs public api, but I can change it for clarity. I see how it could be read as the api you call to authenicate

mobile/asa-go/src/utils/axiosInterceptor.ts
conbrad: > you want the interceptor to handle both auth and route selection?

Yeah it's already where the header configuration is handled, I think it makes sense to keep all request configuration in one place.

> second path would also have to delete any `config.headers.Authorization`

That's a good idea safety-wise, although if we have an expired token the backend should reject the request anyway.

> Like a user ends up back on the splash/login screen with a stale/expired token in redux and select `Continue as guest`. 

In this case the existing response interceptor already covers it. When a token expires:
  1. Request goes out with expired token + `API_BASE_URL` → server returns `401`
  2. Response interceptor catches the `401` → `dispatches resetAuthentication()` → clears token from store
  3. Next request: token is `null` → falls back to `API_PUBLIC_BASE_URL`, no auth header

brettedw: That covers the expired token after a request, agreed
 
The case I’m thinking about is slightly different: route selection would happen before we know whether the token is expired, and it would be based only on token existence. So if someone chooses `Continue as guest` while a token is still present, the first request would still go to the authenticated API. If the token is expired, they get a 401 even though it should be going to the public route, if the token is valid, they silently use authenticated routes even though they chose guest

I'm not sure if that's possible at the moment but I think it does open a potential confusing bug.

I can simplify as suggested if you'd like, but do you think it might also be worth gating on `isAuthenticated`?

conbrad: Oh I see what you mean, what about putting guest/idir into redux state that could be checked as well?

dgboss: Having guest vs idir in redux will eventually be needed (or make life easier) when we implement idir gated features.

brettedw: Made some of the changes discussed, adding a `type AuthSessionMode = "login" | "guest" | "authenticated"` to redux

dgboss (approved):
LGTM!

mobile/asa-go/src/utils/axiosInterceptor.test.ts
conbrad, line 1: Nice tests!

backend/packages/wps-api/src/app/routers/asa_go.py
conbrad, line 1: Nice router composition!

conbrad (approved):
🚀

PR #5400, opened 2026-05-11
Keycloak  - offline_access & auto login

### LINKED ISSUE #5335: ASA Go: Support offline tokens in the mobile Keycloak plugin
**Description**
The Keycloak server is configured for 60-day offline sessions. The mobile plugin needs two changes to take advantage of them: Android is missing the offline_access scope, and iOS has no mechanism to persist tokens across app restarts.

Keycloak issues an offline token (returned in the standard `refresh_token` field) when the client requests the `offline_access` scope. This token has a 60-day TTL and survives server restarts. The JS app always calls `authenticate()` on startup, so the plugin itself must handle silent re-authentication using a stored token — there is no token-check logic in the JS layer.

Preconditions (server-side, assumed already configured)
  - Users have the realm-level `offline_access` role mapping
  - The client has `offline_access` configured as an Optional client scope
  - If these are missing, Keycloak will silently issue a regular refresh token even when the scope is requested — no error is returned

**Acceptance criteria**
  - [ ] Android: user is not prompted to re-authenticate within the 60-day offline session window, including across cold app restarts
  - [ ] iOS: after initial login, subsequent app starts resolve `authenticate()` silently without a browser redirect, for up to 60 days
  - Both: if the offline token is expired or revoked, the user is redirected to the browser login as normal

**Additional Context**
Android (KeycloakPlugin.java:128):
  Add offline_access to the authorization scope:
  - `.setScope("openid profile offline_access")`
  - Android already persists `AuthState` via `SharedPreferences`, so the offline token will survive restarts and `AppAuth` will use it for silent refresh automatically.

iOS (KeycloakPlugin.swift / KeycloakServices.swift):
  iOS currently stores `authState` in memory only — it is lost on every app kill. Three additions are needed:
  1. Add a KeychainService that serializes `OIDAuthState` to/from the iOS Keychain using NSKeyedArchiver + SecItem* APIs, using kSecAttrAccessibleAfterFirstUnlock so background refresh
  works after device restart
  2. Save auth state to Keychain after initial browser authentication and after each subsequent token refresh
  3. In authenticate(), attempt a silent refresh using the stored Keychain state before falling back to the browser flow — if OIDAuthState.isAuthorized and performTokenRefresh
  succeeds, resolve the call immediately without launching a browser

Docs: https://www.keycloak.org/docs/latest/server_admin/index.html#_offline-access

brettedw:
- Before this PR, ASA Go was not reliably using Keycloak’s 60-day offline sessions on mobile.

-Android was not requesting offline_access, so its refresh token payload had `"typ": "Refresh"` instead of the expected `"typ": "Offline"`. iOS was receiving offline tokens, but its `AppAuth` state was not persisted somewhere the app could reuse after restart. Android also persisted AppAuth state but did not check the stored state before starting a new browser login flow.

- Login was relying mostly on browser session cookies stored on the device. On iOS, clearing recent Safari cache/cookies could cause ASA Go to prompt for login again

- This PR
  - Adds the `offline_access` scope to Android authentication requests.
  - Stores Android AppAuth state in encrypted DataStore using Tink, with migration from the legacy SharedPreferences state.
  - Stores iOS `OIDAuthState` in Keychain.
  - Checks stored native auth state before opening the login browser.
  - Refreshes stored auth state when possible and returns a current access token to the app.
  - Avoids opening the login browser when a valid stored offline token can be reused.
  - Adds native refreshAuthState() and clearAuthState() APIs.
  - Updates ASA Go’s 401 handling to try a native token refresh and retry the failed request once before clearing auth.
  - Falls back to browser login when the stored offline token is expired, revoked, missing, or otherwise unusable.
  - Pauses automatic native token refresh while the app is backgrounded and resumes/refreshes as needed when the app returns to the foreground.
  - Added a 401 recovery path for authenticated requests. If an API request fails because the Redux access token is stale, ASA Go asks the native AppAuth state to refresh the token and retries the failed request once. This covers timing gaps where the app resumes or sends a request before automatic token refresh has updated Redux. Guest/login sessions do not trigger this recovery path.

closes #5335 
  
# Test Links:
[Landing Page](https://wps-pr-5400-e1e498-dev.apps.silver.devops.gov.bc.ca/)
[MoreCast](https://wps-pr-5400-e1e498-dev.apps.silver.devops.gov.bc.ca/morecast)
[Percentile Calculator](https://wps-pr-5400-e1e498-dev.apps.silver.devops.gov.bc.ca/percentile-calculator)
[C-Haines](https://wps-pr-5400-e1e498-dev.apps.silver.devops.gov.bc.ca/c-haines)
[FireCalc](https://wps-pr-5400-e1e498-dev.apps.silver.devops.gov.bc.ca/fire-behaviour-calculator)
[FireCalc bookmark](https://wps-pr-5400-e1e498-dev.apps.silver.devops.gov.bc.ca/fire-behaviour-calculator?s=266&f=c5&c=NaN&w=20,s=286&f=c7&c=NaN&w=16,s=1055&f=c7&c=NaN&w=NaN,s=305&f=c7&c=NaN&w=NaN,s=344&f=c5&c=NaN&w=NaN,s=346&f=c7&c=NaN&w=NaN,s=328&f=c7&c=NaN&w=NaN,s=1399&f=c7&c=NaN&w=NaN,s=334&f=c7&c=NaN&w=NaN,s=1082&f=c3&c=NaN&w=NaN,s=388&f=c7&c=NaN&w=NaN,s=309&f=c7&c=NaN&w=16,s=306&f=c7&c=NaN&w=NaN,s=1029&f=c7&c=NaN&w=NaN,s=298&f=c7&c=NaN&w=NaN,s=836&f=c7&c=NaN&w=NaN,s=9999&f=c7&c=NaN&w=NaN)
[Auto Spatial Advisory (ASA)](https://wps-pr-5400-e1e498-dev.apps.silver.devops.gov.bc.ca/auto-spatial-advisory)
[HFI Calculator](https://wps-pr-5400-e1e498-dev.apps.silver.devops.gov.bc.ca/hfi-calculator)
[SFMS Insights](https://wps-pr-5400-e1e498-dev.apps.silver.devops.gov.bc.ca/insights)
[Fire Watch](https://wps-pr-5400-e1e498-dev.apps.silver.devops.gov.bc.ca/fire-watch)
[Weather Toolkit](https://wps-pr-5400-e1e498-dev.apps.silver.devops.gov.bc.ca/weather-toolkit)


mobile/keycloak/android/src/main/java/ca/bcgov/plugins/keycloak/Keycloak.java
```diff
@@ -60,17 +64,129 @@ private AuthState restoreAuthState() {
...
+
+    private String readAuthStateJson() {
+        try {
+            SharedPreferences encryptedPrefs = getEncryptedPreferences();
+            String encryptedStateJson = encryptedPrefs.getString(KEY_AUTH_STATE, null);
+            if (encryptedStateJson != null) {
+                return encryptedStateJson;
+            }
+
+            String legacyStateJson = context
```
conbrad: Thanks for the backwards compat here!

mobile/keycloak/android/src/main/java/ca/bcgov/plugins/keycloak/Keycloak.java
```diff
@@ -60,17 +64,129 @@ private AuthState restoreAuthState() {
...
+            return legacyStateJson;
+        } catch (GeneralSecurityException | IOException e) {
+            Log.e(TAG, "Failed to read auth state from encrypted storage", e);
+            return null;
+        }
+    }
+
+    private SharedPreferences getEncryptedPreferences()
+        throws GeneralSecurityException, IOException {
+        MasterKey masterKey = new MasterKey.Builder(context)
```
conbrad: Looks like `MasterKey` is deprecated and we should be using [javax.crypto.KeyGenerator](https://developer.android.com/reference/javax/crypto/KeyGenerator.html)

https://developer.android.com/reference/androidx/security/crypto/MasterKey

mobile/keycloak/android/src/main/java/ca/bcgov/plugins/keycloak/Keycloak.java
```diff
@@ -60,17 +64,129 @@ private AuthState restoreAuthState() {
...
+            .setKeyScheme(MasterKey.KeyScheme.AES256_GCM)
+            .build();
+
+        return EncryptedSharedPreferences.create(
```
conbrad: Looks like `EncryptedSharedPreferences` is deprecated too: https://developer.android.com/reference/androidx/security/crypto/EncryptedSharedPreferences

Looks like the way to go now is Jetpack DataStore with Google Tink: https://medium.com/@n20/encryptedsharedpreferences-is-deprecated-what-should-android-developers-use-now-7476140e8347

brettedw: Nice catches on the deprecations! Unfortunately it seems like the Jetpack DataStore is in kotlin which might be a bigger refactor to bring in...maybe I'll see if there's any java alternatives

conbrad: Kotlin/Java run on the JVM and have a lot of interop, it looks like you can use it with RxJava. All of the answers here build on one another, the last being a clean solution: https://stackoverflow.com/questions/66067126/how-to-implement-datastore-in-java-based-android-app

brettedw: Now using RxJava and tink

mobile/keycloak/android/src/main/java/ca/bcgov/plugins/keycloak/Keycloak.java
brettedw: MasterKey is no longer being used

mobile/asa-go/src/components/PublicLoginButton.tsx
conbrad, line 1: Maybe we can rename this to `ContinueAsGuestButton` for consistency.

mobile/keycloak/ios/Sources/KeycloakPlugin/KeycloakPlugin.swift
```diff
@@ -51,6 +73,99 @@ public class KeycloakPlugin: CAPPlugin, CAPBridgedPlugin {
...
+                call.resolve([
+                    "isAuthenticated": false,
+                    "error": "refresh_failed",
+                    "errorDescription": error ?? "Failed to refresh stored auth state",
+                ])
+            }
+        }
+    }
+
+    private func refreshExistingAuthState(
```
conbrad, line 136: Some duplication with `refreshAuthState` above, do we need both or can we deduplicate some code?

mobile/keycloak/android/src/main/java/ca/bcgov/plugins/keycloak/Keycloak.java
```diff
@@ -32,19 +31,20 @@ public class Keycloak {
...
     public Keycloak(Context context) {
-        this.context = context;
         this.authService = new AuthorizationService(context);
+        this.authStateStore = createAuthStateStore(context);
         this.authState = restoreAuthState();
         this.refreshHandler = new Handler(Looper.getMainLooper());
         setupAutomaticRefresh();
     }
 
     private AuthState restoreAuthState() {
```
conbrad: All of this runs on the Android main thread (Capacitor calls `KeycloakPlugin.load()` → `new Keycloak(context)` → `restoreAuthState()` → `authStateStore.read()` → `DataStore.data().firstOrError().blockingGet()`. ). On cold start with slow flash I/O, this could exceed Android's 5-second ANR threshold and force-closes the app. `write()` (called from `persistAuthState()` inside `performActionWithFreshTokens` callbacks) and `clear()` carry the same risk if AppAuth delivers those callbacks on the main thread.

We can use a shared IO executor thread pool to keep this off the main thread like: https://github.com/bcgov/wps/compare/task/keycloak-asago-tweaks...task/executor-non-blocking-fix

mobile/asa-go/src/slices/authenticationSlice.ts
```diff
@@ -137,24 +120,15 @@ export const authenticate = (): AppThunk => dispatch => {
...
   // Handle token refresh callback function
-  const handleTokenRefresh = (tokenResponse: {
-    accessToken: string
-    idToken: string
-    refreshToken?: string
-    tokenType?: string
-    expiresIn?: number
-    scope?: string
-  }) => {
+  const handleTokenRefresh = (tokenResponse: KeycloakTokenResponse) => {
```
conbrad: Less likely to happen now but cheap to fix:

Every time `authenticate()` is dispatched a second handleTokenRefresh closure is registered. After N login sessions, N closures fire per native `tokenRefresh` event: N dispatches of `authenticateFinished`, N `Sentry.setUser` calls. The `PluginListenerHandle` returned by addListener is discarded.

We can have a module level handle to register the tokenRefresh handle once to avoid, like: https://github.com/bcgov/wps/compare/task/keycloak-asago-tweaks...task/single-login-listener

mobile/keycloak/ios/Sources/KeycloakPlugin/KeycloakPlugin.swift
```diff
@@ -60,23 +175,126 @@ public class KeycloakPlugin: CAPPlugin, CAPBridgedPlugin {
...
+
+            if success {
+                self.authState = existingAuthState
+                self.services.authStateStorageService.saveAuthState(existingAuthState)
+                self.startTokenRefreshManager(authState: existingAuthState)
+                if let tokenResponse = tokenResponse {
+                    self.notifyListeners("tokenRefresh", data: tokenResponse)
+                }
+            } else {
+                self.notifyListeners(
```
conbrad, line 285: We should `clearStoredAuthState()` here to avoid looping retries, see: https://github.com/bcgov/wps/compare/task/keycloak-asago-tweaks...task/failed-foreground-fix-swift

mobile/asa-go/src/components/PublicLoginButton.tsx
brettedw: This was named without knowing what the button would say. The current design text says `Log-in as guest`. I'd propose leaving as is until the button is settled and implemented in a separate PR since it isn't being used right now

mobile/keycloak/ios/Sources/KeycloakPlugin/KeycloakPlugin.swift
brettedw: Both are needed, but i did dedupe some in [4d22a0c](https://github.com/bcgov/wps/pull/5400/commits/4d22a0cbef552e7400e3f80895844b507cd78982)

mobile/keycloak/ios/Sources/KeycloakPlugin/KeycloakPlugin.swift
brettedw: done in [487bf76](https://github.com/bcgov/wps/pull/5400/commits/487bf7686f937861f12a031ad7d9fb367d79a66c)

mobile/asa-go/src/slices/authenticationSlice.ts
brettedw: Done in: [883a920](https://github.com/bcgov/wps/pull/5400/commits/883a92099d0dfa2000228e5cec1fa3135d0dc25d)
With a bit of a workaround for tests

mobile/keycloak/android/src/main/java/ca/bcgov/plugins/keycloak/Keycloak.java
brettedw: - `AuthStateStore.read()` now runs on a shared IO_EXECUTOR during startup instead of blocking `KeycloakPlugin.load()` on the main thread.
  - `persistAuthState()` and `clearAuthState()` now submit DataStore writes/clears to the same executor, so AppAuth callbacks do not block the main thread.
  - `authenticateWithStoredState()` waits for the startup restore future if restore is still in progress, then decides whether to reuse stored auth or fall back to browser login. This avoids the cold-start race where async restore could otherwise make Android open browser login before stored auth was loaded.

mobile/asa-go/src/slices/authenticationSlice.ts
conbrad: I'd rather we refresh module state with `vi.resetModules` to avoid polluting the real code with test code, something like this:

```
describe('StatefulModule', () => {
  beforeEach(() => {
    vi.resetModules()
  })

  it('resets state', async () => {
    const { count } = await import('./statefulModule')
    expect(count).toBe(0)
  })

  it('starts fresh', async () => {
    const { count } = await import('./statefulModule')
    expect(count).toBe(0) // Independent of previous test
  })
})
```

mobile/keycloak/android/src/main/java/ca/bcgov/plugins/keycloak/Keycloak.java
conbrad: I think my original suggestion wasn't complete and missed a few branches and I think the manual `CompletableFuture` handling here can make it easy to miss a branch. I think we can address that by chaining work with something like:

```
private CompletableFuture<Void> restoreAuthStateAsync() {
    if (authStateStore == null) {
        Log.d(TAG, "No auth state store, creating new");
        return CompletableFuture.completedFuture(null);
    }

    return CompletableFuture
            .supplyAsync(authStateStore::read, IO_EXECUTOR)
            .thenCompose(stateJson -> {
                if (stateJson == null) {
                    Log.d(TAG, "No saved auth state, creating new");
                    return CompletableFuture.completedFuture(null);
                }

                try {
                    AuthState restoredAuthState =
                            AuthState.jsonDeserialize(stateJson);

                    CompletableFuture<Void> future = new CompletableFuture<>();

                    refreshHandler.post(() -> {
                        if (!authState.isAuthorized()) {
                            authState = restoredAuthState;
                            Log.d(TAG, "Restored auth state from storage");
                            checkAndRefreshToken();
                        }
                        future.complete(null);
                    });

                    return future;

                } catch (JSONException e) {
                    Log.e(TAG, "Failed to restore auth state", e);
                    return CompletableFuture.completedFuture(null);
                }
            });
}
```

mobile/asa-go/src/slices/authenticationSlice.ts
brettedw: done here: [2aa7f17](https://github.com/bcgov/wps/pull/5400/commits/2aa7f174ddad20553bb7eb97ca6494574d18a042)

mobile/keycloak/android/src/main/java/ca/bcgov/plugins/keycloak/Keycloak.java
brettedw: Thanks, done in [985711d](https://github.com/bcgov/wps/pull/5400/commits/985711dcf1dd4e9bbe56faf0fd49be68f9beecf3)

conbrad (approved):
Nice work!

PR #5419, opened 2026-05-19
SMURFI endpoints, crud and schema

dgboss:
Some new endpoints, schema and crud ops.


conbrad (approved):

backend/packages/wps-api/src/app/routers/smurfi.py
```diff
@@ -18,38 +42,177 @@
...
-
-
-@router.post("/change_spot_status/{spot_id}/{new_status}", response_model=None)
-async def change_spot_status(spot_id: int, new_status: SpotRequestStatusEnum):
-    spot_service = SpotService()
-    await spot_service.change_spot_status(spot_id, new_status)
+@router.post("/spot_request", response_model=SpotRequestResponse)
+async def upsert_spot_request(data: SpotRequestData):
+    x, y = PointTransformer(
+        SpatialReferenceSystem.WGS84.code, NAD83_BC_ALBERS
```
brettedw, line 56: I missed this in the models and maybe not worth changing now, but wondering if it would have been easier/faster to just store WGS84 lat/long

backend/packages/wps-api/src/app/routers/smurfi.py
```diff
@@ -18,38 +42,177 @@
...
+    async with get_async_write_session_scope() as session:
+        if data.id is None:
+            result = await create_spot_request(session, spot_request)
+        else:
+            result = await update_spot_request(session, spot_request)
+            if result is None:
+                raise HTTPException(
+                    status_code=status.HTTP_404_NOT_FOUND, detail=f"SpotRequest {data.id} not found"
+                )
+        for subscriber in data.subscribers:
```
brettedw: I think this might create subscriber rows on every call, creation or update, rather than possibly updating subscibers?

backend/packages/wps-api/src/app/routers/smurfi.py
dgboss: Yeah, it is annoying but I'd like to use the same spatial ref for all data in the db. I think that is recommended best practice by the architecture team.

backend/packages/wps-shared/src/wps_shared/db/crud/smurfi.py
```diff
@@ -0,0 +1,136 @@
...
+        existing.confidence = updated.confidence
+        existing.fire_size = updated.fire_size
+        existing.representative_station_codes = updated.representative_station_codes
+        existing.for_date = updated.for_date
+        await session.flush()
+    return existing
+
+
+async def update_spot_tabular_weather(session: AsyncSession, updated: SpotTabularWeather):
+    result = await session.execute(select(SpotTabularWeather).where(SpotTabularWeather.id == updated.id))
```
brettedw: For safety, should this also select by?:
```
SpotTabularWeather.spot_forecast_id == updated.spot_forecast_id
```

And maybe same with `update_spot_descriptive_weather`?

backend/packages/wps-api/src/app/routers/smurfi.py
dgboss: Fixed in https://github.com/bcgov/wps/pull/5419/changes/767ab1dbc54e4d379d60076527f4cbbb6dee6c2b.

backend/packages/wps-api/src/app/routers/smurfi.py
```diff
@@ -18,38 +42,177 @@
...
+            result = await create_spot_forecast(session, spot_forecast)
+        else:
+            result = await update_spot_forecast(session, spot_forecast)
+            if result is None:
+                raise HTTPException(
+                    status_code=status.HTTP_404_NOT_FOUND, detail=f"SpotForecast {data.id} not found"
+                )
+        await _upsert_descriptive_weather(session, result.id, data)
+        await _upsert_tabular_weather(session, result.id, data)
+    return SpotForecastResponse(spot_forecast=data.model_copy(update={"id": result.id}))
```
brettedw: I'm thinking this only returns the updated parent spot forecast id, but not the descriptive weather or tabular weather id, if it creates new records for those "children". If we're using that returned data on the frontend and it gets updated again, I think we'll end up creating duplicate rows

backend/packages/wps-shared/src/wps_shared/db/crud/smurfi.py
dgboss: Done in https://github.com/bcgov/wps/pull/5419/changes/0210019d3b06103c02699b7db4eb5970238e1a19

backend/packages/wps-api/src/app/routers/smurfi.py
dgboss: Good catch! I'm now returning the descriptive and tabular forecasts. 
https://github.com/bcgov/wps/pull/5419/changes/edec69346a872174bd78f5cb5b95b1157a4eb104

dgboss: There is a potential issue where we end up with the `SpotForecast` being created but something goes wrong with the creation of the child weather forecasts. Something to watch for.

dgboss: I'm also now returning the subscribers with a `SpotRequest` response.
https://github.com/bcgov/wps/pull/5419/changes/fb0f313296cd89fe1b999fe2b4dc6d39093584e8

brettedw (approved):
Nice work! Looks good!

PR #5424, opened 2026-05-20
SMURFI - Spot Request requests

brettedw:
- Wires up Spot Request form to the backend, so requests can be submitted and stored
- Adds Auth to frontend and endpoints
- Adds `additional_information` field to the spot request table
- Makes the email list a bit easier to use, the user doesn't need to press enter to complete an email address entry

### DIFF
--- a/backend/packages/wps-shared/src/wps_shared/schemas/smurfi.py
+++ b/backend/packages/wps-shared/src/wps_shared/schemas/smurfi.py
@@ -18,26 +18,30 @@ class UpdateSubscriberStatusData(BaseModel):
     status: str
 
 
-class SpotRequestData(BaseModel):
+class SpotRequestUpsertData(BaseModel):
     id: int | None = None
     request_reference: str
     fire_number: list[str] | None = None
     fire_centre: int
     status: str = "Requested"
-    requestor_name: str
-    requestor_idir: str
-    requestor_email: str
     request_frequency: list[str] | None = None
     request_type: str = "Full"
     aspect: str | None = None
     elevation: int | None = None
     geographic_description: str
+    additional_information: str | None = None
     latitude: float
     longitude: float
     requested_at: datetime
     start_at: datetime
     end_at: datetime
     subscribers: list[SpotSubscriberData] = []
+
+
+class SpotRequestData(SpotRequestUpsertData):
+    requestor_name: str
+    requestor_idir: str
+    requestor_email: str
 
 
 class SpotRequestResponse(BaseModel):
--- a/backend/packages/wps-shared/src/wps_shared/schemas/smurfi.py
+++ b/backend/packages/wps-shared/src/wps_shared/schemas/smurfi.py
@@ -18,7 +18,7 @@ class UpdateSubscriberStatusData(BaseModel):
     status: str
 
 
-class SpotRequestUpsertData(BaseModel):
+class SpotRequestInput(BaseModel):
     id: int | None = None
     request_reference: str
     fire_number: list[str] | None = None
@@ -38,7 +38,7 @@ class SpotRequestUpsertData(BaseModel):
     subscribers: list[SpotSubscriberData] = []
 
 
-class SpotRequestData(SpotRequestUpsertData):
+class SpotRequestData(SpotRequestInput):
     requestor_name: str
     requestor_idir: str
     requestor_email: str
--- a/web/apps/wps-web/src/features/smurfi/slices/smurfiSlice.ts
+++ b/web/apps/wps-web/src/features/smurfi/slices/smurfiSlice.ts
@@ -1,6 +1,7 @@
 import { createSlice, PayloadAction } from '@reduxjs/toolkit'
 import { SpotFormData } from '@wps/api/schema/spotForecastSchema'
-import { postSpotForecast, SpotForecastOutput } from '@wps/api/SMURFIAPI'
+import { SpotRequestFormData } from '@wps/api/schema/spotRequestSchema'
+import { postSpotForecast, postSpotRequest, SpotForecastOutput, SpotRequestOutput } from '@wps/api/SMURFIAPI'
 import { AppThunk } from 'app/store'
 
 export interface SmurfiState {
@@ -9,6 +10,9 @@ export interface SmurfiState {
   spotForecastSubmitting: boolean
   spotForecastSubmitError: string | null
   submittedSpotForecast: SpotForecastOutput | null
+  spotRequestSubmitting: boolean
+  spotRequestSubmitError: string | null
+  submittedSpotRequest: SpotRequestOutput | null
 }
 
 const initialState: SmurfiState = {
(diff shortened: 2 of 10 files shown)


backend/packages/wps-shared/src/wps_shared/schemas/smurfi.py
```diff
@@ -18,20 +18,18 @@ class UpdateSubscriberStatusData(BaseModel):
     status: str
 
 
-class SpotRequestData(BaseModel):
+class SpotRequestUpsertData(BaseModel):
```
conbrad:
```suggestion
class SpotRequestUpdate(BaseModel):
```

web/apps/wps-web/src/features/smurfi/slices/smurfiSlice.ts
```diff
@@ -40,6 +47,24 @@ const smurfiSlice = createSlice({
...
+      state.spotRequestSubmitting = true
+    },
+    submitSpotRequestFailed(state: SmurfiState, action: PayloadAction<string>) {
+      state.spotRequestSubmitError = action.payload
+      state.spotRequestSubmitting = false
+    },
+    submitSpotRequestSuccess(state: SmurfiState, action: PayloadAction<{ spotRequest: SpotRequestOutput }>) {
+      state.spotRequestSubmitting = false
+      state.spotRequestSubmitError = null
+      state.submittedSpotRequest = action.payload.spotRequest
```
conbrad, line 62: Do we need to store this if we return it from `submitSpotRequest`?

conbrad (approved):

backend/packages/wps-shared/src/wps_shared/schemas/smurfi.py
brettedw: I could change it to that, but it's not necessarily just for updating, it's also for creating Requests

web/apps/wps-web/src/features/smurfi/slices/smurfiSlice.ts
brettedw: I think you're right, we probably don't need to store that in redux

web/apps/wps-web/src/features/smurfi/slices/smurfiSlice.ts
```diff
@@ -73,3 +102,18 @@ export const submitSpotForecast =
...
+
+export const submitSpotRequest =
+  (formData: SpotRequestFormData): AppThunk<Promise<SpotRequestOutput | undefined>> =>
+  async dispatch => {
+    try {
+      dispatch(submitSpotRequestStart())
+
+      const response = await postSpotRequest(formData)
+      dispatch(submitSpotRequestSuccess({ spotRequest: response.spot_request }))
+      return response.spot_request
```
dgboss, line 114: Should we be returning the spot_request here? Or just updating redux?

web/packages/api/src/SMURFIAPI.ts
```diff
@@ -140,6 +141,76 @@ export interface SpotForecastResponse {
...
+  latitude: number
+  longitude: number
+  requested_at: string
+  start_at: string
+  end_at: string
+  subscribers: SpotSubscriber[]
+}
+
+export interface SpotRequestOutput extends SpotRequestInput {
+  id: number
```
dgboss, line 174: Does re-stating `id` field here make it required whereas it isn't required in a `SpotRequestInput`?

backend/packages/wps-shared/src/wps_shared/schemas/smurfi.py
conbrad: Oh ok in that case can we just do `SpotRequest`?

web/apps/wps-web/src/features/smurfi/slices/smurfiSlice.ts
brettedw: I think it should probably either update redux or fetch the request data again, but we're not really using the request data anywhere yet I don't think. We have an endpoint but nothing on the frontend hitting it. We have a lot of mocked data and I haven't dove into that part yet

I can either expand this PR or follow up in another?

web/packages/api/src/SMURFIAPI.ts
```diff
@@ -140,6 +141,76 @@ export interface SpotForecastResponse {
...
+    request_frequency: formData.requestedFrequency,
+    request_type: spotRequestTypeMap[formData.forecastType],
+    aspect: formData.slopeAspect,
+    elevation: Number(formData.elevation),
+    geographic_description: formData.geographicDescription,
+    additional_information: formData.additionalInformation || undefined,
+    latitude: formData.location.latitude,
+    longitude: formData.location.longitude,
+    requested_at: new Date().toISOString(),
+    start_at: formData.forecastStartDate.toISO()!,
```
dgboss: What do you think about normalizing the start time to the 00:00 of the specified day and end time to 23:59 of the specified day?

backend/packages/wps-shared/src/wps_shared/schemas/smurfi.py
brettedw: `SpotRequest` is the table object name, this would technically be a `SpotRequestRequest`, so I was trying to come up with a clear name that doesn't interfere with our other names

web/apps/wps-web/src/features/smurfi/slices/smurfiSlice.ts
dgboss: I can handle it in my upcoming PR where I'm adding an array of `SpotForecastOutput` to the `smurfiSlice`. I'll either refetch all the spot_requests or push this one into the start of the array.

dgboss (approved):
🔥

web/packages/api/src/SMURFIAPI.ts
brettedw: `id` is still required in the Input object, but it can be `null`, re-stating it this way says "the output is a number, not null". Mainly done that way because our endpoint expects an id, though there may be a cleaner way to do it

backend/packages/wps-shared/src/wps_shared/schemas/smurfi.py
dgboss: `SpotRequestInput` to mimic the front-end?

web/apps/wps-web/src/features/smurfi/slices/smurfiSlice.ts
brettedw: @dgboss do you want me to leave this redux state in here for now for your work?

dgboss: You can leave it in for now as a reminder for me to deal with it. Thanks!

web/packages/api/src/SMURFIAPI.ts
dgboss: Looks good to me!

web/packages/api/src/SMURFIAPI.ts
brettedw: Done in [f38fa49](https://github.com/bcgov/wps/pull/5424/commits/f38fa49b70731da55ba4f40d495489f7074a5f91)

backend/packages/wps-shared/src/wps_shared/schemas/smurfi.py
brettedw: Went with `SpotRequestInput`

PR #5443, opened 2026-05-25
Wire VANITY_DOMAIN into nats deployment for WEB_BASE_URL

conbrad:
- Adds `VANITY_DOMAIN` as a required parameter to the nats OpenShift template
- Replaces the missing `env.web-base-url` ConfigMap lookup with an inline value derived from the vanity route
- Passes `VANITY_DOMAIN` through the provision script and deployment workflow

### DIFF
--- a/openshift/scripts/oc_provision_nats.sh
+++ b/openshift/scripts/oc_provision_nats.sh
@@ -33,7 +33,8 @@ OC_PROCESS="oc -n ${PROJ_TARGET} process -f ${PATH_NATS} \
  ${MEMORY_LIMIT:+ "-p MEMORY_LIMIT=${MEMORY_LIMIT}"} \
  ${CPU_REQUEST:+ "-p CPU_REQUEST=${CPU_REQUEST}"} \
  -p CRUNCHYDB_USER=${CRUNCHY_NAME}-${SUFFIX}-pguser-${CRUNCHY_NAME}-${SUFFIX} \
- -p APP_NAME=${APP_NAME}"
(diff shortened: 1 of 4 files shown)


openshift/scripts/oc_provision_nats.sh
```diff
@@ -33,7 +33,8 @@ OC_PROCESS="oc -n ${PROJ_TARGET} process -f ${PATH_NATS} \
  ${MEMORY_LIMIT:+ "-p MEMORY_LIMIT=${MEMORY_LIMIT}"} \
  ${CPU_REQUEST:+ "-p CPU_REQUEST=${CPU_REQUEST}"} \
  -p CRUNCHYDB_USER=${CRUNCHY_NAME}-${SUFFIX}-pguser-${CRUNCHY_NAME}-${SUFFIX} \
- -p APP_NAME=${APP_NAME}"
+ -p APP_NAME=${APP_NAME} \
+ -p VANITY_DOMAIN=${VANITY_DOMAIN}"
```
dgboss, line 37: How is `VANITY_DOMAIN` populated in a prod deployment?

conbrad: It's not deployed, we'll have to adjust for that in prod when we get there.

conbrad: done here: https://github.com/bcgov/wps/pull/5443/commits/329b7434bd550ba1bfe5568c19b5c3d74bab0381

brettedw (approved):

dgboss (approved):

PR #5453, opened 2026-05-26
Mini forecast detail pages

dgboss:
- adds pretty and printable components for mini spot forecasts
- creates a hook that provides data to the forecast display components to dedupe code
- renames printable to print


web/apps/wps-web/src/features/smurfi/components/forecasts/PrintableMiniSpotForecast.tsx
brettedw, line 1: I think the printable mini might be missing the header with the information about the request/forecaster/etc

brettedw (approved):
👌👌👌

conbrad (approved):

PR #5458, opened 2026-05-26
SMURFI: Data Model rework

brettedw:
- Models the backend data as `SpotRequestBase`, `SpotRequestInstance`, `SpotForecast`
- Changes slope/aspect to be a string field with no dropdown. In forecasts I've seen that values can be all over the place. ie 'Steep North face', 'Flat/variable', 'Variable'
   - These are optional on the request form, required on the forecast form
- Adds functionality to prepopulate fire sizes for the forecast form if they exist

### DIFF
--- a/backend/packages/wps-api/src/app/routers/smurfi.py
+++ b/backend/packages/wps-api/src/app/routers/smurfi.py
@@ -147,11 +147,11 @@ def _get_latest_forecast(spot_request: SpotRequestBase) -> SpotForecast | None:
     )
 
 
-def _get_current_instance(spot_request: SpotRequestBase) -> SpotRequestInstance:
-    latest_forecast = _get_latest_forecast(spot_request)
+def _get_current_instance(spot_request_base: SpotRequestBase) -> SpotRequestInstance:
+    latest_forecast = _get_latest_forecast(spot_request_base)
     if latest_forecast is not None:
         return latest_forecast.spot_request_instance
-    return _get_initial_instance(spot_request)
+    return _get_initial_instance(spot_request_base)
 
 
 def _latest_forecast_to_schema(spot_request_base: SpotRequestBase) -> SpotLatestForecastData | None:
@@ -193,7 +193,7 @@ async def upsert_spot_request_endpoint(
                 detail=f"SpotRequestBase {spot_request_base.id} not found",
             )
 
-        initial_instance = await create_spot_request_instance(
+        instance = await create_spot_request_instance(
             session, _build_spot_request_instance(result.id, spot_request_input.initial_instance)
         )
 
@@ -206,14 +206,14 @@ async def upsert_spot_request_endpoint(
             SpotSubscriberData(id=s.id, email=s.email, subscriber_status=s.subscriber_status)
             for s in subscribers
         ]
-        initial_instance_data = _spot_request_instance_to_schema(initial_instance)
+        spot_request_instance = _spot_request_instance_to_schema(instance)
 
     return SpotRequestResponse(
         spot_request=SpotRequestData(
             **spot_request_input.model_dump(exclude={"id", "initial_instance", "subscribers"}),
             id=spot_request_base_id,
-            initial_instance=initial_instance_data,
-            current_instance=initial_instance_data,
+            initial_instance=spot_request_instance,
+            current_instance=spot_request_instance,
             subscribers=subscriber_data,
             requestor_name=requestor.name,
             requestor_idir=requestor.idir,
--- a/backend/packages/wps-api/src/app/routers/smurfi.py
+++ b/backend/packages/wps-api/src/app/routers/smurfi.py
@@ -4,12 +4,13 @@ from datetime import datetime
 from typing import Annotated
 
 from fastapi import APIRouter, Depends, HTTPException, Response, status
-from geoalchemy2.shape import to_shape
 from wps_shared.auth import authentication_required
 from wps_shared.db.crud.smurfi import (
     create_spot_descriptive_weather,
     create_spot_forecast,
+    create_spot_request_instance,
     create_spot_tabular_weather,
+    get_or_create_spot_request_instance,
     get_spot_forecasts_for_request,
     get_spot_requests_for_year,
     get_subscribed_spot_request_ids,
@@ -24,7 +25,8 @@ from wps_shared.db.database import get_async_read_session_scope, get_async_write
 from wps_shared.db.models.smurfi import (
     SpotDescriptiveWeather,
     SpotForecast,
-    SpotRequest,
+    SpotRequestBase,
+    SpotRequestInstance,
     SpotSubscriberStatusEnum,
     SpotTabularWeather,
 )
@@ -42,6 +44,8 @@ from wps_shared.schemas.smurfi import (
     SpotLatestForecastData,
     SpotRequestData,
     SpotRequestInput,
+    SpotRequestInstanceData,
+    SpotRequestInstanceInput,
     SpotRequestListResponse,
     SpotRequestResponse,
     SpotSubscriberData,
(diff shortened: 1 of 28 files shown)


backend/packages/wps-shared/src/wps_shared/db/models/__init__.py
```diff
@@ -43,7 +43,9 @@
 from wps_shared.db.models.sfms_run import SFMSRunLog
 from wps_shared.db.models.fcm import DeviceToken
 from wps_shared.db.models.smurfi import (
-    SpotRequest,
+    SpotRequestBase,
```
dgboss:
```suggestion
```

backend/packages/wps-api/src/app/routers/smurfi.py
```diff
@@ -151,32 +174,48 @@ async def upsert_spot_request_endpoint(
     data: SpotRequestInput, token: Annotated[dict, Depends(authentication_required)]
```
conbrad:
```suggestion
    spot_request_input: SpotRequestInput, token: Annotated[dict, Depends(authentication_required)]
```

backend/packages/wps-api/src/app/routers/smurfi.py
```diff
@@ -91,44 +95,63 @@ def _get_bc_albers_point(latitude: float, longitude: float) -> str:
...
 
-def _build_spot_request_response(
-    data: SpotRequestInput,
-    requestor: SpotRequestor,
-    spot_request_id: int,
-    subscribers: list[SpotSubscriberData],
-) -> SpotRequestResponse:
-    spot_request_response = SpotRequestData(
+def _build_spot_request_instance(
+    spot_request_base_id: int, data: SpotRequestInstanceInput
```
conbrad:
```suggestion
    spot_request_base_id: int, spot_request_instance_input: SpotRequestInstanceInput
```

backend/packages/wps-api/src/app/routers/smurfi.py
```diff
@@ -91,44 +95,63 @@ def _get_bc_albers_point(latitude: float, longitude: float) -> str:
     return f"POINT({x} {y})"
 
 
-def _build_spot_request(data: SpotRequestInput, requestor: SpotRequestor) -> SpotRequest:
+def _build_spot_request_base(data: SpotRequestInput, requestor: SpotRequestor) -> SpotRequestBase:
```
conbrad:
```suggestion
def _build_spot_request_base(spot_request_input: SpotRequestInput, requestor: SpotRequestor) -> SpotRequestBase:
```

backend/packages/wps-api/src/app/routers/smurfi.py
```diff
@@ -91,44 +95,63 @@ def _get_bc_albers_point(latitude: float, longitude: float) -> str:
...
+
+
+def _get_current_instance(spot_request: SpotRequestBase) -> SpotRequestInstance:
+    latest_forecast = _get_latest_forecast(spot_request)
+    if latest_forecast is not None:
+        return latest_forecast.spot_request_instance
+    return _get_initial_instance(spot_request)
+
+
+def _latest_forecast_to_schema(spot_request: SpotRequestBase) -> SpotLatestForecastData | None:
```
conbrad:
```suggestion
def _latest_forecast_to_schema(spot_request_base: SpotRequestBase) -> SpotLatestForecastData | None:
```

backend/packages/wps-shared/src/wps_shared/db/crud/smurfi.py
```diff
@@ -7,37 +7,73 @@
...
+    SpotRequestInstance,
     SpotRequestStatusEnum,
     SpotSubscriber,
     SpotSubscriberStatusEnum,
     SpotTabularWeather,
 )
 
 
-async def create_spot_request(session: AsyncSession, spot_request: SpotRequest):
+async def create_spot_request(session: AsyncSession, spot_request: SpotRequestBase):
```
conbrad:
```suggestion
async def create_spot_request(session: AsyncSession, spot_request_base: SpotRequestBase):
```

backend/packages/wps-shared/src/wps_shared/db/crud/smurfi.py
```diff
@@ -7,37 +7,73 @@
...
     session.add(spot_request)
     await session.flush()
     return spot_request
 
 
-async def upsert_spot_request(session: AsyncSession, spot_request: SpotRequest):
+async def upsert_spot_request(session: AsyncSession, spot_request: SpotRequestBase):
```
conbrad:
```suggestion
async def upsert_spot_request(session: AsyncSession, spot_request_base: SpotRequestBase):
```

backend/packages/wps-api/src/app/routers/smurfi.py
```diff
@@ -151,32 +174,48 @@ async def upsert_spot_request_endpoint(
...
+        result = await upsert_spot_request(session, spot_request_base)
         if result is None:
             raise HTTPException(
                 status_code=status.HTTP_404_NOT_FOUND,
-                detail=f"SpotRequest {spot_request.id} not found",
+                detail=f"SpotRequestBase {spot_request_base.id} not found",
             )
 
-        logger.info("Syncing subscribers for SpotRequest id: %s", result.id)
+        initial_instance = await create_spot_request_instance(
```
conbrad: I think naming here is confusing since this code always creates an `initial_instance`, this should be only created once, but new instances can be added. I think renaming `initial_instance` to `instance` would clarify.


```suggestion
        instance = await create_spot_request_instance(
```

backend/packages/wps-api/src/app/routers/smurfi.py
```diff
@@ -151,32 +174,48 @@ async def upsert_spot_request_endpoint(
...
         subscribers = await sync_spot_subscribers(
             session, result.id, [s.email for s in data.subscribers]
         )
-        spot_request_id = result.id
+        spot_request_base_id = result.id
         subscriber_data = [
             SpotSubscriberData(id=s.id, email=s.email, subscriber_status=s.subscriber_status)
             for s in subscribers
         ]
+        initial_instance_data = _spot_request_instance_to_schema(initial_instance)
```
conbrad:
```suggestion
        spot_request_instance = _spot_request_instance_to_schema(instance)
```

backend/packages/wps-api/src/app/routers/smurfi.py
```diff
@@ -151,32 +174,48 @@ async def upsert_spot_request_endpoint(
...
 
-    return _build_spot_request_response(data, requestor, spot_request_id, subscriber_data)
+    return SpotRequestResponse(
+        spot_request=SpotRequestData(
+            **data.model_dump(exclude={"id", "initial_instance", "subscribers"}),
+            id=spot_request_base_id,
+            initial_instance=initial_instance_data,
```
conbrad: I think currently `initial_instance` and `current_instance` will always be the same. Do we need to store `initial_instance`?

brettedw: Initial instance is the instance of the initial spot request (could have a different location/anything else a `SpotRequestInstance` owns), it can currently be different from the current instance and it's always stored. 

When a request is posted to this endpoint, the initial and current will be the same (a request is being created). When the frontend gets spots requests, the initial could be different from the current if forecasts have created different instances

conbrad: Ok the naming is confusing then, since `initial_instance` will always be a new instance supplied by the frontend. Do we need to store both instances?

brettedw: Good call on the variable names. Changed them on my end so I didn't miss any

brettedw: Not quite sure what you mean by "store both instances"

brettedw: When a request is created, it's only storing the single instance, but returning a common object to the frontend so it's easy to work with

conbrad: `SpotRequestData` has two fields `initial_instance` and `current_instance` that point to the same thing. Why do we need both, doesn't current suffice?

brettedw: Ah I see. We have the functionality to 'View a Request' on the frontend. This is currently set up to view the initial request. I'm not certain if that's the way to go

The viewing of a request shows metadata like when it was submitted, I thought it might make sense to show that request as it was requested. But that also may lead to some confusion

dgboss: ~~We probably need to update the requests table to show the details of a SpotRequestBase and all related child SpotRequestInstances somehow. Perhaps in a master-detail fashion?~~

brettedw: I was thinking the 'View Request' covers viewing the initial `SpotRequestBase` and that maybe viewing the forecasts would cover the "Instances"? Maybe not though...

dgboss: Ignore my last comment, I think instances can be transparent to users since I now understand a new instance is created by a forecast with different coords.

backend/packages/wps-shared/src/wps_shared/db/crud/smurfi.py
```diff
@@ -7,24 +7,60 @@
...
+    return spot_request_instance
+
+
+async def get_matching_spot_request_instance(
+    session: AsyncSession, spot_request_instance: SpotRequestInstance
+) -> SpotRequestInstance | None:
+    result = await session.execute(
+        select(SpotRequestInstance).where(
+            SpotRequestInstance.spot_request_base_id == spot_request_instance.spot_request_base_id,
+            SpotRequestInstance.latitude == spot_request_instance.latitude,
```
dgboss: Are the float comparisons here safe? Do we need to use some sort of buffer/rounding?

backend/packages/wps-api/src/app/routers/smurfi.py
```diff
@@ -148,35 +175,51 @@ def _latest_forecast_to_schema(spot_request: SpotRequest) -> SpotLatestForecastD
...
+        result = await upsert_spot_request(session, spot_request_base)
         if result is None:
             raise HTTPException(
                 status_code=status.HTTP_404_NOT_FOUND,
-                detail=f"SpotRequest {spot_request.id} not found",
+                detail=f"SpotRequestBase {spot_request_base.id} not found",
             )
 
-        logger.info("Syncing subscribers for SpotRequest id: %s", result.id)
+        instance = await create_spot_request_instance(
```
dgboss, line 196: Is there ever a situation where we want to update a request instance instead of creating a new one?

backend/packages/wps-shared/src/wps_shared/db/crud/smurfi.py
brettedw: I think some rounding is a good idea

backend/packages/wps-api/src/app/routers/smurfi.py
brettedw: Hmmm, I'm not sure honestly

backend/packages/wps-shared/src/wps_shared/db/crud/smurfi.py
brettedw: Added a threshold in [e7b0ae6](https://github.com/bcgov/wps/pull/5458/commits/e7b0ae680ae408100f1d591d378c53e2fe9cbd3f)

backend/packages/wps-api/src/app/routers/smurfi.py
brettedw: I think the case for creating new ones is that if a forecast was ever submitted for a specific request, we need to keep that forecast and it's metadata in tact in case it's ever needed

web/apps/wps-web/src/features/smurfi/components/map/SmurfiRequestsMap.tsx
```diff
@@ -102,7 +104,7 @@ const SmurfiRequestsMap = ({ spotRequest }: SmurfiRequestsMapProps) => {
     return () => {
       mapObject.setTarget('')
     }
-  }, [spotRequest.latitude, spotRequest.longitude])
+  }, [spotInstance.latitude, spotInstance.longitude, spotRequest.status])
```
dgboss, line 107: Perhaps best left for a different PR. Having dependencies in this array is forcing the entire ol map to be recreated any time these dependencies change.

backend/packages/wps-api/src/app/routers/smurfi.py
dgboss: Ya, I think it's okay. The only place we ever POST to this endpoint is from the request form.

dgboss (approved):

conbrad (approved):

PR #5464, opened 2026-05-27
SMURFI: Add fire points to map

brettedw:
- Adds fire points to maps, coloured by status similar to: https://wildfiresituation.nrs.gov.bc.ca/map
- Add custom panning to correct some undesired behaviour by built in openlayers panning where it would occasionally just keep panning up. The likely cause was `Overlay` auto-pan firing at `setPosition` time, before React rendered the new popup contents
- Adds `mapLayerVisibility.ts` to centralize layer vis


web/apps/wps-web/src/features/map/mapPopupUtils.ts
```diff
@@ -0,0 +1,46 @@
+import Map from 'ol/Map'
+
+export const panMapToFitElement = (map: Map, element: HTMLElement | null, margin = 16) => {
```
dgboss, line 3: Nice! I was going to suggest we implement something like this as I didn't like the autopan behaviour.

web/apps/wps-web/src/features/smurfi/components/map/SMURFIMap.tsx
```diff
@@ -370,14 +415,16 @@ const SMURFIMap = ({ selectedCoordinates, spotRequests: propSpotRequests }: SMUR
       <Box sx={{ position: 'relative', width: '100%', height: '100%', minHeight: 0, flex: 1 }}>
         <Box ref={mapRef} data-testid={'smurfi-map'} sx={{ width: '100%', height: '100%' }} />
         <SpotMapLayerSwitcher
-          statusOptions={STATUS_FILTER_OPTIONS}
+          statusOptions={SPOT_REQUEST_STATUS_OPTIONS}
```
dgboss: Can the `SpotMapLayerSwitcher` import this directly from `mapLayerVisilbiity` instead of passing it as a prop?

brettedw: yep good call, done in [550cd55](https://github.com/bcgov/wps/pull/5464/commits/550cd557d3ea4ca40b9fe40ebeb5b09ca2475ea3)

web/apps/wps-web/src/features/smurfi/components/map/SMURFIMap.tsx
conbrad, line 1: How about encapsulating the SMURFI map fire-layer click handling into a `CurrentFiresInteraction` OL plugin to eliminate duplication between `SMURFIMap.tsx` and `SmurfiRequestsMap.tsx

We could extend the `ol` `Interaction` that finds fire layers dynamically by a tag set at creation time (e.g. `layer.set('smurfiLayerType', 'currentFires')`). Handles click and exposes `setVisible(bool)` and `setStatuses(statuses[])` that update the layers it finds. Component collapses to `map.addInteraction(fireInteraction)` plus two method calls when React state changes — no layer refs, no visibility/style effects.

Similar approach to what we do in `web/apps/wps-web/src/features/sfmsInsights/components/map/rasterTooltipInteraction.ts`

brettedw: Honestly I kind of got carried away and I'm not even sure the fire points needs to be on the `SmurfiRequestsMap`. What do you think? It really is just meant to be a simple map to show the spot location

dgboss: It would actually be really nice to have one re-useable, flexible map component.

brettedw: You know what though, I think it's likely we'll want to use these fire layers in other applications. Maybe I'll explore centralizing something

conbrad: I'm just concerned with the corner cutting wrt to code clarity -- I'm not sure I should be but it feels like we're at a near beta feature and I think it's worth switching the focus away from experimental code to more maintainable code.

dgboss (approved):

web/apps/wps-web/src/features/smurfi/components/map/SMURFIMap.tsx
brettedw: Ok, pretty big refactor but thoughts on this?

 Refactored current fire map support into reusable OpenLayers utilities outside SMURFI.

  - Moved current fire WFS layer creation, styling, status visibility defaults, and fire attribute extraction into `features/currentFires/map`.
  - Added a reusable `CurrentFiresClickInteraction` for fire point/perimeter click handling.
  - Added a `CurrentFireLayerController` to centralize visibility and status-filter style updates.
  - Moved generic popup panning logic to `features/map/mapPopupUtils`.
  - Kept popup rendering SMURFI-specific since presentation may differ across apps, especially mobile.

conbrad (approved):
Love the refactor!

web/apps/wps-web/src/features/currentFires/map/CurrentFiresClickInteraction.ts
```diff
@@ -0,0 +1,98 @@
...
+  setMap(map: Map | null) {
+    this.removeClickListener()
+    super.setMap(map)
+
+    if (map) {
+      this.listenerKey = map.on('click', this.handleClick.bind(this))
+    }
+  }
+
+  setFireClickCallback(callback: (data: CurrentFireClickData) => void) {
```
dgboss: I don't think these are ever called?

PR #5465, opened 2026-05-27
Allow new request on map click

dgboss:
- adds a NewRequestClickInteraction module for handling click interaction related to opening a popup at a clicked location which allows a user to create a new request 
- the clicked coordinates are passed through to the new request form
- adds a 'pointer' to all popups
- a click on the map when any popup is open closes the popup


brettedw (approved):
This is great!

conbrad (approved):
Awesome, thanks for that!

PR #5470, opened 2026-05-28
Spot Request Edits and Forecast Views

dgboss:
- add an `EditSpotRequestPage` and routing
- only way to edit currently is from the Edit button on `smurfi/requests/:id`
- adds `@router.patch("/spot_requests/{spot_request_id}"` for editing requests. This will edit a request and it's instance in place
  - this does result in an implicit relationship in that the oldest `SpotRequestInstance` (oldest `created_at`) is the initial/request instance (that can be edited), while the others are forecast instances 
- changes SMURFI Map to always show the request location. When a location is clicked, it "explodes" to show connected forecast locations if they are different from the request location.
- adds a map to the forecast view so when a user is viewing a forecast, they can see the exact point it was for 
- adds `ForecastLocationMap` for viewing the position of a specific forecast


backend/packages/wps-api/src/app/routers/smurfi.py
```diff
@@ -370,9 +441,9 @@ async def create_spot_forecast_endpoint(
     payload = SpotUpdatePayload(
         spot_request_id=spot_forecast_input.spot_request_base_id, spot_forecast_id=spot_forecast_id
     )
-    await publish(
-        stream=stream_name, subject=smurfi_spot_update_subject, payload=payload, subjects=subjects
-    )
+    # await publish(
```
conbrad: Do we want to add this back in?

brettedw: Ah yep, definitely, thanks

brettedw: Done in [1db728a](https://github.com/bcgov/wps/pull/5470/commits/1db728a6fc1af0981019cbb1917ffb3db3e0a4ae)

conbrad (approved):

PR #5484, opened 2026-06-04
SMURFI: Status Change

brettedw:
- Allow owner and forecaster to change statuses (but not back to REQUESTED)
- Add error snackbar to the page for any request error


backend/packages/wps-api/src/app/routers/smurfi.py
```diff
@@ -512,6 +536,39 @@ def _spot_forecast_to_schema(spot_forecast: SpotForecast) -> SpotForecastData:
...
+
+    async with get_async_write_session_scope() as session:
+        spot_request = await get_spot_request_by_id(session, spot_request_id)
+        if spot_request is None:
+            raise HTTPException(
+                status_code=status.HTTP_404_NOT_FOUND,
+                detail=f"SpotRequestBase {spot_request_id} not found",
+            )
+
+        if not _can_update_spot_request_status(token, spot_request, next_status):
```
conbrad: We should just be able to create a dependency for this check in `auth.py` like the existing `auth_with_forecaster_role_required` but with the spot ownership check as well.

web/apps/wps-web/src/features/smurfi/components/SpotStatusControl.tsx
conbrad, line 1: Thanks for putting all this in a component!

web/apps/wps-web/src/features/smurfi/components/SpotStatusControl.tsx
```diff
@@ -0,0 +1,120 @@
...
+    setAnchorEl(null)
+  }
+
+  const handleStatusChange = async (status: SpotRequestStatus) => {
+    handleClose()
+    if (status === spotRequest.status) {
+      return
+    }
+
+    setIsUpdating(true)
```
conbrad: Usually we include request status in the slice, any reason why we separate that state here?

backend/packages/wps-api/src/app/routers/smurfi.py
brettedw: I think the auth check would need to query our db to confirm spot ownership, do you think that belongs in `auth.py`?

web/apps/wps-web/src/features/smurfi/components/SpotStatusControl.tsx
brettedw: Good call, done by id in [58e9ed0](https://github.com/bcgov/wps/pull/5484/commits/58e9ed044a342ebc9c3e95a5cfab8b2981fd1d24)

backend/packages/wps-api/src/app/routers/smurfi.py
conbrad: Yep, if spot ownership decides if someone is authorized it makes sense. We also interact with the db there already for auditing.

brettedw: Done in [c09e7b9](https://github.com/bcgov/wps/pull/5484/commits/c09e7b94901fdf846e24f12a7e4e8c5ce74fe56f)

brettedw: (with tests)

conbrad (approved):
😎

PR #5492, opened 2026-06-08
SMURFI Forecaster Info & Stats

brettedw:
- Adds forecaster and forecast by information to the dashboard and map popup
- Filter by forecaster on the dashboard
  - it would be great to filter by forecaster on the map page as well, but it's a bit awkward adding another search bar to the legend/layer switcher 
- add Stats button to dashboard to summarize spots
- improve dashboard layout
- cleanup unused code
- table header css closer to design -> uses our primary colour

**Added forecaster phone number**
 - Added a top forecaster info section to `SpotForecastForm` using new `SpotForecasterInfo.tsx`
  - Name/email are readonly from auth token details, phone is editable
  - Phone prefills from sourceForecast when:
      - using “edit forecast” behaviour
      - the source forecast has the same forecaster name and email
  - Added `forecasterPhone` to the form schema/defaults and marshalled it as `forecaster_phone`
  - Backend create forecast now saves and returns `forecaster_phone` instead of dropping it.
  - Forecast views now always show a Phone row with `—` when missing.


conbrad (approved):
🔥

PR #5513, opened 2026-06-16
Dev: Bump GDAL to 3.13.1 and overhaul local mac setup + docs

conbrad:
Bumps GDAL from `3.12.3` → `3.13.1` so the project tracks current `homebrew-core` (which removes
  the need for the custom pinned-tap workaround on macOS and unblocks using Homebrew
  Postgres/PostGIS locally), then overhauls `setup/mac.sh` and the setup docs to match, and adds an
  automated e2e check for the setup script.

  ### What changed

  **GDAL bump (the headline)**
  - `gdal==3.12.3` → `gdal==3.13.1` in `wps-api`, `wps-shared`, `wps-jobs` `pyproject.toml`s;
  `uv.lock` regenerated (`gdal` is the only resolution change).
  - Base image tags bumped to `ghcr.io/osgeo/gdal:ubuntu-{full,small}-3.13.1` in
  `openshift/wps-api-base`, `openshift/wps-jobs-base`, and
  `backend/packages/wps-weather/Dockerfile`.

  **`setup/mac.sh` overhaul**
  - Fixes the Homebrew bootstrap (it previously printed the installer instead of executing it) and
  puts `brew` on `PATH` for the rest of the run.
  - Adds `set -euo pipefail` and makes every step idempotent (safe to re-run).
  - GDAL: plain `brew install gdal` + `brew pin gdal`
  - Postgres: installs `postgresql@17` + `postgis` and provisions the `wps` DB/users (viable now
  that GDAL matches core).
  - Adds `eccodes` (required by weather-model tests); installs `wkhtmltopdf` from the packaging
  releases with Rosetta 2 on Apple Silicon (optional, non-fatal).
  - Java via `JAVA_HOME` instead of an interactive `sudo` symlink.

  **Documentation**
  - Root `README.md`: native setup is the recommended/supported path; Docker Compose and the dev
  container are marked **currently unsupported**; fixes broken env-file/dev-container paths; fills
  the empty Dependencies section.
  - `backend/README.md` / `backend/packages/wps-api/README.md`: correct the package list, env-file
  location, Docker-build cwd; replace obsolete poetry/`rpy2` troubleshooting with current notes.
  - `docs/MANUAL_SETUP.md`: rewritten from poetry-era to `uv` (removes the R/`cffdrs` section,
  fixes broken links/paths, invalid SQL, and nonexistent `Makefile` targets).
  - `.env.example`: corrected `CLASSPATH` path and docker-vs-native DB host guidance.

  **CI**
  - `.github/workflows/setup_mac_test.yml`: runs `setup/mac.sh` end-to-end on a clean arm64
  `macos-15` runner, then smoke-tests the result (`uv sync`, `from osgeo import gdal`). Triggers: manual and a weekly cron
  to catch external drift.

  ## Testing

  - `setup/mac.sh`: `shellcheck` clean; full stubbed dry-run passes under `set -euo pipefail` (exit
  `0`); idempotent DB guards verified against a live cluster.
  - GDAL `3.13.1` verified locally: `uv sync --all-extras` clean, `from osgeo import gdal` →
  `3.13.1`, full test run identical to the `3.12.3` baseline (no version-sensitive regressions).
# Test Links:
[Landing Page](https://wps-pr-5513-e1e498-dev.apps.silver.devops.gov.bc.ca/)
[MoreCast](https://wps-pr-5513-e1e498-dev.apps.silver.devops.gov.bc.ca/morecast)
[Percentile Calculator](https://wps-pr-5513-e1e498-dev.apps.silver.devops.gov.bc.ca/percentile-calculator)
[C-Haines](https://wps-pr-5513-e1e498-dev.apps.silver.devops.gov.bc.ca/c-haines)
[FireCalc](https://wps-pr-5513-e1e498-dev.apps.silver.devops.gov.bc.ca/fire-behaviour-calculator)
[FireCalc bookmark](https://wps-pr-5513-e1e498-dev.apps.silver.devops.gov.bc.ca/fire-behaviour-calculator?s=266&f=c5&c=NaN&w=20,s=286&f=c7&c=NaN&w=16,s=1055&f=c7&c=NaN&w=NaN,s=305&f=c7&c=NaN&w=NaN,s=344&f=c5&c=NaN&w=NaN,s=346&f=c7&c=NaN&w=NaN,s=328&f=c7&c=NaN&w=NaN,s=1399&f=c7&c=NaN&w=NaN,s=334&f=c7&c=NaN&w=NaN,s=1082&f=c3&c=NaN&w=NaN,s=388&f=c7&c=NaN&w=NaN,s=309&f=c7&c=NaN&w=16,s=306&f=c7&c=NaN&w=NaN,s=1029&f=c7&c=NaN&w=NaN,s=298&f=c7&c=NaN&w=NaN,s=836&f=c7&c=NaN&w=NaN,s=9999&f=c7&c=NaN&w=NaN)
[Auto Spatial Advisory (ASA)](https://wps-pr-5513-e1e498-dev.apps.silver.devops.gov.bc.ca/auto-spatial-advisory)
[HFI Calculator](https://wps-pr-5513-e1e498-dev.apps.silver.devops.gov.bc.ca/hfi-calculator)
[SFMS Insights](https://wps-pr-5513-e1e498-dev.apps.silver.devops.gov.bc.ca/insights)
[Fire Watch](https://wps-pr-5513-e1e498-dev.apps.silver.devops.gov.bc.ca/fire-watch)
[Weather Toolkit](https://wps-pr-5513-e1e498-dev.apps.silver.devops.gov.bc.ca/weather-toolkit)

### DIFF
--- a/.github/workflows/setup_mac_test.yml
+++ b/.github/workflows/setup_mac_test.yml
@@ -9,13 +9,6 @@ name: Setup script (macOS)
 
 on:
   workflow_dispatch:
-  pull_request:
-    paths:
-      - "setup/mac.sh"
-      - "backend/packages/*/pyproject.toml"
-      - "backend/uv.lock"
-      - ".env.example"
-      - ".github/workflows/setup_mac_test.yml"
   schedule:
     # Mondays 09:00 UTC -- catch external drift (homebrew-core gdal vs the pyproject
     # pin, the wkhtmltopdf release, osgeo base images) even when the repo hasn't changed.
--- a/.github/workflows/setup_mac_test.yml
+++ b/.github/workflows/setup_mac_test.yml
@@ -1,0 +1,82 @@
+name: Setup script (macOS)
+
+# Runs setup/mac.sh on a clean, disposable arm64 macOS runner to verify the local
+# backend setup works end-to-end (Homebrew deps, GDAL, Postgres/PostGIS, Redis, uv).
+# This is the e2e test for setup/mac.sh that can't be safely run on a dev machine.
+#
+# Note: GitHub runners ship with Homebrew/Python preinstalled, so the "fresh machine"
+# Homebrew bootstrap branch in mac.sh is skipped here (brew is already present).
+
+on:
+  workflow_dispatch:
+  pull_request:
+    paths:
+      - "setup/mac.sh"
+      - "backend/packages/*/pyproject.toml"
+      - "backend/uv.lock"
+      - ".env.example"
+      - ".github/workflows/setup_mac_test.yml"
+  schedule:
+    # Mondays 09:00 UTC -- catch external drift (homebrew-core gdal vs the pyproject
+    # pin, the wkhtmltopdf release, osgeo base images) even when the repo hasn't changed.
+    - cron: "0 9 * * 1"
+
+concurrency:
+  group: setup-mac-${{ github.ref }}
+  cancel-in-progress: true
+
+permissions:
+  contents: read
+
+jobs:
+  mac-setup:
+    name: Provision a clean macOS machine with setup/mac.sh
+    # Apple Silicon (arm64) -- matches dev machines and exercises the Rosetta/wkhtmltopdf path.
+    runs-on: macos-15
+    timeout-minutes: 75
+    env:
+      HOMEBREW_NO_INSTALL_CLEANUP: "1"
+    steps:
+      - uses: actions/checkout@v6
+
+      - name: Run setup/mac.sh
+        # Runners have passwordless sudo, so the wkhtmltopdf .pkg / Rosetta steps run
+        # non-interactively.
+        run: bash setup/mac.sh
+
+      - name: Put provisioned tools on PATH for the smoke test
+        # mac.sh exports these for its own run only; re-expose them for later steps.
+        run: |
+          echo "$(brew --prefix postgresql@17)/bin" >> "$GITHUB_PATH"
+          echo "JAVA_HOME=$(brew --prefix openjdk)/libexec/openjdk.jdk/Contents/Home" >> "$GITHUB_ENV"
+
+      - name: Smoke test -- tools installed
+        run: |
+          set -euo pipefail
+          gdal-config --version
+          psql --version
+          redis-cli ping
+          uv --version
+          command -v wkhtmltopdf >/dev/null && wkhtmltopdf --version || echo "wkhtmltopdf not installed (optional)"
+
+      - name: Smoke test -- database provisioned
+        run: |
+          set -euo pipefail
+          psql -d wps -c "select postgis_version();"
+
+      - name: Create .env for the backend
+        run: |
+          set -euo pipefail
+          cp .env.example .env
+          # native dev: DB on localhost; hostname-safe object-store placeholders
+          sed -i '' -e 's/^POSTGRES_WRITE_HOST=db/POSTGRES_WRITE_HOST=localhost/' \
+                    -e 's/^POSTGRES_READ_HOST=db/POSTGRES_READ_HOST=localhost/' \
+                    -e 's/=object_store_server/=object-store-server/' \
+                    -e 's/=wx_object_store_server/=wx-object-store-server/' .env
+
+      - name: Smoke test -- uv sync with GDAL binding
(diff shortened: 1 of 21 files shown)


setup/mac.sh
```diff
@@ -1,55 +1,144 @@
...
-curl -s https://raw.githubusercontent.com/Homebrew/homebrew-core/2761c8e5f5547753c8bebc39e95968006f5deb69/Formula/g/gdal.rb > $(brew --repository)/Library/Taps/$(whoami)/homebrew-local-gdal/Formula/gdal.rb
-brew install $(whoami)/local-gdal/gdal # if you have gdal/postgis already installed you'll have to uninstall them both, see MANUAL.md for more details
+# The python gdal binding (pinned in backend/packages/*/pyproject.toml) builds from source
+# against the installed libgdal, which must be >= the pinned version. The pin tracks the
+# ghcr.io/osgeo/gdal base image tag (see openshift/wps-api-base/docker/Dockerfile); keep
+# homebrew gdal, the base image tag, and the pyproject pins in step when bumping.
+brew install gdal
+# Prevent `brew upgrade` from moving the libgdal soname out from under backend/.venv.
+# When the project bumps its gdal pin: brew unpin gdal && brew upgrade gdal && uv sync.
+brew list --pinned | grep -qx gdal || brew pin gdal
```
brettedw: I've gone down this road before trying to pin gdal. I believe it only pins gdal, but not it's dependencies. So when brew updates other packages, it broke gdal for me

conbrad: Hmm good point, I think I'll unpin and rely on the job to catch failures so we're notified, not sure there's any other thing we can without jumping into devcontainers.

conbrad: https://github.com/bcgov/wps/pull/5513/commits/84f00a46f3c5444a409d9ff02f3072e5642a0f4e

setup/mac.sh
```diff
@@ -1,55 +1,147 @@
...
-# grant connect on database wps to wpsread; grant usage on schema public to wpsread; grant select on all tables in schema public to wpsread;
-# "
-# echo "finished installing and configuration postgres, run migrations in poetry shell"
+### postgres + postgis
+# Homebrew's postgis depends on the current homebrew-core gdal, which now matches the
+# project's pinned gdal (see the gdal note above), so brew postgis no longer collides
+# with it -- the Nov 2024 reason for avoiding brew postgres is resolved.
+# postgis ships bottles for postgresql@17 and @18; we use @17.
+echo "installing and configuring postgres"
+brew install postgresql@17 postgis
```
brettedw, line 85: We're still running postgres 16 in deployments, not sure we want to update this to 17 yet

conbrad: `brew` doesn't have formulas for 16 or below anymore, we're stuck with 17 as a minimum.

brettedw: https://formulae.brew.sh/formula/postgresql@16

Looks like it should be available for several more years?

conbrad: I should have said we have no formula combo for pg16/gdal/postgis anymore. There is no bottled PostGIS for `postgresql@16`.

brettedw: I think that's why we were using the postgres app, pg 16 and postgis didn't work well when installed with brew. I guess it's not a big deal, the functionality shouldn't change that much, but installing with the postgres app also handles installing it's own version of compatible postgis/gdal I believe, which seems to work well

conbrad: I prefer sticking with one build source for now, we don't use any breaking features between pg16 and pg17.

.github/workflows/setup_mac_test.yml
brettedw, line 1: This is cool, awesome to know when setup breaks. I'm just wondering if it's worth the ~7min action runtime on pull requests. Do you think it would be useful to just keep it as a workflow_dispatch and on the schedule?

conbrad: https://github.com/bcgov/wps/pull/5513/commits/2b5e7cf053ad109c4e8b18731d96aed367b4634e

brettedw (approved):
Too scared to try most of this right now but it looks great, thanks for the updates!

PR #5573, opened 2026-06-29
SFMS: Provide Daily FWI endpoint

### LINKED ISSUE #5502: SFMS: Provide Daily FWI End Point
**Describe the task**

Endpoint for downloading FFMC, DMC, DC, Hourly FFMC, ISI , FWI, BUI (all) actuals rasters.

**Acceptance Criteria**
- [ ] Put endpoint behind API gateway so users can register for a token
- [ ] Endpoint should allow requesting FFMC, DMC, DC, Hourly FFMC, ISI , FWI, BUI (all) actuals rasters, reusing the object store proxy method if it suffices.
- [ ] Another endpoint for querying underlying raster value by lat/long

**Additional context**
- Users need to be able to download the whole raster at a minimum

conbrad:
Adds `/api/sfms/daily-fwi` endpoints to download SFMS daily FWI actuals rasters (FFMC, DMC, DC, ISI, FWI, BUI) and hourly FFMC rasters, and to query raster values by lat/lon.

## Endpoints

| Method | Path | Description |
|--------|------|-------------|
| `GET` | `/api/sfms/daily-fwi/{date}/{parameter}` | Download daily FWI actuals raster (dc, dmc, bui, ffmc, isi, fwi) |
| `GET` | `/api/sfms/daily-fwi/{date}/hffmc?hour=` | Download hourly FFMC actuals raster |
| `GET` | `/api/sfms/daily-fwi/{date}/{parameter}/value?lat=&lon=` | Query raster value at a WGS84 lat/lon point |
| `GET` | `/api/sfms/daily-fwi/{date}/hffmc/value?hour=&lat=&lon=` | Query hourly FFMC value at a lat/lon point |

Endpoints are open at the Python level and protected via the BC Government APS gateway (Kong `key-auth` plugin, 60 req/min rate limit per consumer) consumers register for an API key through the APS portal rather than using a Keycloak IDIR token. Public URL: `psu.api.gov.bc.ca/api/sfms/daily-fwi/...`.

The backing rasters served here are those produced by the existing SFMS system run by geospatial (uploaded to `sfms/uploads/*`), not either of WPS's own in-house calculation pipelines (`sfms/calculated/*`, `sfms_ng/*`).

#### Standalone service (not publicly exposed)

The daily FWI routes are served by a dedicated `wps-sfms-fwi-api` deployment rather than the main API:

- `app/sfms_fwi_main.py` — separate FastAPI entrypoint including only the `sfms_fwi` router; the shared API image selects it via the `APP_MODULE` env var. The main API no longer serves these routes (guarded by a test).
- `openshift/templates/sfms_fwi_api.yaml` — Deployment + ClusterIP Service with **no Route**; the only ingress is a NetworkPolicy allowing the APS gateway namespace (`allow_gateway_to_wps_sfms_fwi_api.yaml`). Uses the same Crunchy DB credentials as the other API pods so startup migrations work.
- Reuses the existing `object_store_proxy._proxy` method for raster strea

#### CI / gateway publishing

- ASA Go and SFMS FWI deploy as separate, parallel dev jobs; a shared `publish-aps-gateway-dev` job (depending on both) publishes gateway config, so Kong never routes to a not-yet-deployed upstream.
- `aps-publish` composite action consolidates render + publish + upload fzed to work for both ASA Go and SFMS FWI. Gateway configs publish underindependent qualifiers (`pr-N`/`prod` for ASA Go, `fwi-pr-N`/`fwi-prod` for SFMS FWI) and are pruned on PR cleanup.
- New `gwa-setup` action installs the `gwa` CLI once per job (previously up to 3x per job across separate steps).
- **PR/dev builds now publish entirely to a separate APS test tenant** — gateway `gw-11008`, host `api-gov-bc-ca.test.api.gov.bc.ca` — for both gateway config and the SFMS FWI dataset/product listing. Only real production releases (`production.yml`) publish to the production tenant (`gw-313f6`). `host` is now a required input on all four `gwa-*` actions.

See `openshift/aps/README.md` for the full gateway/tenant/credential breakdown.

Closes #5502
# Test Links:
[Landing Page](https://wps-pr-5573-e1e498-dev.apps.silver.devops.gov.bc.ca/)
[MoreCast](https://wps-pr-5573-e1e498-dev.apps.silver.devops.gov.bc.ca/morecast)
[Percentile Calculator](https://wps-pr-5573-e1e498-dev.apps.silver.devops.gov.bc.ca/percentile-calculator)
[C-Haines](https://wps-pr-5573-e1e498-dev.apps.silver.devops.gov.bc.ca/c-haines)
[FireCalc](https://wps-pr-5573-e1e498-dev.apps.silver.devops.gov.bc.ca/fire-behaviour-calculator)
[FireCalc bookmark](https://wps-pr-5573-e1e498-dev.apps.silver.devops.gov.bc.ca/fire-behaviour-calculator?s=266&f=c5&c=NaN&w=20,s=286&f=c7&c=NaN&w=16,s=1055&f=c7&c=NaN&w=NaN,s=305&f=c7&c=NaN&w=NaN,s=344&f=c5&c=NaN&w=NaN,s=346&f=c7&c=NaN&w=NaN,s=328&f=c7&c=NaN&w=NaN,s=1399&f=c7&c=NaN&w=NaN,s=334&f=c7&c=NaN&w=NaN,s=1082&f=c3&c=NaN&w=NaN,s=388&f=c7&c=NaN&w=NaN,s=309&f=c7&c=NaN&w=16,s=306&f=c7&c=NaN&w=NaN,s=1029&f=c7&c=NaN&w=NaN,s=298&f=c7&c=NaN&w=NaN,s=836&f=c7&c=NaN&w=NaN,s=9999&f=c7&c=NaN&w=NaN)
[Auto Spatial Advisory (ASA)](https://wps-pr-5573-e1e498-dev.apps.silver.devops.gov.bc.ca/auto-spatial-advisory)
[HFI Calculator](https://wps-pr-5573-e1e498-dev.apps.silver.devops.gov.bc.ca/hfi-calculator)
[SFMS Insights](https://wps-pr-5573-e1e498-dev.apps.silver.devops.gov.bc.ca/insights)
[Fire Watch](https://wps-pr-5573-e1e498-dev.apps.silver.devops.gov.bc.ca/fire-watch)
[Weather Toolkit](https://wps-pr-5573-e1e498-dev.apps.silver.devops.gov.bc.ca/weather-toolkit)


backend/packages/wps-api/src/app/sfms_fwi_main.py
```diff
@@ -0,0 +1,106 @@
...
+
+    This service only reads from the object store — it has no database
+    dependency, so being up is being healthy."""
+    return {"healthy": True, "message": "ok"}
+
+
+if __name__ == "__main__":
+    import uvicorn
+
+    uvicorn.run(app, host="0.0.0.0", port=8080)
```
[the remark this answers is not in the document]
conbrad: Container env

.github/actions/aps-publish/action.yml
brettedw, line 1: Why the rename of `aps-publish` to `gw-config-publish`? I find the new naming a bit confusing, being so similar to `gwa-publish`.  In my mind `gwa-publish` installs/runs gwa, while `aps-publish` tied it together for the higher level APS flow. Could we keep the `aps-publish` name and just generalize the inputs instead?

.github/actions/sfms-fwi-aps-apply/action.yml
```diff
@@ -0,0 +1,52 @@
...
+      run: |
+        set -euo pipefail
+        OUTPUT_DIR=".github/tmp/sfms-fwi-aps-resources-${SUFFIX}"
+        mkdir -p "${OUTPUT_DIR}"
+        bash "${{ github.workspace }}/openshift/scripts/render_sfms_fwi_aps_resources.sh" \
+          "${SUFFIX}" \
+          "${OUTPUT_DIR}"
+
+    - name: Apply SFMS FWI dataset and product
+      uses: ./.github/actions/gwa-apply
```
brettedw, line 37: The `gwa-apply` custom action is only used once here, do you think it's worth having a custom action for? 
Side note: I think in our workflows we're now installing the gwa cli 3 times. Maybe we should just install it once as an explicit step?

backend/packages/wps-api/src/app/routers/sfms_fwi.py
```diff
@@ -0,0 +1,155 @@
...
+@router.get("/{for_date}/{parameter}", responses=_RASTER_RESPONSES)
+async def get_daily_fwi_raster(
+    request: Request,
+    for_date: Annotated[date, Path(description=_FOR_DATE_DESC)],
+    parameter: Annotated[
+        FWIParameter, Path(description="FWI parameter: dc, dmc, bui, ffmc, isi, fwi")
+    ],
+):
+    """Download the daily FWI actuals raster for the given date and parameter."""
+    key = _addresser.get_calculated_index_key(datetime(for_date.year, for_date.month, for_date.day, tzinfo=timezone.utc), parameter, RunType.ACTUAL)
```
brettedw: These endpoints are getting the incorrect data, they need to be getting data from the current SFMS

brettedw (changes requested):

backend/packages/wps-shared/src/wps_shared/geospatial/wps_dataset.py
```diff
@@ -434,6 +435,35 @@ def get_nodata_mask(self) -> Tuple[Optional[np.ndarray], Optional[Union[float, i
...
+        src_srs.SetAxisMappingStrategy(osr.OAMS_TRADITIONAL_GIS_ORDER)
+
+        wgs84 = osr.SpatialReference()
+        wgs84.ImportFromEPSG(4326)
+        wgs84.SetAxisMappingStrategy(osr.OAMS_TRADITIONAL_GIS_ORDER)
+
+        x, y, _ = osr.CoordinateTransformation(wgs84, src_srs).TransformPoint(lon, lat)
+
+        col = int((x - geotransform[0]) / geotransform[1])
+        row = int((y - geotransform[3]) / geotransform[5])
```
brettedw:
```suggestion
      inverse_geotransform = gdal.InvGeoTransform(geotransform)
      pixel_x, pixel_y = gdal.ApplyGeoTransform(inverse_geotransform, x, y)

      col = math.floor(pixel_x)
      row = math.floor(pixel_y)
```

brettedw: We might want to do something like to avoid returning values from pixels that are just fractionally outside the raster

.github/workflows/deployment.yml
```diff
@@ -183,6 +162,93 @@ jobs:
...
+        with:
+          config-name: sfms-fwi
+          render-script: openshift/scripts/render_sfms_fwi_gateway_config.sh
+          gateway-id: ${{ env.APS_GATEWAY_ID }}
+          client-id: ${{ secrets.GWA_ACCT_ID }}
+          client-secret: ${{ secrets.GWA_ACCT_SECRET }}
+          suffix: ${{ env.SUFFIX }}
+          qualifier-prefix: fwi-
+
+      - name: Publish SFMS FWI dataset and product to test API Directory
```
brettedw, line 247: Useful for testing to see the dataset/product this time for sure, but do you think we want this regularly? I think this will list every PR we have as a dataset and product in the API Directory

backend/packages/wps-api/src/app/routers/sfms_fwi.py
brettedw, line 1: It might be worth adding something to `launch.json` so we can easily run the SFMS api locally

.github/workflows/deployment.yml
conbrad: I'd agree if this published to the production API directory but this publishes to the test API Directory. Then it's cleaned up in `cleanup.yml` so I think it's worth keeping to exercise the complete deployment path.

brettedw: Ah ya, I see the intent. It's currently not being published to the test directory though. It looks like the test host needs to be passed to the action. It's currently listed in the production directory here: https://api.gov.bc.ca/devportal/api-directory

backend/packages/wps-shared/src/wps_shared/geospatial/wps_dataset.py
conbrad: Nice catch, done here: https://github.com/bcgov/wps/pull/5573/commits/a528c3bbb725d41d3ee35d036cb07760d66a4888

backend/packages/wps-api/src/app/routers/sfms_fwi.py
conbrad: Pointed to SFMSng here: https://github.com/bcgov/wps/pull/5573/commits/4f17d5fd352caf3098530e5033f68a148663d16a

brettedw: Sorry, by current SFMS I meant the existing SFMS being run by geospatial. That was my understanding of the task anyway

brettedw: We were already providing the hourlies from that SFMS

.github/actions/aps-publish/action.yml
conbrad: Yep, went with your suggestion and kept it as `aps-publish`, just generalized the inputs (config-name, render-script, qualifier-prefix) so it works for both ASA Go and SFMS FWI. It still calls `gwa-publish` internally same as before, but the two names don't share a prefix anymore, so they shouldn't read as near-twins in the actions folder.

https://github.com/bcgov/wps/pull/5573/changes/a833b76171233c6b8da19d978a153702a73cc5ad

.github/actions/sfms-fwi-aps-apply/action.yml
conbrad: Yeah install action is a good call, split that out into a `gwa-setup` action and call it once per job now, so we're down to 1 install instead of 3.

On keeping `gwa-apply` as its own action: I'd lean toward keeping it. Now that CLI setup is factored out, `gwa-apply` and `gwa-publish` are both just "authenticate, then run one gwa subcommand", same shape, same abstraction level. `gwa-publish` already has 3 callers, so if we ever register another dataset/product with the API Directory (which seems likely if we onboard more gateway-fronted APIs the way we're doing for SFMS FWI), `gwa-apply` picks that up for free. Folding it into `sfms-fwi-aps-apply` wouldn't save much today, and we'd lose that symmetry.

Not a strong objection either way though, happy to inline it if you'd still rather not carry the extra file for a single caller.

https://github.com/bcgov/wps/pull/5573/commits/4b73738c4bc6d22557d73b098f1ec8ac9fc67ab8

backend/packages/wps-api/src/app/routers/sfms_fwi.py
conbrad: Done here: https://github.com/bcgov/wps/pull/5573/commits/61f2610d13b6ee4de765d3203d9584dc3b15f7dd

backend/packages/wps-api/src/app/routers/sfms_fwi.py
conbrad: Hmm I see, yeah it's a bit ambiguous and we don't have great agreed upon naming for old/new SFMS. Perhaps we should add some documentation to solidify either {legacy, current, old} and {new, ng} etc.

I'll update.

.github/workflows/deployment.yml
conbrad: That's no good, missed a string, it should match intent now: https://github.com/bcgov/wps/pull/5573/commits/188fa5b83cb30ff518aca417da135c59c62d9eae

backend/packages/wps-api/src/app/routers/sfms_fwi.py
conbrad: Updated and added more doc comments

.github/actions/gwa-apply/action.yml
```diff
@@ -0,0 +1,45 @@
+name: GWA Apply
+description: Authenticate and apply APS resources (DraftDataset, Product, etc.) with gwa. Requires ./.github/actions/gwa-setup to have run earlier in the job.
```
brettedw:
```suggestion
description: Apply APS resources (DraftDataset, Product, etc.) with gwa. Requires ./.github/actions/gwa-setup to have run earlier in the job.
```

openshift/aps/sfms-fwi-dataset.yaml
```diff
@@ -0,0 +1,24 @@
...
+notes: |
+  The SFMS Daily Fire Weather Index (FWI) API provides daily fire weather index values
+  calculated by the Wildfire Predictive Services Unit (PSU) using the Canadian Forest
+  Fire Weather Index System (CFFDRS).
+
+  FWI values represent the fire danger rating for a given day and location, derived from
+  temperature, relative humidity, wind speed, and 24-hour precipitation observations at
+  BC weather stations.
+
+  Visit the [WPS API specification](https://psu.nrs.gov.bc.ca/api/docs) for endpoint
```
brettedw: I think we may need another way to document these endpoints now, I don't think they'll be included here anymore: https://wps-pr-5573-e1e498-dev.apps.silver.devops.gov.bc.ca/api/docs

.github/actions/gwa-publish/action.yml
```diff
@@ -1,5 +1,5 @@
 name: GWA Publish
-description: Install gwa, authenticate, and publish APS gateway configuration
+description: Authenticate and publish APS gateway configuration with gwa. Requires ./.github/actions/gwa-setup to have run earlier in the job.
```
brettedw:
```suggestion
description: Publish APS gateway configuration with gwa. Requires ./.github/actions/gwa-setup to have run earlier in the job.
```

openshift/aps/sfms-fwi-dataset.yaml
conbrad: Good catch, you're right: moving these to the standalone service means the main API's `/api/docs` doesn't include them anymore. Added a separate, unauthenticated route config (`/api/docs`, `/api/openapi.json`, `/api/redoc`) alongside the key-authed data route on the same gateway service, and pointed the dataset notes there instead: https://psu-pr-5573-api-gov-bc-ca.test.api.gov.bc.ca/api/docs

Still rate-limited even without a key, just IP-based instead of consumer-based (60/min per IP) since there's no API key on that route to key off of, so it can't be hammered anonymously.

.github/actions/aps-publish/action.yml
```diff
@@ -11,45 +17,60 @@ inputs:
...
       run: |
         set -euo pipefail
-        CONFIG_PATH=".github/tmp/asa-go-gw-config-${SUFFIX}.yaml"
+        # Prod uses the bare vanity host and prod namespace; PRs are suffixed and live in
+        # dev. Do NOT dash/test-ify this host manually: the APS platform itself remaps
+        # whatever prod-style host is given here to the test tenant's real public domain
+        # once gwa is configured against the test gateway (host/--gateway set by the
+        # caller). Applying that transform ourselves double-applies it, since gwa applies
+        # its own remap on top -- confirmed against a real CI run's `gwa publish-gateway`
+        # diff output publishing a doubled host.
```
brettedw: Could we simplify this comment? What is "dash/test-ify"?

conbrad: gwa dashes test urls automatically: https://teams.microsoft.com/l/message/19:c81bf553c07647cebeb2aeb034ec0d25@thread.tacv2/1783372703689?tenantId=6fdb5200-3d0d-4a8a-b036-d3685e359adc&groupId=a80418da-c27b-406e-89ab-7695b61924d8&parentMessageId=1782763600762&teamName=Developer%20Community&channelName=API-ProgramServices-operations&createdTime=1783372703689
we gotta adjust urls to that. Not sure of other language to use will happily take suggestions.

conbrad: attempt here: https://github.com/bcgov/wps/pull/5573/commits/f96bdc12506162824a785aa35173e7cc874fdede

brettedw: that makes sense, I just confused about what that term meant and where the adjustment was happening

backend/packages/wps-api/src/app/routers/sfms_fwi.py
```diff
@@ -0,0 +1,258 @@
...
+_addresser = RasterKeyAddresser()
+
+_FOR_DATE_DESC = "Date of the raster (YYYY-MM-DD)"
+_FOR_DATE_EXAMPLE = "2025-11-02"
+_PARAMETER_DESC = "FWI parameter: dc, dmc, bui, ffmc, isi, fwi"
+_LAT_DESC = "Latitude in WGS84"
+_LON_DESC = "Longitude in WGS84"
+_LAT_EXAMPLE = 49.0
+_LON_EXAMPLE = -123.0
+_HOUR_DESC = "Hour of the raster (0-23, PST/PDT -- matches the SFMS upload filename convention)"
```
brettedw: From our docs it suggests SFMS runs in local time, so going forward it will always run in PDT. Do you think that's worth changing these docs?

brettedw (approved):
that was a big one, nice work!!

backend/packages/wps-api/src/app/routers/sfms_fwi.py
conbrad: Yep good idea, done in: https://github.com/bcgov/wps/pull/5573/commits/bfa91ac26be5689e71835e63671a3b8048417023

Issue #5478, opened 2026-06-03
Dev: PSU Landing Page updates


NicolasLivanos:
**Describe the task**

PSU Landing Page update to indicate which apps require login, which are public access, and a favourites option

**Acceptance Criteria**
- [x] Follow design for app section
- [x] Pinning a favourite
- [x] Public vs Internal apps 
- [x] Add a Mobile Footer 
- [x] Remove the left-hand menu from the home page.
- [x] Add the API link: https://api.gov.bc.ca/devportal/api-directory/4336 
- [x] Add  Sam’s app (supported by the PSU): https://huggingface.co/spaces/ssidds/WX_data_viewer 
- [x] Move collaborate with us links to the descriptive text at the top of the page, i.e.) Teams Sprint Review and Collaboard links.
- [x] Support section edit @nealmcloughlin request: " To report bugs or receive support on technical issues, please use the Submit Feedback in-app functionality; for after-hours support please email: BCWS.TechServices@gov.bc.ca "
- [x] `Managed By:` is `{CSBC | BCWS} - Predictive Services` with link to either the shared email for BCWS or sentry feedback for CSBC. 
- [x] Uses ASA Go icon for ASA

**Additional context**
- Design: https://github.com/bcgov/wps/issues/4466
- Figma Make in a zip file below

rajpersram:
Updated design and content found at:
https://www.figma.com/make/63iP6po6v8Zx1kpHV5KyFp/Decision-Support-Tools-Revamp?code-node-id=0-9&fullscreen=1

Note: there still needs to be a way to access the shortened menu from each app, so we likely need to retain the hamburger menu and have it show in each app at the top left in the blue BC Gov header bar.

rajpersram:
<img width="1133" height="1185" alt="Image" src="https://github.com/user-attachments/assets/5f292e05-9a73-4cce-a9d6-239ddfcbbb15" />

rajpersram:
[Decision Support Tools Revamp.zip](https://github.com/user-attachments/files/29523867/Decision.Support.Tools.Revamp.zip)

brettedw:
[...]

PR #5632, opened 2026-07-21
Landing Page Revamp

brettedw:
A complete overhaul of the PSU Landing Page

Based on designs in #5478 

# Test Links:
[Landing Page](https://wps-pr-5632-e1e498-dev.apps.silver.devops.gov.bc.ca/)
[MoreCast](https://wps-pr-5632-e1e498-dev.apps.silver.devops.gov.bc.ca/morecast)
[Percentile Calculator](https://wps-pr-5632-e1e498-dev.apps.silver.devops.gov.bc.ca/percentile-calculator)
[FireCalc](https://wps-pr-5632-e1e498-dev.apps.silver.devops.gov.bc.ca/fire-behaviour-calculator)
[FireCalc bookmark](https://wps-pr-5632-e1e498-dev.apps.silver.devops.gov.bc.ca/fire-behaviour-calculator?s=266&f=c5&c=NaN&w=20,s=286&f=c7&c=NaN&w=16,s=1055&f=c7&c=NaN&w=NaN,s=305&f=c7&c=NaN&w=NaN,s=344&f=c5&c=NaN&w=NaN,s=346&f=c7&c=NaN&w=NaN,s=328&f=c7&c=NaN&w=NaN,s=1399&f=c7&c=NaN&w=NaN,s=334&f=c7&c=NaN&w=NaN,s=1082&f=c3&c=NaN&w=NaN,s=388&f=c7&c=NaN&w=NaN,s=309&f=c7&c=NaN&w=16,s=306&f=c7&c=NaN&w=NaN,s=1029&f=c7&c=NaN&w=NaN,s=298&f=c7&c=NaN&w=NaN,s=836&f=c7&c=NaN&w=NaN,s=9999&f=c7&c=NaN&w=NaN)
[Auto Spatial Advisory (ASA)](https://wps-pr-5632-e1e498-dev.apps.silver.devops.gov.bc.ca/auto-spatial-advisory)
[HFI Calculator](https://wps-pr-5632-e1e498-dev.apps.silver.devops.gov.bc.ca/hfi-calculator)
[SFMS Insights](https://wps-pr-5632-e1e498-dev.apps.silver.devops.gov.bc.ca/insights)
[Fire Watch](https://wps-pr-5632-e1e498-dev.apps.silver.devops.gov.bc.ca/fire-watch)
[Weather Toolkit](https://wps-pr-5632-e1e498-dev.apps.silver.devops.gov.bc.ca/weather-toolkit)


web/apps/wps-web/src/features/landingPage/components/ToolRow.tsx
```diff
@@ -0,0 +1,128 @@
...
+              component="h3"
+              sx={{
+                fontWeight: 700,
+                lineHeight: 1.35,
+                pt: 1,
+                pb: { sm: 1, md: 3 }
+              }}
+              variant="body1"
+            >
+              <Link
```
conbrad: These render every tool as a plain anchor (`href={tool.route}, no component={RouterLink}`), including internal SPA routes like `/hfi-calculator` and /morecast. The old `ToolCard.tsx` deliberately branched on route.startsWith('http') to use react-router's Link for internal routes and a plain <a> only for external ones.

As it stands, clicking any internal tool (title link or "Open" button) forces a full page reload of the whole app instead of a client-side route change. Can we bring back the internal/external branch: `component={RouterLink} to={tool.route}` for internal, `href+target="_blank"+rel="noreferrer"` for `isExternal` tools? `Footer.tsx` already does this same internal/external split via `StyledLink`/`StyledA`, so there's a precedent to follow.

web/apps/wps-web/src/features/landingPage/components/ToolRow.tsx
```diff
@@ -0,0 +1,128 @@
...
+            color: 'black',
+            fontSize: { xs: '0.8125rem', sm: '0.875rem', md: '0.9rem' },
+            lineHeight: 1.55
+          }
+        }}
+      >
+        {tool.description}
+      </Box>
+
+      <Button
```
conbrad: Similar to https://github.com/bcgov/wps/pull/5632/changes#r3640037703

web/apps/wps-web/src/features/landingPage/components/QuickAccessList.tsx
```diff
@@ -0,0 +1,68 @@
...
+            key={`${title}-${tool.route}`}
+            secondaryAction={
+              <FavouriteButton
+                isFavourite={favouriteRoutes.includes(tool.route)}
+                onToggle={() => onToggleFavourite(tool.route)}
+                toolName={tool.name}
+              />
+            }
+          >
+            <ListItemButton
```
conbrad: Similar to: https://github.com/bcgov/wps/pull/5632/changes#r3640037703

web/apps/wps-web/src/features/landingPage/components/ToolRow.tsx
conbrad: Perhaps the `tool` object can contain it's own `internal` or `external` context to make this easier.

web/apps/wps-web/src/features/landingPage/components/LandingPageHeader.tsx
```diff
@@ -0,0 +1,70 @@
...
+          aria-label="Open quick access"
+          onClick={onOpenQuickAccess}
+          startIcon={<MenuIcon />}
+          sx={{ alignSelf: { xs: 'stretch', sm: 'center' } }}
+          variant="outlined"
+        >
+          Quick Access
+        </Button>
+      </Tooltip>
+      <Box
```
conbrad: On main, the same logo (in the now-deleted Sidebar.tsx) was wrapped in `<a href=\"https://www2.gov.bc.ca/gov/content/safety/wildfire-status\" target=\"_blank\">`. A user clicking the logo expecting to reach the public Wildfire Status page doesn't have access through this path anymore, not sure how important that is though.

web/apps/wps-web/src/features/landingPage/landingPageTools.ts
```diff
@@ -0,0 +1,7 @@
+import { externalToolInfos, wxDataViewerInfo, wxWeatherAlertsInfo } from './ExternalToolInfos'
+import { fbpGoInfo, percentileCalcInfo, toolInfos, weatherToolkitInfo } from './toolInfo'
+
+export const publicTools = [fbpGoInfo, percentileCalcInfo, weatherToolkitInfo, wxDataViewerInfo, wxWeatherAlertsInfo]
+export const landingPageTools = [...toolInfos, ...externalToolInfos]
+
+export const bcwsTools = toolInfos.filter(tool => !publicTools.includes(tool))
```
conbrad:
```ts
export const bcwsTools = toolInfos.filter(tool => !publicTools.includes(tool))
```

This decides "BCPS Access Only" by exclusion where a tool is BCWS-only if it's *not* also referenced by identity in the separate `publicTools` array above. `ToolInfo` already has an explicit `isExternal?: boolean` as precedent for this kind of thing, but access-tier itself isn't a field on the tool at all, it's implicit in which of two arrays happens to include the object.

Concretely if we add a new tool to `toolInfos` in `toolInfo.tsx` and forget to also add it to `publicTools` here, and it silently lands in "BCPS Access Only" with no compiler or lint signal — since `access` isn't a declared property, nothing forces you to make the choice.

Could we make this an explicit field instead, e.g. `access: 'public' | 'bcws'` on `ToolInfo`, set per-tool in `toolInfo.tsx`/`ExternalToolInfos.tsx`? Then this file just partitions on the field:

```ts
export const publicTools = landingPageTools.filter(tool => tool.access === 'public')
export const bcwsTools = landingPageTools.filter(tool => tool.access === 'bcws')
```

That makes the categorization a required, type-checked property on every tool definition rather than something the compiler can't catch if it's missed.

web/apps/wps-web/src/features/landingPage/components/ToolRow.tsx
```diff
@@ -0,0 +1,128 @@
...
+
+interface ToolRowProps {
+  isFavourite: boolean
+  onToggleFavourite: () => void
+  tool: ToolInfo
+}
+
+const ToolRow = ({ isFavourite, onToggleFavourite, tool }: ToolRowProps) => {
+  const isExternal = tool.isExternal === true
+  const openFeedbackForm = async (event: React.MouseEvent<HTMLAnchorElement>) => {
```
conbrad:
```ts
const openFeedbackForm = async (event: React.MouseEvent<HTMLAnchorElement>) => {
  event.preventDefault()
  const feedback = getFeedback()
  if (!feedback) {
    return
  }

  const form = await feedback.createForm()
  form.appendToDom()
  form.open()
}
```

This is the same `getFeedback()` -> `createForm()` -> `appendToDom()` -> `open()` sequence already wrapped by `web/packages/ui/src/FeedbackButton.tsx` (used elsewhere via `MenuHeader.tsx`). If Sentry's feedback API ever changes a new required option on `createForm`, or the null-check for a missing integration needs adjusting, a fix applied to `FeedbackButton.tsx` will miss this second, independent copy, and the landing page's "Managed by" feedback link will quietly break while the header button keeps working.

We can't drop `<FeedbackButton />` in directly here since it renders its own "Submit Feedback" button, not a `Link` reading "CSBC - Predictive Services", but could we pull just the open-form action out into a small shared helper (e.g. `web/packages/ui/src/openFeedbackForm.ts`) that both `FeedbackButton.tsx` and `ToolRow.tsx` call?

```ts
// web/packages/ui/src/openFeedbackForm.ts
import { getFeedback } from '@sentry/react'

export const openFeedbackForm = async () => {
  const feedback = getFeedback()
  if (!feedback) {
    return
  }

  const form = await feedback.createForm()
  form.appendToDom()
  form.open()
}
```

Then `ToolRow.tsx`'s handler becomes:

```ts
onClick={tool.managedBy.opensFeedback ? (e) => { edbackForm() } : undefined}
```

That keeps one source of truth for opening sentry feedback forms while leaving each component free to render it however it needs to.

web/apps/wps-web/src/features/landingPage/components/Sidebar.tsx
```diff
@@ -1,250 +0,0 @@
...
-import SvgIcon from '@mui/material/SvgIcon'
-import { styled, useTheme } from '@mui/material/styles'
-import Tooltip from '@mui/material/Tooltip'
-import Typography from '@mui/material/Typography'
-import useMediaQuery from '@mui/material/useMediaQuery'
-import { LANDING_BACKGROUND_COLOUR, theme } from '@wps/ui/theme'
-import { MS_TEAMS_SPRINT_REVIEW_URL, SPRINT_REVIEW_BOARD_URL } from '@wps/utils/env'
-import SidebarToolList from 'features/landingPage/components/SidebarToolList'
-import Subheading from 'features/landingPage/components/Subheading'
-import CollaboardIcon from 'features/landingPage/images/collaboardIcon.svg?react'
```
conbrad, line 20: Can we delete these svgs now that they're unused?

web/apps/wps-web/src/features/landingPage/pages/LandingPage.tsx
```diff
@@ -1,50 +1,110 @@
...
+  const [isDrawerOpen, setIsDrawerOpen] = useState(false)
+  const favouriteTools = landingPageTools.filter(tool => favouriteRoutes.includes(tool.route))
+  const visibleBcwsTools = bcwsTools.filter(tool => !favouriteRoutes.includes(tool.route))
+  const visiblePublicTools = publicTools.filter(tool => !favouriteRoutes.includes(tool.route))
 
   useEffect(() => {
     document.title = LANDING_PAGE_DOC_TITLE
   }, [])
 
+  useEffect(() => {
```
conbrad:
```ts
const [favouriteRoutes, setFavouriteRoutes] = useState<string[]>(readFavouriteRoutes)
...
useEffect(() => {
  storeFavouriteRoutes(favouriteRoutes)
}, [favouriteRoutes])
```

`favouriteRoutes` is initialized from `readFavouriteRoutes()`, so on the very first mount this effect fires and writes the exact same value straight back to `localStorage`, a synchronous `setItem` on every landing-page visit before the user has touched anything. Harmless given the payload size, but it's a write-on-read pattern that would be worth avoiding if it's ever copied to a heavier key or a component that mounts more often.

Minor, but could we move the persist call into `toggleFavourite` instead, so it only writes when favourites actually change?

```ts
const toggleFavourite = (route: string) => {
  setFavouriteRoutes(current => {
    const next = current.includes(route) ? current.filter(r => r !== route) : [...current, route]
    storeFavouriteRoutes(next)
    return next
  })
}
```

That drops the `useEffect` entirely and the redundant initial write with it. Not blocking either way.

web/apps/wps-web/src/features/landingPage/landingPageConfig.ts
```diff
@@ -0,0 +1,13 @@
+export const BCPS_SECTION_ID = 'bcps-access-only-heading'
+export const FAVOURITES_SECTION_ID = 'favourites-heading'
+export const PUBLIC_SECTION_ID = 'public-access-heading'
+export const LANDING_PAGE_FAVOURITES_STORAGE_KEY = 'wps-landing-page-favourites'
+
+export const ICON_TILE_RADIUS = '14px'
+export const SECTION_RADIUS = '18px'
+export const TOOL_ROW_RADIUS = '16px'
+export const FAVOURITES_COLOUR = '#3f743f'
```
conbrad: `FAVOURITES_COLOUR` has two consumers, and one is the direct parent of the other. There's no reader anywhere else in the app who needs to discover this value so colocating it in `LandingPage.tsx` and passing it down as `favouritesColour` (the same way `backgroundColour`/`borderColour` are already threaded into `ToolSection` in this same file) means the value lives where it's used.

web/apps/wps-web/src/features/landingPage/landingPageConfig.ts
```diff
@@ -0,0 +1,13 @@
+export const BCPS_SECTION_ID = 'bcps-access-only-heading'
```
conbrad: `BCPS_SECTION_ID`, `FAVOURITES_SECTION_ID`, `PUBLIC_SECTION_ID` are each used in just `LandingPage.tsx`, and consumed only as aria-labelledby/id pairing inside `ToolSection.tsx` so I think we can move them there so readers don't expect them to be reused elsewhere.

web/apps/wps-web/src/features/landingPage/landingPageConfig.ts
```diff
@@ -0,0 +1,13 @@
...
+export const FAVOURITES_SECTION_ID = 'favourites-heading'
+export const PUBLIC_SECTION_ID = 'public-access-heading'
+export const LANDING_PAGE_FAVOURITES_STORAGE_KEY = 'wps-landing-page-favourites'
+
+export const ICON_TILE_RADIUS = '14px'
```
conbrad: `ICON_TILE_RADIUS` / `SECTION_RADIUS` / `TOOL_ROW_RADIUS` are each used in just one place, `ToolIconTile.tsx`, `ToolSection.tsx`, `ToolRow.tsx` respectively. I think they can just live where they are used so readers don't expect them to be shared elsewhere.

web/apps/wps-web/src/features/landingPage/landingPageConfig.ts
```diff
@@ -0,0 +1,13 @@
+export const BCPS_SECTION_ID = 'bcps-access-only-heading'
+export const FAVOURITES_SECTION_ID = 'favourites-heading'
+export const PUBLIC_SECTION_ID = 'public-access-heading'
+export const LANDING_PAGE_FAVOURITES_STORAGE_KEY = 'wps-landing-page-favourites'
```
conbrad: `LANDING_PAGE_FAVOURITES_STORAGE_KEY` is only used in `favouritesStorage.ts` plus two test files (`favouritesStorage.test.ts`, `LandingPage.test.tsx`) that assert against the same key. Tests can just import it from `favouritesStorage.ts`.

web/apps/wps-web/src/features/landingPage/landingPageConfig.ts
```diff
@@ -0,0 +1,13 @@
...
+
+export const ICON_TILE_RADIUS = '14px'
+export const SECTION_RADIUS = '18px'
+export const TOOL_ROW_RADIUS = '16px'
+export const FAVOURITES_COLOUR = '#3f743f'
+export const PUBLIC_TOOL_ICON_COLOUR = '#FF6900'
```
conbrad: `PUBLIC_TOOL_ICON_COLOUR` has three unrelated consumers (`toolInfo.tsx`, `LandingPage.tsx`, `QuickAccessDrawer.tsx`) and `@wps/ui/theme` already houses the existing `LANDING_BACKGROUND_COLOUR` so it could be moved there.

web/apps/wps-web/src/features/landingPage/landingPageConfig.ts
```diff
@@ -0,0 +1,13 @@
...
+
+export const BCWS_PREDICTIVE_SERVICES_EMAIL = 'BCWS.PredictiveServices@gov.bc.ca'
```
conbrad: This is used in `toolInfo.tsx` and `Footer.tsx`, I think we could pull out the non-tsx related managed data into `managedBy.ts` that houses some of the related interfaces and constants:
```tsc
// managedBy.ts
export interface ManagedByInfo {
  name: string
  href?: string
  opensFeedback?: boolean
}

export const CSBC_PREDICTIVE_SERVICES_MANAGED_BY: ManagedByInfo = {
  name: 'CSBC - Predictive Services',
  opensFeedback: true
}

const BCWS_PREDICTIVE_SERVICES_EMAIL = 'BCWS.PredictiveServices@gov.bc.ca'

export const BCWS_PREDICTIVE_SERVICES_MANAGED_BY: ManagedByInfo = {
  name: 'BCWS - Predictive Services',
  href: `mailto:${BCWS_PREDICTIVE_SERVICES_EMAIL}`
}
```
Then `toolInfo.tsx` and `Footer.tsx`can import what they need from `managedBy.ts`

web/apps/wps-web/src/features/landingPage/landingPageConfig.ts
```diff
@@ -0,0 +1,13 @@
...
+export const TECH_SERVICES_EMAIL = 'BCWS.TechServices@gov.bc.ca'
```
conbrad: Only used in `SupportSection.tsx` so it can be moved there.

conbrad (commented):
Looks good! Most of my comments relate to some low-cohesion code in `landingPageConfig.ts`.

We can also delete `LANDING_BACKGROUND_COLOUR` in `theme.ts`.

web/apps/wps-web/src/features/landingPage/landingPageConfig.ts
brettedw: I'm curious, why suggest `FAVOURITES_COLOUR` be passed as a prop, but `PUBLIC_TOOL_ICON_COLOUR` be moved to `theme`?

conbrad: `toolInfo.tsx`'s `ToolInfo` objects (`fireBehaviourAdvisoryInfo`, `percentileCalcInfo`, etc.) are top-level constants built once at module-import time, before any component tree exists. `icon: <PercentIcon htmlColor={PUBLIC_TOOL_ICON_COLOUR} />` is constructed exactly once and reused everywhere that tool's icon is rendered. There's no component instance there to hand a prop to. `FAVOURITES_COLOUR` is different: both its consumers are components (`LandingPage.tsx` renders `QuickAccessDrawer` directly), so the value can be decided once by the parent and threaded down as a normal prop.

web/apps/wps-web/src/features/landingPage/landingPageConfig.ts
brettedw: All of the config comments have been addressed in [69d4a56](https://github.com/bcgov/wps/pull/5632/commits/69d4a567008f337e3dbc69c0c49eea68fb079b00)

web/apps/wps-web/src/features/landingPage/components/Sidebar.tsx
brettedw: Done in [9ca48d3](https://github.com/bcgov/wps/pull/5632/commits/9ca48d3019a3afdc4ef07147544ab3caf0de5164)

web/apps/wps-web/src/features/landingPage/pages/LandingPage.tsx
brettedw: Done in [0e4e7d4](https://github.com/bcgov/wps/pull/5632/commits/0e4e7d4603b78cb5f4ab613c56760a756a8ca961)

web/apps/wps-web/src/features/landingPage/components/ToolRow.tsx
brettedw: Great catch! Done in [c45a053](https://github.com/bcgov/wps/pull/5632/commits/c45a053114ca17469bab00679c99b73e936de8b4)

web/apps/wps-web/src/features/landingPage/components/LandingPageHeader.tsx
brettedw: Ah never noticed that, added to the main page icon here: [500575b](https://github.com/bcgov/wps/pull/5632/commits/500575bce214cec8123fced40c8ac5b3d3e37d23)

web/apps/wps-web/src/features/landingPage/components/ToolRow.tsx
brettedw: Good call, done here: [45092c7](https://github.com/bcgov/wps/pull/5632/commits/45092c7a581d7e8b18daeb4a10930603b9cbe63a)

web/apps/wps-web/src/features/landingPage/landingPageTools.ts
brettedw: Yep that's a good idea, done here: [92fc556](https://github.com/bcgov/wps/pull/5632/commits/92fc5569b46d5385169fcf586a55f06009173248) (used `bcps` rather than `bcws`)

web/apps/wps-web/src/features/landingPage/landingPageConfig.ts
conbrad: Thank you!

conbrad (commented):
I think we want to include the `component={RouterLink} to={tool.route}` for internal and `href+target="_blank"+rel="noreferrer"` external for the links in the quick access menu too.

web/apps/wps-web/src/features/landingPage/components/QuickAccessList.tsx
brettedw: Oops, you're right, missed this one

brettedw: done in[17b7709](https://github.com/bcgov/wps/pull/5632/commits/17b7709f6c7e5aa40f59d08b5f95c12996634c76)

conbrad (approved):
Great work, thanks!!!

PR #5634, opened 2026-07-22
Dev: Force gc.collect() after each 4-panel chart to fix wx-4panel-charts-rdps OOMKilled

conbrad:
`wx-4panel-charts-rdps` pods were being OOMKilled against their `4Gi` memory limit. Profiling against real RDPS GRIB2 data (`profile_4panel_memory.py` (not committed here, scratch script)) showed RSS climbing ~365–420MB per panel generated, with no ceiling. A full 29-panel production run (`END_HOUR=84`, `STEP=3`) crosses `4Gi` by roughly the 10th–11th panel.

`FourPanelChartRunner._make_4panel_chart` already calls `plt.close(fig)` in its `finally` block, but that's not enough: matplotlib's `Figure`/`Axes`/`Artist` object graph forms reference cycles (colorbars, legends, tick/font caches, the Agg canvas) that plain refcounting can't break, only the cyclic garbage collector can, and it wasn't running often enough on its own to keep pace with how fast this pipeline generates garbage. Toggling an explicit `gc.collect()` on and off in the profiling harness confirmed it directly — without it, `gc.collect()` finds 20,000+ collectable objects per panel and RSS keeps climbing; with it forced right after `plt.close(fig)`, RSS goes flat after the first panel.

Getting this fix through CI surfaced a second, unrelated problem: `wps-weather`'s OpenShift-native image build (`oc process`/`BuildConfig`) was timing out against the cluster's API server, and separately its Dockerfile failed to compile `gdal`/`fiona` from source (missing `python3-dev`/`build-essential`, exposed once the base image's newer Ubuntu release stopped having prebuilt wheels available for its Python version). Rather than just patch around it, `wps-weather`'s build+push moved off `oc` entirely onto `bcgov/actions/builder-ghcr`, pushing to `ghcr.io/bcgov/wps/wps-weather`. Production promotion is now a registry-side image copy (`oc_promote_gh.sh`, written generically so other packages can reuse it, not just `wps-weather`) instead of `oc tag`, and the old OpenShift-native build path is kept available as a separate, manually-triggered fallback (`build_openshift_manual.yml`) rather than deleted. Also cleaned up the SonarCloud/GitHub Advanced Security findings this introduced along the way: SHA-pinning every external Action, moving `workflow_dispatch` inputs out of `run:` blocks to avoid script injection, scoping `packages: write` to only the job that needs it, and adding `--no-install-recommends` to the Dockerfile's apt install.

# Test Links:
[Landing Page](https://wps-pr-5634-e1e498-dev.apps.silver.devops.gov.bc.ca/)
[MoreCast](https://wps-pr-5634-e1e498-dev.apps.silver.devops.gov.bc.ca/morecast)
[Percentile Calculator](https://wps-pr-5634-e1e498-dev.apps.silver.devops.gov.bc.ca/percentile-calculator)
[FireCalc](https://wps-pr-5634-e1e498-dev.apps.silver.devops.gov.bc.ca/fire-behaviour-calculator)
[FireCalc bookmark](https://wps-pr-5634-e1e498-dev.apps.silver.devops.gov.bc.ca/fire-behaviour-calculator?s=266&f=c5&c=NaN&w=20,s=286&f=c7&c=NaN&w=16,s=1055&f=c7&c=NaN&w=NaN,s=305&f=c7&c=NaN&w=NaN,s=344&f=c5&c=NaN&w=NaN,s=346&f=c7&c=NaN&w=NaN,s=328&f=c7&c=NaN&w=NaN,s=1399&f=c7&c=NaN&w=NaN,s=334&f=c7&c=NaN&w=NaN,s=1082&f=c3&c=NaN&w=NaN,s=388&f=c7&c=NaN&w=NaN,s=309&f=c7&c=NaN&w=16,s=306&f=c7&c=NaN&w=NaN,s=1029&f=c7&c=NaN&w=NaN,s=298&f=c7&c=NaN&w=NaN,s=836&f=c7&c=NaN&w=NaN,s=9999&f=c7&c=NaN&w=NaN)
[Auto Spatial Advisory (ASA)](https://wps-pr-5634-e1e498-dev.apps.silver.devops.gov.bc.ca/auto-spatial-advisory)
[HFI Calculator](https://wps-pr-5634-e1e498-dev.apps.silver.devops.gov.bc.ca/hfi-calculator)
[SFMS Insights](https://wps-pr-5634-e1e498-dev.apps.silver.devops.gov.bc.ca/insights)
[Fire Watch](https://wps-pr-5634-e1e498-dev.apps.silver.devops.gov.bc.ca/fire-watch)
[Weather Toolkit](https://wps-pr-5634-e1e498-dev.apps.silver.devops.gov.bc.ca/weather-toolkit)

### DIFF
--- a/openshift/scripts/oc_deploy_to_production.sh
+++ b/openshift/scripts/oc_deploy_to_production.sh
@@ -34,12 +34,13 @@ MODULE_NAME=web bash $(dirname ${0})/oc_promote.sh ${SUFFIX} ${RUN_TYPE}
 MODULE_NAME=jobs bash $(dirname ${0})/oc_promote.sh ${SUFFIX} ${RUN_TYPE}
 
 # wps-weather lives on GHCR now, promoted via a registry-side copy rather than `oc tag`.
-# Tolerant of a missing PR image, same as the other modules -- wps-weather isn't always
-# rebuilt if nothing in it changed, in which case "prod" just keeps pointing at whatever
-# it already was. The resulting path is always ghcr.io/bcgov/wps/wps-weather:prod, which
-# the provisioning scripts below already default WEATHER_IMAGE to (since they're called
-# with SUFFIX=prod) -- nothing needs to be captured or passed through here.
-PACKAGE=wps-weather bash $(dirname ${0})/oc_promote_gh.sh ${SUFFIX} ${RUN_TYPE} || true
+# wps-weather isn't always rebuilt if nothing in it changed, in which case there's no PR
+# image to promote and oc_promote_gh.sh exits 0, leaving "prod" pointing at whatever it
+# already was. An actual promotion failure (registry/auth/network error) exits non-zero 
+#and fails this deploy like everything else here. The resulting path is always
+# ghcr.io/bcgov/wps/wps-weather:prod, which the provisioning scripts below
+# already default WEATHER_IMAGE to (since they're called with SUFFIX=prod).
+PACKAGE=wps-weather bash $(dirname ${0})/oc_promote_gh.sh ${SUFFIX} ${RUN_TYPE}
 
 echo Provision database
 PROJ_TARGET=${PROJ_TARGET} BUCKET=lwzrin CPU_REQUEST=2 MEMORY_REQUEST=2Gi MEMORY_LIMIT=16Gi DATA_SIZE=65Gi WAL_SIZE=15Gi REPLICAS=3 bash $(dirname ${0})/oc_provision_crunchy.sh prod ${RUN_TYPE}
--- a/openshift/scripts/oc_deploy_to_production.sh
+++ b/openshift/scripts/oc_deploy_to_production.sh
@@ -32,7 +32,14 @@ echo Promote
 MODULE_NAME=api bash $(dirname ${0})/oc_promote.sh ${SUFFIX} ${RUN_TYPE}
 MODULE_NAME=web bash $(dirname ${0})/oc_promote.sh ${SUFFIX} ${RUN_TYPE}
 MODULE_NAME=jobs bash $(dirname ${0})/oc_promote.sh ${SUFFIX} ${RUN_TYPE}
-MODULE_NAME=weather bash $(dirname ${0})/oc_promote.sh ${SUFFIX} ${RUN_TYPE}
+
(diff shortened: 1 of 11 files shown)


openshift/scripts/oc_deploy_to_production.sh
```diff
@@ -32,7 +32,14 @@ echo Promote
...
 MODULE_NAME=jobs bash $(dirname ${0})/oc_promote.sh ${SUFFIX} ${RUN_TYPE}
-MODULE_NAME=weather bash $(dirname ${0})/oc_promote.sh ${SUFFIX} ${RUN_TYPE}
+
+# wps-weather lives on GHCR now, promoted via a registry-side copy rather than `oc tag`.
+# Tolerant of a missing PR image, same as the other modules -- wps-weather isn't always
+# rebuilt if nothing in it changed, in which case "prod" just keeps pointing at whatever
+# it already was. The resulting path is always ghcr.io/bcgov/wps/wps-weather:prod, which
+# the provisioning scripts below already default WEATHER_IMAGE to (since they're called
+# with SUFFIX=prod) -- nothing needs to be captured or passed through here.
+PACKAGE=wps-weather bash $(dirname ${0})/oc_promote_gh.sh ${SUFFIX} ${RUN_TYPE} || true
```
brettedw: Is there any significant risk of hiding real failures/errors here?

conbrad: Yeah, this is real riskthanks. `|| true` swallows both "no PR image, nothing to promote" (expected) and "the actual `docker buildx imagetools create` copy failed" (not expected) identically, so a transient GHCR/auth failure would silently leave prod on the stale tag with no signal in CI. I'll mirror `oc_promote.sh`'s `tag_if_exists_or_fail` pattern, check image existence explicitly and only tolerate that case, let a failure in the actual promote step propagate and fail the deploy like every other module does.

conbrad: Adjusted here: https://github.com/bcgov/wps/pull/5634/commits/f24a03e41423b213f3bca66af5f10951d1403e47

brettedw (approved):
👌👌👌

PR #5641, opened 2026-07-24
SFMS: Vectorize cffdrs FBP functions and validate against published GLC-X-10 test data

### LINKED ISSUE #4893: SFMS: Test Data
**Describe the task**

Create a SFMS Insights: Test Data Environment

**Acceptance Criteria**
- [ ] first
- [ ] second
- [ ] third

**Additional context**
- #4874 
- Or here

conbrad:
Bumps `cffdrs` to its latest commit and adds a new `cffdrs_vec` package that numba-vectorizes the full FBP function family alongside the FWI functions that were already vectorized in `wps_shared` (now moved here too), so both can run over whole rasters instead of station-by-station. Composing the vectorized functions the same way `cffdrs`'s own `_fire_behaviour_prediction` does reproduces all 20 published FBP test cases from GLC-X-10 (Wotton et al. 2009) to within floating-point noise, addressing `#4893`.

Several FBP functions (`rate_of_spread`, `back_rate_of_spread`, `slope_adjustment`, ...) call other plain-Python `cffdrs` functions internally, which numba can't compile through directly. JIT-compiling them against `cffdrs`'s real modules in place worked, but silently broke `app.fire_behaviour.cffdrs`'s normal scalar calls anywhere both were imported in the same process. `numba`'s `nopython` mode has to type-check every branch up front, so a call like `rate_of_spread(..., pc=None)` (fine in plain Python, since the `pc`-using branch is only for M1-M4) started raising. `fbp.py` instead builds each composite function against a private, isolated clone of the relevant `cffdrs` submodule, so nothing outside this file is affected.

`numba`'s `@vectorize` turns a plain scalar function (one number in per argument, one number out) into a real numpy `ufunc`. It broadcasts automatically over arrays of any shape, so a function written to handle one case at a time gets applied elementwise across an entire raster with no Python-level loop. Most of the `cffdrs` functions this package wraps fit that shape directly.

`@guvectorize` is for the handful that need to return more than one value at once (`vectorized_slope_adjustment` returns `wsv` and `raz`; `vectorized_rate_of_spread_extended` returns `ros`, `cfb`, `csi`, `rso`). `@vectorize` only supports a single return value, so those need the generalized form, an explicit signature (`"(),(),...->(),()"`) describing each argument's shape, and the function writes results into output-array parameters instead of returning them, but the effect is the same: still one call, still broadcasts elementwise over the whole array, just able to hand back several results per call instead of one.

Closes #4893
 
# Test Links:
[Landing Page](https://wps-pr-5641-e1e498-dev.apps.silver.devops.gov.bc.ca/)
[MoreCast](https://wps-pr-5641-e1e498-dev.apps.silver.devops.gov.bc.ca/morecast)
[Percentile Calculator](https://wps-pr-5641-e1e498-dev.apps.silver.devops.gov.bc.ca/percentile-calculator)
[FireCalc](https://wps-pr-5641-e1e498-dev.apps.silver.devops.gov.bc.ca/fire-behaviour-calculator)
[FireCalc bookmark](https://wps-pr-5641-e1e498-dev.apps.silver.devops.gov.bc.ca/fire-behaviour-calculator?s=266&f=c5&c=NaN&w=20,s=286&f=c7&c=NaN&w=16,s=1055&f=c7&c=NaN&w=NaN,s=305&f=c7&c=NaN&w=NaN,s=344&f=c5&c=NaN&w=NaN,s=346&f=c7&c=NaN&w=NaN,s=328&f=c7&c=NaN&w=NaN,s=1399&f=c7&c=NaN&w=NaN,s=334&f=c7&c=NaN&w=NaN,s=1082&f=c3&c=NaN&w=NaN,s=388&f=c7&c=NaN&w=NaN,s=309&f=c7&c=NaN&w=16,s=306&f=c7&c=NaN&w=NaN,s=1029&f=c7&c=NaN&w=NaN,s=298&f=c7&c=NaN&w=NaN,s=836&f=c7&c=NaN&w=NaN,s=9999&f=c7&c=NaN&w=NaN)
[Auto Spatial Advisory (ASA)](https://wps-pr-5641-e1e498-dev.apps.silver.devops.gov.bc.ca/auto-spatial-advisory)
[HFI Calculator](https://wps-pr-5641-e1e498-dev.apps.silver.devops.gov.bc.ca/hfi-calculator)
[SFMS Insights](https://wps-pr-5641-e1e498-dev.apps.silver.devops.gov.bc.ca/insights)
[Fire Watch](https://wps-pr-5641-e1e498-dev.apps.silver.devops.gov.bc.ca/fire-watch)
[Weather Toolkit](https://wps-pr-5641-e1e498-dev.apps.silver.devops.gov.bc.ca/weather-toolkit)


backend/packages/cffdrs-vec/src/cffdrs_vec/fwi.py
```diff
@@ -0,0 +1,20 @@
...
+    fine_fuel_moisture_code,
+    fire_weather_index,
+)
+from cffdrs.fwi import _initial_spread_index
+from numba import vectorize
+
+vectorized_bui = vectorize(buildup_index)
+vectorized_dc = vectorize(drought_code)
+vectorized_dmc = vectorize(duff_moisture_code)
+vectorized_ffmc = vectorize(fine_fuel_moisture_code)
```
brettedw, line 14: Don't these other fwi functions also have range-validating `ValueError`s? I don't recall that being an issue before unless I'm missing something here

conbrad: Yeah the comment is wrong, I'll update it.

The actual difference is that `initial_spread_index` is the only one of the six where `cffdrs` itself split validation from computation into two functions - the public wrapper just does `if ... raise ValueError` then `return _initial_spread_index(ffmc, ws, fbp_mod)`. `numba`'s nopython mode can't compile a call into a second function it hasn't itself compiled, so vectorizing the public wrapper throws a `TypingError` on `_initial_spread_index`. `buildup_index`, `drought_code`, `duff_moisture_code`, `fine_fuel_moisture_code`, `fire_weather_index` are each one self-contained function (guards + math, no delegation), so they vectorize fine, guards included.

backend/packages/cffdrs-vec/src/cffdrs_vec/tests/_glc_x_10_data.py
brettedw, line 1: There a lot happening in this file, I'm a bit confused by what the fbp functions starting on line 591 are. Do they belong in this file referencing "data"? Should our own vectorized fbp calculations be capable of creating a primary output?

backend/packages/cffdrs-vec/src/cffdrs_vec/tests/_glc_x_10_data.py
```diff
@@ -0,0 +1,719 @@
...
+RASTER_GEOTRANSFORM = (-140.0, 1.0, 0.0, 60.0, 0.0, -1.0)
+
+# Fuel types with no crown to burn - _fire_behaviour_prediction always zeroes FMC for these.
+NO_CROWN_FUEL_TYPE_CODES = (D1, S1, S2, S3, O1A, O1B)
+
+FIRE_TYPE_BY_FD_CODE = {"S": "SUR", "I": "IC", "C": "CC"}
+
+# Table 7 (FCFDG 1992) CBH/CFL defaults, used because the GLC-X-10 dataset leaves CBH/CFL blank
+# for every case (relying on cffdrs's own per-fuel-type defaults). Matches
+# wps_shared.fuel_types.FUEL_TYPE_DEFAULTS exactly.
```
brettedw: I know this is it's own package, but what are your thoughts on duplicating all of these fuel type defaults vs sharing them between packages?

backend/packages/cffdrs-vec/src/cffdrs_vec/tests/_glc_x_10_data.py
```diff
@@ -0,0 +1,719 @@
...
+    output_csv: str
+    id_input_col: str
+    id_output_col: str
+    raster_dir: Path
+    columns: dict
+    # ROS/HFI/CFB tolerances - looser for "paper" to absorb the small, known drift between the
+    # paper's 2009-vintage published values and what the current FBP equations compute (see
+    # test_fbp_glc_x_10.py).
+    rtol_ros_hfi: float
+    atol_cfb: float
```
brettedw, line 214: Maybe worth a more descriptive docstring for inputs on some of these so we aren't re-creating the same R package issues where someone less familiar with these wouldn't know what they are

backend/packages/cffdrs-vec/src/cffdrs_vec/tests/test_fbp_glc_x_10_rasters.py
```diff
@@ -0,0 +1,70 @@
...
+Same 20 published GLC-X-10 FBP System test cases and the same calculate_primary_output() as
+test_fbp_glc_x_10.py, but sourced from the GeoTIFF fixtures under fixtures/glc_x_10/ and
+fixtures/glc_x_10_paper/ (generated by generate_glc_x_10_rasters.py, via GLCX10Source.
+load_raster_inputs()) instead of the CSVs directly, showing cffdrs_vec.fbp's vectorized functions
+also work correctly when driven by real GDAL raster reads (float32 storage, WPSDataset's nodata
+handling) rather than plain float64 arrays parsed straight from CSV.
+
+Parametrized over the same GLCX10Source objects as test_fbp_glc_x_10.py (see _glc_x_10_data.py),
+so the tolerance divergence between them (paper's ROS/HFI/CFB need looser tolerance, see
+test_fbp_glc_x_10.py for why) lives in exactly one place.
```
brettedw: I'm a bit unclear what this means, the tolerance divergence between them lives in one place?

backend/packages/cffdrs-vec/src/cffdrs_vec/fwi.py
brettedw: Wasn't that change made in the most recent `cffdrs_py` PR? I never got a chance to fully review that PR, should that change be reverted for consistency, or is it needed for other reasons?

I guess I'm just noticing in the `cffdrs_py` package, the isi calc guard on ffmc is no longer used on other functions that take ffmc as a direct input parameter because they now use `_initial_spread_index`

backend/packages/cffdrs-vec/src/cffdrs_vec/tests/test_sfms_bui_vectorized.py
```diff
@@ -0,0 +1,1907 @@
...
+    dmc_array = read_raster_array(os.path.join(bui_tif_dir, "dmc20240528.tif"))
+    bui_array = read_raster_array(os.path.join(bui_tif_dir, "bui20240528.tif"))
+
+    res = vectorized_bui(dmc_array, dc_array)
+
+    assert np.allclose(bui_array, res)
+
+
+@pytest.mark.parametrize(
+    "dmc,dc,expected_bui",
```
brettedw, line 49: Where are all of these test cases from?

backend/packages/cffdrs-vec/src/cffdrs_vec/tests/test_nan_and_none_handling.py
```diff
@@ -0,0 +1,578 @@
...
+(every function needs both float-only branches to still work), so each case just fixes it
+wherever's convenient. call_reference takes the same nan_args list and unwraps whatever it needs
+to call the plain, unjitted cffdrs function.
+
+multi_output marks the 2 guvectorize-based functions (slope_adjustment, rate_of_spread_extended),
+which return a tuple of arrays instead of one array.
+
+Argument-array construction always happens before entering `with pytest.raises`, so those blocks
+contain nothing but the single call under test - see python:S5778.
+"""
```
brettedw, line 11: I don't know if it's possible but are module level comments like this possible to simplify? It's a lot to read and understand before getting into tests

backend/packages/cffdrs-vec/src/cffdrs_vec/fwi.py
conbrad: So it's not a `numba` blocker at all, that was wrong. `numba` can compile the `ValueError` raises and can compile a function called from inside another compiled function.

The actual blocker for `initial_spread_index` specifically is that its public wrapper is two functions: `if ...: raise ValueError(...)` then `return _initial_spread_index(ffmc, ws, fbp_mod)` - a call into a second, undecorated Python function. `numba` can't compile a call into a function it hasn't itself compiled, which is a completely different problem from the guard. `buildup_index`, `drought_code`, `duff_moisture_code`, `fine_fuel_moisture_code`, `fire_weather_index` are each one self-contained function (no delegation), so they never hit this at all.

So the `cffdrs` splitting ISI into two functions is not a `numba` requirement, but a design/convention one:
- `_initial_spread_index` is called internally elsewhere in cffdrs - `slope_calc.py`'s `_slope_adjustment` and `fire_behaviour_prediction.py` both use it as an intermediate step in other FBP formulas. None of the other five are ever called internally by `cffdrs` itself, so nothing forced a private sibling for them.
- An existing convention within the `cffdrs_py`, inputs are already validated/clamped once at the top , so re-validating an already-valid value on every internal call would be redundant. 

Both are deliberate design choices, not compiler necessities.

backend/packages/cffdrs-vec/src/cffdrs_vec/tests/_glc_x_10_data.py
conbrad: Yeah so this orchestration logic is substantial, general-purpose, and currently trapped in a file whose stated job is loading test fixtures... I'll change that.

conbrad: Ok so this surfaces vectorization issue in `cffdrs_py`'s `fire_behaviour_prediction` function that I've put up for review here: https://github.com/cffdrs/cffdrs_py/pull/16

backend/packages/cffdrs-vec/src/cffdrs_vec/tests/_glc_x_10_data.py
conbrad: We could do, there's one wrinkle though: this file's dicts encode Table 7 (FCFDG 1992) values as transcribed for validating against the published paper (GLC-X-10/Wotton et al. 2009), the one in `wps_shared` includes extra fuel types like D2/C7B/M2_25). But I think it's ok to reuse it, I'll update.

backend/packages/cffdrs-vec/src/cffdrs_vec/fwi.py
brettedw: > An existing convention within the cffdrs_py, inputs are already validated/clamped once at the top

Not quite sure what this is referring to. 
I'm more wondering why in the most recent `cffdrs_py` PR was the private `_initial_spread_index` created? It seems like it created the issue where the public `initial_spread_index` can't be vectorized, and it also seems like it had an indirect effect of functions like `fire_behaviour_prediction` no longer checking the bounds of `ffmc` and `ws`

conbrad: > Not quite sure what this is referring to.

A convention emerged for the FBP system during that commit where public function validates, private and the `_`-prefixed sibling do the bare computation. That pattern got applied mechanically to every function touched by `fire_behaviour_prediction`'s internal call graph just for the sake of consistency, ISI included, rather than each function being evaluated individually for whether it actually needed the split.

> I'm more wondering why in the most recent cffdrs_py PR was the private _initial_spread_index created?

`_initial_spread_index` exists because it is also called internally by `_slope_adjustment` (in `slope_calc.py)` and `_fire_behaviour_prediction`, following the convention above. In hindsight not a great pattern to follow in this case.

brettedw: I see, though the comment above suggests the validation is not an issue if I understand correctly
> numba can compile the ValueError raises and can compile a function called from inside another compiled function.

So is the separation (public/private) required if the bigger orchestration functions also rely on that same input/arg validation?

conbrad: Correct, validation in terms of `ValueError` raises are not an issue, at least for `numba`.

Separation was only required for two cases for a function: it branches on a fuel-type string or recurses with a different string argument. `numba` can't compile those cases. The private functions were added to address that without changing the library's existing public API functions. It then looked like the public API functions all handled validation and not the private functions which blindly I accepted as a convention which I think lead to this ISI issue. 

I've opened a PR `cffdrs_py` here: https://github.com/cffdrs/cffdrs_py/pull/17

backend/packages/cffdrs-vec/src/cffdrs_vec/tests/_glc_x_10_data.py
conbrad: Done in: https://github.com/bcgov/wps/pull/5641/commits/e26838c4f899574fcb22cd3c5be7bef95f39438a

backend/packages/cffdrs-vec/src/cffdrs_vec/tests/test_sfms_bui_vectorized.py
conbrad: They were moved from `wps_shared` along with the vectorized `fwi` functions.

backend/packages/cffdrs-vec/src/cffdrs_vec/tests/_glc_x_10_data.py
conbrad: Added descriptive docstrings here: https://github.com/bcgov/wps/pull/5641/commits/3b1edad6a156a044b268a1b0f4fe4eedbd6df33f

backend/packages/cffdrs-vec/src/cffdrs_vec/tests/test_fbp_glc_x_10_rasters.py
conbrad: Yeah it was confusing, unnecessary context, I removed it here: https://github.com/bcgov/wps/pull/5641/commits/3fe560f2f6513da9b7ac7417ecad6467695dff00

It was trying to communicate how `GLCX10Source` contains all inputs, outputs and tolerances all in one class.

backend/packages/cffdrs-vec/src/cffdrs_vec/tests/test_nan_and_none_handling.py
conbrad: Yeah... bit of a mouthful, simplified here: https://github.com/bcgov/wps/pull/5641/commits/808d1d8a19bc17f578d0fe95df1ab2b89f9c3bc7

I'm also gonna see if I can simplify the test parameters.

backend/packages/cffdrs-vec/src/cffdrs_vec/tests/test_fbp_glc_x_10.py
```diff
@@ -0,0 +1,167 @@
+"""
+Validates cffdrs_vec.fbp.vectorized_fire_behaviour_prediction against the 20 published FBP System
+test cases from GLC-X-10 (Wotton et al. 2009), Table 4 (inputs) and Table 5 (primary outputs).
+See https://github.com/bcgov/wps/issues/4893.
+
+Parametrized over two independent data sources (see _glc_x_10_data.py's GLCX10Source for exact
+provenance of each):
+- "r_package": the R cffdrs package's own digitized transcription of the paper's Tables 4/5,
+  vendored here by way of cffdrs_py's own copy of those same files.
+- "paper": transcribed independently, straight from the published PDF via `pdftotext -layout`,
```
brettedw, line 10: is `pdftotext` a python package?

conbrad: Yep: https://pypi.org/project/pdftotext/

backend/packages/cffdrs-vec/src/cffdrs_vec/fbp.py
```diff
@@ -0,0 +1,763 @@
...
+    trost: np.ndarray
+    tcfb: np.ndarray
+    tfi: np.ndarray
+    ttfc: np.ndarray
+    tti: np.ndarray
+    wsv0: np.ndarray
+    raz0: np.ndarray
+
+
+def vectorized_fire_behaviour_prediction(
```
brettedw, line 582: If I'm right this is going to be own main function for calling fbp calculations. Can we provide docstrings for what each parameter is so we aren't recreating the R package issue of not knowing what the parameters actually is without already being very familiar with it?

conbrad: Done in: https://github.com/bcgov/wps/pull/5641/commits/a77a8d445082aad0f6b5902d6da0a883be86d766

brettedw (approved):
Thanks for taking that one on, looks good!

PR #5657, opened 2026-07-30
ASA Go: Subscription Fixes

brettedw:
Issue:   Some devices had multiple token rows with the same `device_id`, including old inactive rows. Notification setting requests could include data from multiple rows or update the wrong row, resulting in duplicate zone inserts and a database unique constraint error
  
This fixes notification subscription errors caused by multiple `device_token` rows sharing the same `device_id`

  - Reads and updates notification settings only for the active device-token row
  - Cleans up existing duplicate device rows, keeping the active row with a fallback to the most recently updated row
  - Adds a unique constraint so each `device_id` can only have one row
  - Updates the existing device row when its FCM token changes, preserving its notification settings.
  - Returns 409 when a token and device ID belong to different existing rows rather than deleting potentially valid data (this is an unlikely scenario at this point, and hopefully we never run into it, but it's just an additional safety guard)

# Test Links:
[Landing Page](https://wps-pr-5657-e1e498-dev.apps.silver.devops.gov.bc.ca/)
[MoreCast](https://wps-pr-5657-e1e498-dev.apps.silver.devops.gov.bc.ca/morecast)
[Percentile Calculator](https://wps-pr-5657-e1e498-dev.apps.silver.devops.gov.bc.ca/percentile-calculator)
[FireCalc](https://wps-pr-5657-e1e498-dev.apps.silver.devops.gov.bc.ca/fire-behaviour-calculator)
[FireCalc bookmark](https://wps-pr-5657-e1e498-dev.apps.silver.devops.gov.bc.ca/fire-behaviour-calculator?s=266&f=c5&c=NaN&w=20,s=286&f=c7&c=NaN&w=16,s=1055&f=c7&c=NaN&w=NaN,s=305&f=c7&c=NaN&w=NaN,s=344&f=c5&c=NaN&w=NaN,s=346&f=c7&c=NaN&w=NaN,s=328&f=c7&c=NaN&w=NaN,s=1399&f=c7&c=NaN&w=NaN,s=334&f=c7&c=NaN&w=NaN,s=1082&f=c3&c=NaN&w=NaN,s=388&f=c7&c=NaN&w=NaN,s=309&f=c7&c=NaN&w=16,s=306&f=c7&c=NaN&w=NaN,s=1029&f=c7&c=NaN&w=NaN,s=298&f=c7&c=NaN&w=NaN,s=836&f=c7&c=NaN&w=NaN,s=9999&f=c7&c=NaN&w=NaN)
[Auto Spatial Advisory (ASA)](https://wps-pr-5657-e1e498-dev.apps.silver.devops.gov.bc.ca/auto-spatial-advisory)
[HFI Calculator](https://wps-pr-5657-e1e498-dev.apps.silver.devops.gov.bc.ca/hfi-calculator)
[SFMS Insights](https://wps-pr-5657-e1e498-dev.apps.silver.devops.gov.bc.ca/insights)
[Fire Watch](https://wps-pr-5657-e1e498-dev.apps.silver.devops.gov.bc.ca/fire-watch)
[Weather Toolkit](https://wps-pr-5657-e1e498-dev.apps.silver.devops.gov.bc.ca/weather-toolkit)


.github/workflows/deployment.yml
```diff
@@ -37,6 +37,7 @@ jobs:
     name: Build ${{ matrix.name }} Image
     runs-on: ubuntu-24.04
     strategy:
+      fail-fast: false
```
brettedw, line 40: This let's some image builds finish, rather than failing the whole run if a single image fails. The action will still fail instead of trying to progress is any of the image builds fail

With all the login issues lately this should help prevent multiple re-runs where different image builds take turns failing, since we'll just let them complete if they're running and then re-run whichever one(s) failed

backend/packages/wps-api/src/app/routers/fcm.py
```diff
@@ -30,22 +30,53 @@
...
+    responses={409: {"description": "Token is already registered to another device."}},
+)
 async def register_device(request: RegisterDeviceRequest):
-    """
-    Upsert a device token for a device_id.
+    """Register or update the FCM token for a device.
+
+    Flow:
+    - If the device exists, update its row. This preserves its notification settings.
+    - If only the token exists, reuse that row with the new device ID.
```
conbrad: Is it possible for a token to exist without a device ID?

backend/packages/wps-api/src/app/routers/fcm.py
```diff
@@ -30,22 +30,53 @@
...
+            device_token is not None
+            and token_match is not None
+            and device_token.id != token_match.id
+        ):
+            logger.error(
+                "Device registration conflict: device_id row %s differs from token row %s",
+                device_token.id,
+                token_match.id,
+            )
+            raise HTTPException(
```
conbrad, line 56: Don't we expect tokens not to match if the request contains a new token to update with?

backend/packages/wps-api/alembic/versions/e4a9f2c7d1b6_unique_device_token_device_id.py
```diff
@@ -0,0 +1,60 @@
...
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision = "e4a9f2c7d1b6"
+down_revision = "cbde211a0c72"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
```
conbrad, line 51: Can we use sqlalchemy instead? Something like:

```
device_token = sa.table(
    "device_token",
    sa.column("id", sa.Integer),
    sa.column("device_id", sa.String),
    sa.column("is_active", sa.Boolean),
    sa.column("updated_at", sa.DateTime),
)
notification_settings = sa.table(
    "notification_settings",
    sa.column("id", sa.Integer),
    sa.column("device_token_id", sa.Integer),
)


def _stale_device_token_ids() -> sa.Select:
    """Selects device_token.id for every row that isn't the one to keep per device_id."""
    row_number = (
        sa.func.row_number()
        .over(
            partition_by=device_token.c.device_id,
            order_by=[
                device_token.c.is_active.desc(),
                device_token.c.updated_at.desc(),
                device_token.c.id.desc(),
            ],
        )
        .label("row_number")
    )
    ranked = sa.select(device_token.c.id, row_number).cte("ranked_device_tokens")
    return sa.select(ranked.c.id).where(ranked.c.row_number > 1)


def upgrade():
    stale_ids = _stale_device_token_ids()
    op.execute(
        notification_settings.delete().where(
            notification_settings.c.device_token_id.in_(stale_ids)
        )
    )
    op.execute(device_token.delete().where(device_token.c.id.in_(stale_ids)))
```

backend/packages/wps-api/src/app/routers/fcm.py
brettedw: This is: if the token exists in our db, but the device doesn't exist. We were taking this into account before with `get_device_by_token` because when installing dev versions with xcode on ios, the device id changes a lot, but the token may not

backend/packages/wps-api/src/app/routers/fcm.py
brettedw: Yes that's correct, that's covered by the first scenario: If the device exists, update its row.

This is safe guarding a scenario when both the device and the token are found in our db, but they're in different rows. I'm hoping we don't run into this ever, but it's kind of the scenario that could have been possible with the original error reported by NL

backend/packages/wps-api/src/app/routers/fcm.py
conbrad: I guess I assumed a token is only every created and stored if a device ID is stored. Doesn't a token depend on a device? Maybe there's a time between a token existing but not a device ID existing I'm not aware of?

backend/packages/wps-api/src/app/routers/fcm.py
conbrad: What if I have deviceID `A`, and token `A`  in the database, then FCM wants to update my token to `B`, this would return a 409 and my token wouldn't get updated right?

conbrad: Nevermind, I think I see what you mean, I just find the logic paths a bit hard to parse.

brettedw: No not quite, that would result in a token rotation on that device row.

If the request contains:
  `device_id=A, token=B`

  the lookups return:

  `device_token = row_for_device_A`
  `token_match = None`

  A 409 only occurs if token B is already assigned to a different database row. That is ambiguous data, so the endpoint avoids overwriting either device.

```
The last case should be uncommon, but legacy data or an unusual token reassignment could cause
it. We cannot safely know which row and notification settings to keep, so the guard prevents us
from silently deleting or overwriting a user's data. Hopefully we never run into this, it's there
as a safety measure.
```

backend/packages/wps-api/src/app/routers/fcm.py
brettedw: You’re right that a token is never stored without a device ID, both are required on every row. “Only the token exists” means the incoming token matches an existing row, but the incoming device ID is different from the one stored on that row. We handle this because ios dev installs can produce a new device ID while keeping the same FCM token. In that case, we update the existing row with the new device ID.

I'll try to update the comment to make it clearer

backend/packages/wps-api/src/app/routers/fcm.py
conbrad: I think what's confusing me is the multiple row lookup and the `device_token` variable. By the time you're reading the update block, `device_token` could be three different things depending on which branch got you there: 

- Row matched by `device_id`
- Row matched by token (when above was None)
- A new, unsaved `DeviceToken` (when neither matched)

I think we could pull out some of the complexity from the router and have the lookup check go in the `crud/fcm.py` layer with something like:
```
class DeviceTokenConflictError(Exception):
    """Raised when device_id and token already exist as two different rows."""

    def __init__(self, device_id_row_id: int, token_row_id: int):
        self.device_id_row_id = device_id_row_id
        self.token_row_id = token_row_id
        super().__init__(
            f"Device registration conflict: device_id row {device_id_row_id} "
            f"differs from token row {token_row_id}"
        )
```

```
async def get_device_token_match(
    session: AsyncSession, device_id: str, token: str, for_update: bool = False
) -> DeviceToken | None:
    """Return the existing row for device_id and/or token, or None if neither exists.

    Raises DeviceTokenConflictError if device_id and token belong to different rows.
    """
    statement = select(DeviceToken).where(
        or_(DeviceToken.device_id == device_id, DeviceToken.token == token)
    )
    if for_update:
        statement = statement.with_for_update()
    rows = list((await session.scalars(statement)).all())

    row_by_device_id = next((r for r in rows if r.device_id == device_id), None)
    row_by_token = next((r for r in rows if r.token == token), None)

    if (
        row_by_device_id is not None
        and row_by_token is not None
        and row_by_device_id.id != row_by_token.id
    ):
        raise DeviceTokenConflictError(row_by_device_id.id,

    return row_by_device_id or row_by_token
```

Then the router just becomes:

```
try:
    existing_row = await get_device_token_match(
        session, request.device_id, request.token, for_update=True
    )
except DeviceTokenConflictError as exc:
    logger.error(str(exc))
    raise HTTPException(
        status_code=409, detail="Token is already registered to another device"
    ) from exc

if existing_row:
    existing_row.is_active = True
    existing_row.token = request.token
    existing_row.device_id = request.device_id
    existing_row.platform = request.platform
    existing_row.updated_at = get_utc_now()
    existing_row.user_id = request.user_id
    logger.info(f"Updated existing DeviceToken record for token: {request.token}")
else:
    new_device_token = DeviceToken(...)
    save_device_token(session, new_device_token)
    logger.info("Successfully created new DeviceToken record.")
```

This allows for:
- `existing_row` to be assigned once  and never changes meaning after that. 
- one query and one lock instead of two. The original code does two separate `SELECT ... FOR UPDATE` round trips. The new lookup does one `OR` query, which is strictly fewer round trips and also closes the (small but real) window between the two locks where a concurrent registration could interleave.
- The 409 logic lives where the context is. Deciding "same row / different rows / nothing" requires knowing both search keys and correlating results and the crud function already has that context from running the query.

backend/packages/wps-api/src/app/routers/fcm.py
conbrad: Ah ok yes makes sense now, thanks.

backend/packages/wps-api/src/app/routers/fcm.py
brettedw: done in [3d7c482](https://github.com/bcgov/wps/pull/5657/commits/3d7c48247538988bd43ce61be547518263831abc)

backend/packages/wps-api/alembic/versions/e4a9f2c7d1b6_unique_device_token_device_id.py
brettedw: migration updated here: [e1ee87d](https://github.com/bcgov/wps/pull/5657/commits/e1ee87d875fcacd08b461fc53fdf538ec93fcc78)

conbrad (approved):
Looks good, thanks for fixing!

PR #5682, opened 2026-08-10
Docs: Fix broken links and commands that no longer work

Gabriel-Dalton:
While reading through the documentation to get oriented in the project, I ran into a number of links and commands that no longer work. This PR fixes the ones I could verify.

Everything below was checked against the repository or the referenced source rather than assumed. Happy to split this up, trim it, or close it if any of it is unwelcome or already in progress.

## Broken links

| Fix | How it was verified |
| --- | --- |
| README linked `LICENSE.md` | No such file; the file is `LICENSE` |
| README linked `CONDUCT.md` at `/blob/master/` | `git ls-remote --heads origin master` returns nothing, so there is no `master` branch and this 404s |
| README PEP8 and PEP20 links | `python/peps/blob/master/pep-0008.txt` and `pep-0020.txt` both return HTTP 404. `python/peps` moved off `master` and away from `.txt`. Replaced with the canonical `peps.python.org` URLs |
| `CONVENTIONS.md` linked `scripts/lint.sh` | No such file anywhere in the repo |
| `CONVENTIONS.md` linked `.githooks/pre-push` | Relative from `docs/`, this resolves to `docs/.githooks/pre-push`. The file is at the repo root |
| `model_data.md` image | `../../../docs/images/Grid_wx_station.png` needs six `../` from `backend/packages/wps-api/src/app/jobs/`. Looks like it was missed when the backend moved to `packages/` |
| `asa-go/README.md` android link | Pointed at `/Users/breedwar/projects/other/wps/mobile/asa-go/android`, a local absolute path |

After this change, every relative Markdown link in the repository resolves. Before it, the four repo-relative ones above did not.

## Commands that do not work as written

**`CONVENTIONS.md`** documents `make lint`, but the Makefile has no `lint` target. Replaced with the ruff invocation the `lint-api` job in `integration.yml` actually runs, and pointed at that workflow so the two stay visibly linked.

**`NATS.md`** says to run the API with `cd {path_to_repo}/wps/api` and `make run`. There is no `api` directory and no `run` target. Updated to the command documented in `backend/packages/wps-api/README.md`.

**`CLUSTER_DB.MD`** manual backup example reads `--overwritepostgres-operator.crunchydata.com/...`. The missing space makes the annotation part of the flag, so the command fails.

**`sfms/README.md`** config example sets `destination=`, but `sfms.py` reads `config.get('url')` at lines 164 and 175, and the prose above the example correctly lists `url`. Following the example produces a config the script cannot use.

`NATS.md` also had its step numbering run 1, 4, 5, 6, which I have renumbered.

## Rendering and metadata

`openshift/README.md` had `#####Homebrew install:` with no space after the hashes, so it rendered as literal text instead of a heading.

The README licence badge's alt text read `MIT License`, while `LICENSE` is Apache 2.0 and the badge itself renders "Apache-2.0". Anyone using a screen reader was told the wrong licence.

## Notes for review

Documentation only. No code, configuration, or generated files are touched.

This overlaps with #5683 (spelling and grammar) in five files, but on non-adjacent lines, so the two merge cleanly in either order.

I left the `docs/architecture/sfms.md` and `docs/ARCH.md` prose alone here, since the mermaid diagrams are marked work in progress and I did not want to touch diagram content in a fix-up PR.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,5 @@
 [![Issues](https://img.shields.io/github/issues/bcgov/wps.svg?style=for-the-badge)](/../../issues)
-[![MIT License](https://img.shields.io/github/license/bcgov/wps.svg?style=for-the-badge)](/LICENSE)
+[![Apache 2.0 License](https://img.shields.io/github/license/bcgov/wps.svg?style=for-the-badge)](/LICENSE)
 [![Lifecycle](https://img.shields.io/badge/Lifecycle-Stable-97ca00?style=for-the-badge)](https://github.com/bcgov/repomountie/blob/master/doc/lifecycle-badges.md)
 [![codecov](https://codecov.io/gh/bcgov/wps/branch/main/graph/badge.svg?token=QZh80UTLpT)](https://codecov.io/gh/bcgov/wps)
 
@@ -60,7 +60,7 @@ Refer to [web/README.md](web/README.md)
 
 ## License
 
-[Apache 2.0](https://www.apache.org/licenses/LICENSE-2.0) License - see the [LICENSE.md](https://github.com/bcgov/wps/blob/main/LICENSE)
+[Apache 2.0](https://www.apache.org/licenses/LICENSE-2.0) License - see the [LICENSE](https://github.com/bcgov/wps/blob/main/LICENSE)
 
 ## Contributing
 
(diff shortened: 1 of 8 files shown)


conbrad (approved):

conbrad:
Thanks for contributing @Gabriel-Dalton !

Issue #5689, opened 2026-08-11
Percentile Calculator offers stations it has no data for


Gabriel-Dalton:
## Summary

The calculator's dropdown offers 334 stations from `/api/stations/`, but only 271 have usable pre-calculated percentile data. The remainder fall into two groups:

- 63 stations with no data file at all. These currently return a 500 (#5688).
- 10 stations whose data file exists but is entirely null, e.g. 321 GREENSTONE HUB, 3190 ROCKINGHAM. These return 200 and render every index as "Not available" with "Data only available for 0 of 10 years".

Even once the 500 is fixed, both groups are selectable options that cannot produce a result.

A station with no data file, selected alongside two that work (#5688):

![500 error](https://raw.githubusercontent.com/Gabriel-Dalton/wps/assets/percentile-calculator-screenshots/.screenshots/03-error-poisons-batch.png)

A station whose data file is entirely null. No error, but no result either:

![no data rendering](https://raw.githubusercontent.com/Gabriel-Dalton/wps/assets/percentile-calculator-screenshots/.screenshots/04-blank-321.png)

## The 63 with no data file

| Count | Category | Examples |
|------:|----------|----------|
| 20 | Environment Canada `(EC)` | TOFINO AP, VANCOUVER AP, VICTORIA AP |
| 11 | Other agency `(NC)` / `(MC)` / `(CRD)` | 31N (CRD) 543, 14G (CRD) 1142 |
| 12 | Test / project stations | QD 9 TEST, QD1 10M TEST, BLOWDOWN PASS PROJECT |
| 20 | Other | TERRACE, KOMIE, BLUEBERRY, COBBLE HILL |

Full list: 1, 2, 26, 44, 49, 69, 83, 88, 122, 142, 157, 186, 203, 205, 229, 395, 434, 435, 480, 543, 896, 915, 967, 1013, 1015, 1016, 1141, 1142, 1143, 1166, 1173, 1851, 3730, 3810, 3851, 3873, 4232, 4270, 4332, 4352, 4393, 4513, 4552, 4653, 4713, 4793, 4973, 5154, 5716, 5796, 5798, 5799, 5996, 5997, 6016, 6076, 6117, 6136, 6156, 6177, 6196, 6216, 6236

## Context that may be relevant

[...]

Issue #5688, opened 2026-08-11
Percentile Calculator returns 500 for stations that have no pre-calculated data file


Gabriel-Dalton:
## Summary

`POST /api/percentiles/` returns a 500 when a requested station has no pre-calculated data file. Because codes are processed as a batch, one such station also fails every other station selected alongside it.

## Steps to reproduce

Breaks (VANCOUVER AP, code 69):
https://psu.nrs.gov.bc.ca/percentile-calculator?codes=69&timeRange=10

Works (codes 213 and 267):
https://psu.nrs.gov.bc.ca/percentile-calculator?codes=213,267&timeRange=10

Adding the unsupported station breaks the working pair:
https://psu.nrs.gov.bc.ca/percentile-calculator?codes=213,267,69&timeRange=10

Selecting only stations that have data works:

![working result](https://raw.githubusercontent.com/Gabriel-Dalton/wps/assets/percentile-calculator-screenshots/.screenshots/01-works-213-267.png)

Adding VANCOUVER AP (EC) (69) to that same pair fails the whole request:

![error state](https://raw.githubusercontent.com/Gabriel-Dalton/wps/assets/percentile-calculator-screenshots/.screenshots/03-error-poisons-batch.png)

Against production:

```
POST /api/percentiles/ {"stations":[213],          ...}  -> 200
POST /api/percentiles/ {"stations":[267],          ...}  -> 200
POST /api/percentiles/ {"stations":[213,267],      ...}  -> 200
POST /api/percentiles/ {"stations":[1142],         ...}  -> 500
POST /api/percentiles/ {"stations":[1142,213,267], ...}  -> 500
```

The page shows "Error occurred (while getting the calculation result)."

## Cause

`app/percentile.py` builds a filename per station code and calls `StationSummary.parse_file(filename)` with no existence check, so a missing file raises `FileNotFoundError` and surfaces as a 500.

The same function already returns a clean 400 for its other two failure modes (empty station list, unsupported year range). A missing data file looks like the case that was not covered.

## Expected

A 400 naming the unsupported codes, consistent with the existing 400s, and without failing the other stations in the request.

There is precedent that a data-less station is not meant to error: ten currently-selectable stations have null pre-calculated values and return 200, rendering as "Not available" with a "Data only available for 0 of 10 years" notice, e.g.

https://psu.nrs.gov.bc.ca/percentile-calculator?codes=321&timeRange=10

[...]

PR #5690, opened 2026-08-11
Fix: return 400 instead of 500 for stations without percentile data

Gabriel-Dalton:
Refs #5688

`get_precalculated_percentiles` opened a data file per station code with no existence check, so requesting a station without pre-calculated data raised `FileNotFoundError` and surfaced as a 500. Because the codes are processed as a batch, one unsupported station also failed every other station in the same request.

This adds an up-front check and raises a 400 naming the unsupported codes, matching the existing handling for an empty station list and an unsupported year range.

Verified against the checked-in data in `app/data/2014-2023/`:

```
[213, 267]        -> 200, mean BUI 100.49
[1142]            -> 400 "No percentile data available for station code(s): 1142."
[1142, 213, 267]  -> 400 "No percentile data available for station code(s): 1142."
[543, 1142, 213]  -> 400 "No percentile data available for station code(s): 543, 1142."
[]                -> 400 "Weather station is not found." (unchanged)
2004-2019 range   -> 400 "The year range is not currently supported." (unchanged)
```

This does not make those stations calculable; it replaces the crash with a clear message. The wider question of which stations should be offered is in #5689.

Am I missing something here? If a different status code or message would fit your conventions better, happy to change it.


brettedw (approved):

brettedw:
Closing for https://github.com/bcgov/wps/pull/5709 containing the same changes submitted here

PR #5758, opened 2026-08-25
Dev: Migrate remaining raw GDAL cutline and masking code onto WPSDataset

conbrad:
`hfi_percent_conifer.py`, `hfi_minimum_wind_speed.py`, `process_fuel_type_area.py`, `viirs_snow.py`, `fuel_grid/tpi_fuel_area.py`, `fuel_grid/fuel_type_area.py`, and `snow.py` each hand-rolled their own `gdal.Warp` cutline clipping and manual `del`/`=None` dataset cleanup, duplicating the same pattern with slightly different resource lifecycles.

`WPSDataset` gains a `clip_to_geometry` method, a `read_array()` method, and a `Georeference` (geotransform, projection) type that `from_array` accepts directly from a source `WPSDataset` or explicitly, so every call site above shares one resource-cleanup path instead of reimplementing it. `snow.py`, `classify_hfi.py`, and `viirs_snow.py` were simplified to chain `WPSDataset` operations in memory instead of round-tripping through intermediate file paths, `polygonize_in_memory` moved from `wps_shared/utils` to `wps_shared/geospatial` and onto `WPSDataset`, and the previously-untested pure area-calculation helpers in `fuel_grid/fuel_type_area.py` and `fuel_grid/tpi_fuel_area.py` gained unit tests.

# Test Links:
[Landing Page](https://wps-pr-5758-e1e498-dev.apps.silver.devops.gov.bc.ca/)
[MoreCast](https://wps-pr-5758-e1e498-dev.apps.silver.devops.gov.bc.ca/morecast)
[Percentile Calculator](https://wps-pr-5758-e1e498-dev.apps.silver.devops.gov.bc.ca/percentile-calculator)
[FireCalc](https://wps-pr-5758-e1e498-dev.apps.silver.devops.gov.bc.ca/fire-behaviour-calculator)
[FireCalc bookmark](https://wps-pr-5758-e1e498-dev.apps.silver.devops.gov.bc.ca/fire-behaviour-calculator?s=266&f=c5&c=NaN&w=20,s=286&f=c7&c=NaN&w=16,s=1055&f=c7&c=NaN&w=NaN,s=305&f=c7&c=NaN&w=NaN,s=344&f=c5&c=NaN&w=NaN,s=346&f=c7&c=NaN&w=NaN,s=328&f=c7&c=NaN&w=NaN,s=1399&f=c7&c=NaN&w=NaN,s=334&f=c7&c=NaN&w=NaN,s=1082&f=c3&c=NaN&w=NaN,s=388&f=c7&c=NaN&w=NaN,s=309&f=c7&c=NaN&w=16,s=306&f=c7&c=NaN&w=NaN,s=1029&f=c7&c=NaN&w=NaN,s=298&f=c7&c=NaN&w=NaN,s=836&f=c7&c=NaN&w=NaN,s=9999&f=c7&c=NaN&w=NaN)
[Auto Spatial Advisory (ASA)](https://wps-pr-5758-e1e498-dev.apps.silver.devops.gov.bc.ca/auto-spatial-advisory)
[HFI Calculator](https://wps-pr-5758-e1e498-dev.apps.silver.devops.gov.bc.ca/hfi-calculator)
[SFMS Insights](https://wps-pr-5758-e1e498-dev.apps.silver.devops.gov.bc.ca/insights)
[Fire Watch](https://wps-pr-5758-e1e498-dev.apps.silver.devops.gov.bc.ca/fire-watch)
[Weather Toolkit](https://wps-pr-5758-e1e498-dev.apps.silver.devops.gov.bc.ca/weather-toolkit)


backend/packages/wps-shared/src/wps_shared/geospatial/wps_dataset.py
brettedw, line 1: It's a file untouched by this PR so far but `generate_glc_x_10_rasters.py` uses `from_array` and will need it's call updated

backend/packages/wps-shared/src/wps_shared/geospatial/wps_dataset.py
```diff
@@ -45,31 +54,68 @@ def __enter__(self):
...
+        return np.asarray(self) < other
+
+    def __le__(self, other):
+        return np.asarray(self) <= other
+
+    def __gt__(self, other):
+        return np.asarray(self) > other
+
+    def __ge__(self, other):
+        return np.asarray(self) >= other
```
brettedw: I'm not quite sure what I think about these. When we're doing comparisons like `np.select([source < 4000, source < 10000], [0, 1], default=2)` aren't we reading the raster array twice?

I guess in my mind it's cleaner and easier to read if we just read the data once and then do the comparison like:

```Python
  source_data = np.asarray(source)
  classified = np.select(
      [source_data < 4000, source_data < 10000],
      [0, 1],
      default=2,
  )
```

brettedw: Or maybe even add a `read_array()` method?

backend/packages/wps-api/src/app/auto_spatial_advisory/classify_hfi.py
```diff
@@ -1,54 +1,18 @@
...
-
-    # Explicit delete to make sure underlying resources are cleared up!
-    del source_band
-    del source_tiff
-    del target_band
-    del target_tiff
-    del output_driver
+    with gdal_s3_context(), WPSDataset(source_path) as source:
+        classified = np.select([source < 4000, source < 10000], [0, 1], default=2)
+        WPSDataset.from_array(
```
brettedw, line 23: should this be context managed for cleanup?

backend/packages/wps-shared/src/wps_shared/geospatial/wps_dataset.py
conbrad: Good catch on the double-read. Your fix would work, but it puts the burden on every caller to
remember `np.asarray(source)` once before comparing. We've already got three call sites doing
`source < X` / `source > X` directly (`classify_hfi.py`, `snow.py`, and now `viirs_snow.py`), and
it's an easy thing to forget on the next one.

Pushed the fix into `__array__` itself instead: it caches the read on first access, so
`np.select([source < 4000, source < 10000], ...)` now only calls `ReadAsArray()` once
automatically, no matter how many comparisons chain off the same instance. Verified it's safe by
checking every method on the class — nothing mutates `self.ds`'s band data in place, every writer
(`__mul__`, `warp_to_match`, `clip_to_geometry`, `from_array`) builds a new `WPSDataset` instead,
so there's no stale-cache risk. The cache also clears in `close()`, so a use-after-close bug still
fails loudly instead of silently returning old data.

https://github.com/bcgov/wps/pull/5758/commits/6d990d0855d9970a06a2bf6244a9e2b69bf64845

backend/packages/wps-api/src/app/auto_spatial_advisory/classify_hfi.py
conbrad: Yes thanks for catching, fixed here with regression test: https://github.com/bcgov/wps/pull/5758/commits/021856db067aa1b82859f42672dde7666ed2bf31

backend/packages/wps-api/src/app/auto_spatial_advisory/snow.py
```diff
@@ -1,121 +1,55 @@
...
+        ):
+            # Reproject the snow coverage data to match the HFI grid (same projection, extent and
+            # pixel size), classify it into a mask, then apply that mask to the HFI raster.
+            with (
+                snow_source.warp_to_match(hfi_source) as warped_snow,
+                classify_snow_mask(warped_snow) as snow_mask,
+            ):
+                # The snow mask has values of 0 (snow covered) or 1 (snow free); multiplying
+                # applies the mask.
+                masked = hfi_source * snow_mask
```
brettedw: I think this is a WPSDataset too, should it be context managed?

backend/packages/wps-shared/src/wps_shared/geospatial/wps_dataset.py
brettedw: Hmm, I think this could potentially lead to a complex bug. I don't think we should cache here.

There's 2 ways to read the data as an array and I think if we cache like this they could disagree. Couldn't the cached array be edited? Then if someone were to call: `dataset.ds.GetRasterBand(1).ReadAsArray()` they would get the original gdal array? 
In my brain it's also increasing complexity of WPSDataset

conbrad: You're right, I reverted and adopted your approach here: https://github.com/bcgov/wps/pull/5758/commits/307199b54bd4355d74318cd480fe2b74a8848fa0

backend/packages/wps-api/src/app/auto_spatial_advisory/snow.py
conbrad: Yes thanks! https://github.com/bcgov/wps/pull/5758/commits/740c2a6e0baadd09c9329e9bad33b27a98897153

backend/packages/wps-shared/src/wps_shared/geospatial/wps_dataset.py
brettedw: I almost think a `read_array()` method might even be clearer/easier, but whatever you think there!

brettedw: Also I think `__le__`, etc are unused now with that change? Do we need them?

backend/packages/wps-api/src/app/auto_spatial_advisory/classify_hfi.py
```diff
@@ -1,54 +1,19 @@
...
-
-    # Explicit delete to make sure underlying resources are cleared up!
-    del source_band
-    del source_tiff
-    del target_band
-    del target_tiff
-    del output_driver
+    with gdal_s3_context(), WPSDataset(source_path) as source:
+        source_data = np.asarray(source)
+        classified = np.select([source_data < 4000, source_data < 10000], [0, 1], default=2)
```
brettedw: I'm not sure that we have any, I don't think we do, but how does this classify np.NaN values? Would both be false and possibly get classified as warning?

backend/packages/wps-shared/src/wps_shared/geospatial/wps_dataset.py
conbrad: Yeah the abstraction failed because we can't rely on everything being in memory, went with `read_array` and removed all the protocols including `__array__`: https://github.com/bcgov/wps/pull/5758/commits/ee9367040753c588d37d73e6302a2fcdd4a2cc93

backend/packages/wps-api/src/app/auto_spatial_advisory/classify_hfi.py
conbrad: Thanks for catching this, huge issue, I added a fix with a regression test: https://github.com/bcgov/wps/pull/5758/commits/4450c6b64184534bd560145043ca6bab1efa92fd

conbrad: Also for classifying by threshold: https://github.com/bcgov/wps/pull/5758/commits/f77b9113a48cba47b14cb7285821c89e475b0713

